PR #345, opened 2015-04-16
DB Connection not static

bmackinney:
Line 99 needs more information.
```Situations when you might want to do this include [ Add list of situations] For these situations, please [Add recommended fixes]```

Please please add
- [ ] Situations where users want to encode their db connection info, or delete the phrase. 
- [ ] Recommended solution, adding an example of the appropriate environmental variable to the codebase in a snippet.


rachelwhitton:
@bmackinney is this related to issue #393 ? Recently closed that one out.

Issue #423, opened 2015-04-21
Create articles for every directory


bmackinney:
Each directory should have an article of the same name in its parent directory, for the URLs to be active, to increase the amount of documentation we have for each part of the platform, and to re-enable breadcrumbs. These articles represent the three dashboards, all of the tools within the sites dashboard, the hosting architecture, and the frameworks we run.   

Missing/draft articles: 
- [x] architecture
- [x] architecture/edge.md
- [x] drupal
- [x] wordpress
- [x] users
- [x] sites
 - [x] backups
 - [x] code
 - [x] create
 - [x] database
 - [x] domains
 - [x] files
 - [x] newrelic
 - [x] security
- [x] organizations

PR #478, opened 2015-04-26
Add Domains and SSL Tool article

suzannealdrich:
Add Domains and SSL Tool article with overview of base domains, recommended DNS settings, and adding HTTPS for secure communications. Addresses Domains in #423


bmackinney:
@timani @newtoid  please review.

timani:
+1

bmackinney:
@nataliejeremy ready for editing

ari-gold:
After this gets published what do you think about linking to this doc from other docs that reference adding domains and setting up DNS (like Going Live)?

suzannealdrich:
Yes, and the other articles it is linking to will also get some once-overs from me, at least.

PR #486, opened 2015-04-28
Updating symlinks and desriptions for non-standard-files-locations.md

atjuch:
Creating clearer instructions for our Drupal and WordPress users since we have had numerous tickets for symlink requests.

@rachelwhitton or @nataliejeremy Please review


atjuch:
Going to come back to this after some further testing.

PR #488, opened 2015-04-29
Correct hotfixes article

### LINKED ISSUE #410: Add more detail to hotfix docs
Hotfix docs should include more detail so that people don't do strange things We should add a note to the hotfix page that tags we create can NOT be deleted.
## Problem
`pantheon_test_260` trying to remove via git using `git tag -d` results in an error stating the tag is a deployment tag and it can't be removed.
>Can you help me deleting the tag or pointing me in the right direction to get that done?

Unfortunately we do support deleting those tags as the message noted. 
## Solution
If you need to have a change pushed to Test you should use the workflow and go from Dev → Test → Live rather than delete the tag.
Push a change from Dev → Test and this should fix this, we use those tags as part of the internal deployment process so deleting is not an option.
In the event you want to remove a commit, **rebasing** is a better option so the commit itself is no longer in the codebase.

Just to be clear that this NOT a Pantheon issue

--submitted by @timani

### LINKED ISSUE #455: Remove note from hotfix documentation about deploy being available
When you have a hotfix in Test there is no option to deploy that change to Live. We should remove this and recommend pushing the hotfix to Live when they are ready.

bmackinney:
Closes #455 CSE please review. We should also try to resolve #410 with this pull request.


source/docs/articles/sites/code/hot-fixes.md
```diff
@@ -2,8 +2,10 @@
 title: Hot Fixes
 description: Learn how to deploy a hotfix on Pantheon.
 category:
-  - developing
+  - developing
 ---
+For Experts only. You should not need to attempt this if you use [Multidev](/docs/articles/sites/multidev) and keep commits from reaching Dev that you do not intend on deploying.
```
timani: What is the CSS class for making this a bootstrap alert? This should smell of a bad idea from first sight

bmackinney: https://github.com/pantheon-systems/documentation/issues/508

source/docs/articles/sites/code/hot-fixes.md
```diff
@@ -2,8 +2,10 @@
 title: Hot Fixes
 description: Learn how to deploy a hotfix on Pantheon.
 category:
-  - developing
+  - developing
 ---
+For Experts only. You should not need to attempt this if you use [Multidev](/docs/articles/sites/multidev) and keep commits from reaching Dev that you do not intend on deploying.
+
 Sometimes it's necessary to push a quick fix without pushing everything that's been going on in dev. This is called a "hotfix", and this article will help you do it on Pantheon.
```
ari-gold: How about removing this line?

source/docs/articles/sites/code/hot-fixes.md
ari-gold: Suggest adding something like language on *Using the Pantheon Workflow*:  "We do not recommend hotfixing. Hotfixes should be the exception, not the norm. Pushing a hotfix via Git is the only way to push code directly to Live without having to go through Dev and Test. Hotfixing is not a best practice."

source/docs/articles/sites/code/hot-fixes.md
bmackinney: @ari-gold I agree with your proposed change. Go ahead and replace the line with the comments of your comment.

PR #495, opened 2015-04-29
Changing "snippit" to "snippet"

stevector:
A spelling fix.


nataliejeremy:
@stevector Thanks for catching that!

Issue #502, opened 2015-04-30
Move articles beneath /architecture to /sites 


nataliejeremy:

rachelwhitton:
Closed via #522 
I will complete redirect tasks upon next deploy

PR #517, opened 2015-05-02
Restore breadcrumbs

### LINKED ISSUE #424: Restore breadcrumbs (requires parent article creation)
The repo has a library called autobreadcrumbs.js which takes each directory in the path and turns it into a hyperlink, assuming there will be an article at that path. I disabled the feature before we deployed because not all of the parent articles exist. To re-enable breadcrumbs in this fashion, we will need to ensure that for each subdirectory `/name`, the parent directory must have an article `name.md`

- [ ] restore the library
- [ ] style the block
- [ ] deployment blocked by #423

bmackinney:
Closes #424. Blocked by #502 and #503 and #505 
image

Open questions: 
 - Is this the right way to go? I used this: http://www.javascriptsource.com/navigation/automatic-breadcrumbs.html I came across this as well: http://stackoverflow.com/questions/14382848/create-scenario-like-breadcrumb
 - Should we get rid of the link to `Docs`? We should probably deactivate the link to the current page.


rachelwhitton:
+1 for removing link to the current page. Some show and some don't. Example: 
screen shot 2015-05-04 at 9 58 40 am
I don't mind the Docs link being there. 

+1 for the method used. The js script being local is best imo - as opposed to relying on an externally hosted resource.

bmackinney:
@JonathanPorta or @bensheldon can you take a look at the first open question on this PR? Does it even matter?

bensheldon:
@bmackinney this looks fine. If it works and you're happy with the output :+1:

source/_views/article.html
```diff
@@ -5,11 +5,11 @@
...
               <h1 class="pio-docs-title">{{ page.title }}</h1>
-              <!-- <script type="text/javascript">
-            #  breadcrumbs()
-
-            </script> -->
-
+              <script type="text/javascript">
+                <!--
+                breadcrumbs()
+                //-->
```
bensheldon: I don't think you want these `<!--` , `//-->` comments within your script tag

bmackinney: @bensheldon I thought that was odd too, but it's straight from the source. @rachelwhitton can you test without the comments?

rachelwhitton:
@bmackinney tested without comments on both guide.html and article.html
Screenshot:
screen shot 2015-05-05 at 2 37 52 pm

+1

rachelwhitton:
I also merged master, so you can view all landing pages when you build locally.

bmackinney:
Ok so it works without the comments. edit on github doesn't look right, but that may not persist into `output_prod`
@bensheldon can you take a look at the `breadcrumbs.js` file and let us know how to get rid of the link to the active page? @rachelwhitton if ben doesn't have time, feel free to tackle it.

rachelwhitton:
Moved to #528 
@bmackinney we okay to close this out and delete the branch?

PR #540, opened 2015-05-09
Fix typo: hotfixs => hotfixes

mdeltito:


nataliejeremy:
Thanks for catching this @mdeltito

Issue #453, opened 2015-04-22
Document SASS/Compass


rachelwhitton:
@suzannealdrich: have we documented anywhere that SASS/Compass isn’t supported? I see Less on the Known limitations page, probably want to add those keywords there.

suzannealdrich:
@rachelwhitton we completed this work in #548

PR #548, opened 2015-05-26
Update Known Limitations with SASS/Compass

suzannealdrich:
Along with LESS, Sass/Compass is not currently compatible with the platform and these resources will need to be pre-compiled on a local development server before being pushed to the site code repository. #453

### DIFF
--- a/source/docs/articles/sites/known-limitations.md
+++ b/source/docs/articles/sites/known-limitations.md
@@ -77,7 +77,7 @@ Large backups take longer, use more resources, and have a higher likelihood of f
 
 ## LESS (css)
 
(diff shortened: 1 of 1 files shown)


rachelwhitton:
+1 Thank you @suzannealdrich! Much appreciated :)

source/docs/articles/sites/known-limitations.md
```diff
@@ -77,7 +77,7 @@ Large backups take longer, use more resources, and have a higher likelihood of f
 
 ## LESS (css)
```
ari-gold: change heading to "CSS Preprocessors" ?

PR #562, opened 2015-05-29
Disable the posts content type

opdavies:
When running the `sculpin generate` command to generate the site, the following message is displayed.

> Didnt find at least one of this type : posts

This is because the core "Post" content type is being used. This PR disables the content type so that Sculpin doesn't look for it whilst generating, so doesn't show the error.


bmackinney:
:+1: Thanks. worked on my machine, @nataliejeremy has some pain from this, I think.

nataliejeremy:
This is great--thanks!

PR #570, opened 2015-06-01
Replaced upstream screenshot to show drupal upstream

hengkit:
Current screenshot has 2 issues. It's showing a bare URL instead of the name of the upstream and the article is Drupal oriented and it's showing a WordPress upstream


rachelwhitton:
+1 Thanks @hengkit!

ari-gold:
Looks good. Image should be compressed.

PR #571, opened 2015-06-02
Link goes to terminus 1 article instead of cli.

hengkit:
Updated link to point to correct article


rachelwhitton:
Thanks @hengkit!

PR #577, opened 2015-06-03
Adding Drupal Organizational Step to Define the SERVER_PORT Value

populist:
Here is an update to the TFA guide to address a change to the way SERVER_PORT is handled in the 2.4 version of the SAML library.


rachelwhitton:
Thanks @populist! Much appreciated

PR #587, opened 2015-06-04
Highlight CloudFlare HTTPS limitations

iamEAP:
Biggest bummer for me when I set this up a similar way was that New Relic's availability monitor no longer worked.  Would be good to highlight that.

Also, the legacy IE thing.

### DIFF
--- a/app/config/sculpin_site.yml
+++ b/app/config/sculpin_site.yml
@@ -84,3 +84,13 @@ authors:
       twitter: KateKligman
       linkedin: http://www.linkedin.com/in/katekligman
       bio: Open source contributor, indie game developer, retro gamer.
+  iameap:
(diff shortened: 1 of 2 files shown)


joshkoenig:
Thanks for the quick enhancement, Erik!

That's a bummer about New Relic. Seems bizarre that they wouldn't support SNI.

I agree that adding this section makes sense. I'd also suggest we add a credit for you. :)

iamEAP:
Rad. Credit ^.

I asked NR about SNI support in the pinger awhile back... I got the impression they're putting very little effort into improving their "classic" availability monitoring while they put more resources into [Synthetics](http://newrelic.com/synthetics).

nataliejeremy:
@iamEAP Thanks for adding this!

PR #595, opened 2015-06-09
update spindown time for LIVE sites

andrewmallis:
update scope of support to reflect new spindown time for LIVE environments, as per recent power-users post from @joshkoenig


joshkoenig:
Quick nitpick:

1) The non-live idle time is actually 1 hour, so we should update that also.
2) Can we change:

> Pantheon servers spin down...

to

> Pantheon *containers* spin down...

We don't ever spin down hosts unless they've been drained of containers and are being returned to the MotherCloud.

rachelwhitton:
Moved to PR #600

dustinleblanc:
When you said MotherCloud...I thought Mother Brain Mother Brain

PR #598, opened 2015-06-09
add mime types article

nataliejeremy:
Adding new doc per sprintly issue #2282. @newtoid @joshkoenig Please review--thanks!

### DIFF
--- a/source/docs/articles/sites/mime-types.md
+++ b/source/docs/articles/sites/mime-types.md
@@ -1,0 +1,22 @@
+---
+title: MIME Types
+description: Learn how to add, change, and remove a file extension to a MIME type on Pantheon.
+keywords: mime types, file extension, IANA, mime
+---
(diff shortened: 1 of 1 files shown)


source/docs/articles/sites/mime-types.md
```diff
@@ -0,0 +1,22 @@
+---
+title: MIME Types
+description: Learn how to add, change, and remove a file extension to a MIME type on Pantheon.
+keywords: mime types, file extension, IANA, mime
+---
```
joshkoenig: Let's add an intro paragraph:

Pantheon promotes and respects internet standards, and for MIME Types we treat the [IANA Media Type List](http://www.iana.org/assignments/media-types/media-types.xhtml) as the canonical source of what headers to serve for what file extensions. This doc page describes what to do if you have a file type that isn't handled by IANA, or require an alternative.

bmackinney:
:+1:

PR #605, opened 2015-06-10
WordPress Multisite not supported

ari-gold:


bmackinney:
@ari-gold please provide an explanation. @joshkoenig wrote the section this negates in February. @joshkoenig are you OK with the rewording? If not, we should establish guidelines for running multisite and supporting it. I rejected a similar pull request a while ago from @timani but I think if we don't do the above, CSE is expressing @hengkit tried to get it to work but could not and that's the last I heard.

ari-gold:
Deferring to @joshkoenig. Right now seems like the docs are conflicting. One saying yes, one saying no.

timani:
This is causing confusion and resulting in users wasting time and getting frustrated or upset. This ambiguity is putting CSE in a compromising position as the response is "We heard we can so you are not correct" and I have had to get on a call or two about this too. 

Until this is an official product offering, it's not supported, quite simple imo. When it is, then we update accordingly. By this logic we should have the same thing about Drupal multisite but we don't.  

+100

bmackinney:
@ari-gold is right. [One doc says yes, the other no.](https://github.com/pantheon-systems/documentation/search?utf8=%E2%9C%93&q=wordpress+multisite). I think we could say "possible, but unsupported. You can try it, but please don't submit a ticket if something goes wrong" - I don't like that idea, but it's causing confusion and we don't have the use cases.

suzannealdrich:
How about saying it voids your warranty, just like turducken sites.

nataliejeremy:
@sukottokun Have you reached a decision on this yet? We'd like to move this forward and out of the queue. Thanks!

sukottokun:
@nataliejeremy Not supported until config changes make it to the roadmap.

PR #608, opened 2015-06-10
Add UUID definition

### LINKED ISSUE #376: Create clear definition of UUID and consistency in the term's usage
Story:  As a developer I want a clear definition of UUID and consistency in the term's usage so I don't have to figure out what UUID means in this context.

https://pantheon.io/docs/articles/local/ssh-tunnels-for-secure-connections-to-pantheon-services/

In this document, we use UUID to mean something that looks like this: 
0ee74e78-d8aa-4488-a530-7a2d4c82fda9

We don't define it, but that is what is implied.

In terminus, if you do a 
$ terminis sites list

it gives you the slug for the site under the heading UUID.

It would be good to define exactly what is meant by UUID and then be consistent in our usage of the term. (Possibly even linking it back to the definition the first time it is used in a document.

rachelwhitton:
Closes #376 

@nataliejeremy can you review? Does it make sense to add this to the Workflow doc? Is there a better doc for it?


ari-gold:
Definitions look great! 

But, like you said, not sure they belong in Workflow doc. Maybe a new doc? 

Or 
Site UUID definition could go in https://pantheon.io/docs/articles/sites/ 
Org UUID could go in https://pantheon.io/docs/articles/organizations/
Product UUID could go in https://pantheon.io/docs/articles/sites/create/choosing-start-state/

rachelwhitton:
@ari-gold I like the idea of adding them to the docs you linked! Makes sense to have a place to link back to for specific uses. 

+1

@nataliejeremy let me know what you think and I'll update. Adding WIP to the PR.

nataliejeremy:
+1

rachelwhitton:

PR #628, opened 2015-06-16
Update Redis configuration troubleshooting guide

suzannealdrich:
Explain why configuration of cache backend inside `isset($_SERVER['PANTHEON_ENVIRONMENT']` block will result in issues for a Pantheon site.

### DIFF
--- a/source/docs/articles/sites/redis-as-a-caching-backend.md
+++ b/source/docs/articles/sites/redis-as-a-caching-backend.md
@@ -66,7 +66,9 @@ The common community module for Drupal to use Redis is simply called [redis](htt
       // Use Redis for Drupal locks (semaphore).
       $conf['lock_inc'] = 'sites/all/modules/redis/redis.lock.inc';
     }
-  ```<div class="alert alert-info">
+  ```
+  
+  <div class="alert alert-info">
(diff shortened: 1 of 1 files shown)


ari-gold:
LGTM :+1:

JessiFischer:
Can we set this line to bold to make sure it stands out to readers?

"Here is an example of a block that will result in an incorrectly configured cache backend:"

source/docs/articles/sites/redis-as-a-caching-backend.md
```diff
@@ -163,6 +165,44 @@ To check the # of keys in the cache, you can use the `DBSIZE` command. The follo
 
 ## Troubleshooting
 
+### Redis is enabled but doesn't have any data
+
+When the Dashboard Status check reports that Redis is enabled but doesn't have any data (0 keys found), you'll want to confirm the logic behind the check for PANTHEON_ENVIRONMENT in your `settings.php` Redis cache configuration. Depending on the kind of test you're performing, you’ll get different results. Here is an example of a block that will result in an incorrectly configured cache backend:
```
JessiFischer: Sorry, should have commented here. The last line in this paragraph should be bold in order to ensure readers don't see the code snippet and grab without understanding it is an example of an *incorrect* code block.

Issue #516, opened 2015-05-01
Document how to rename a branch for Multidev compatibility


bmackinney:
![image](https://cloud.githubusercontent.com/assets/646061/7437978/f706fbaa-f011-11e4-9f01-a8039602d3ca.png)
The 11-character limit means that branches with names exceeding the limit pushed to Pantheon cannot turn into CDE's. The solution is to rename the branch. 
Aritcle and first comment offer the procedure: http://www.benjaminlhaas.com/blog/locally-and-remotely-renaming-branch-git
thanks to @alphex for the prompt in power-users.

nataliejeremy:
@bmackinney Think it's an update to this doc? https://pantheon.io/docs/articles/sites/multidev/

bmackinney:
#637

PR #637, opened 2015-06-17
add rename a branch, move create branch locally

nataliejeremy:
This closes issue #516 @bmackinney Please review.


source/docs/articles/sites/multidev.md
```diff
@@ -90,16 +97,30 @@ A branch with no environment associated with it can be deleted by going to Multi
 
 Branches can be deleted locally and the commit can be pushed to Pantheon, but this may have unintended consequences if an environment is associated with it; use the interface instead.​
 
+## Rename a Branch
+There is an 11-character limit for branch names. If you push a branch to Pantheon that exceeds the character limit, it cannot become a cloud development environment (CDE). The solution is to rename the branch. This is only recommended if you don't have any other users working on this branch, or if you have already coordinated with them.
+
+1. Rename the branch. Example: `git branch -m old-branch-name new-name`
+2. Change the remote branch. Because you can't directly rename the remote branch, you'll need to delete the remote branch with the old name and re-create it with a new name. Start by getting the remote name:  
+`git remote -v`
```
rachelwhitton: You don't need to delete/re-create the branch to rename it - you can run `git branch -m new-name` while you have the branch in question checked out locally and then `git push origin new-name` to push it remotely. The name change automatically populates in the Multidev branch overview and the Create Environment button becomes clickable. 

@bmackinney can you confirm?

bmackinney: image
Name isn't updated, extra branch reflected. 
`git branch` reveals:
```
git branch
* config
  master
```
It does work.

nataliejeremy: Which one do we want to recommend? If it changes it should be updated here: https://github.com/pantheon-systems/documentation/issues/516?

rachelwhitton: Ah that's right. It creates the branch with the same working index and commit history, but leaves the original branch. Would it be easier to tell them to use the steps I mentioned and then delete the original branch with the longer name? Something like..

1. Rename the branch. Example: `git branch -m old-branch-name new-name`
2. Push the renamed branch. Example `git push origin new-name`
    This will create a new branch with the commit history intact. From the Multidev overview, click on **Git Branches** and delete the original branch. You will now be able to create an environment associated with the renamed Git branch. 

<hr>

I don't think we need to advise them to get the name of their remote repo (`git remote -v`) - by default it's origin when they clone.

PR #647, opened 2015-06-22
Update Bots-and-indexing.md

ttrowell:
This PR is to update the snippet so it won't throw errors:
@bmackinney please review

use alternate line adding hash marks:

$pattern = '#/pantheon.io/sitemap.xml$/#';


bmackinney:
Newton +1 to this.

PR #657, opened 2015-06-25
deleted the dash in between "sites" and "create" because sites-create…

okoslow:
…-from-import is not a registered command (modification in source/docs/articles/sites/create/importing-drush-archives-with-terminus.md)


nataliejeremy:
@okoslow Thanks for updating that!

Issue #556, opened 2015-05-27
Remove content recommendations for STYXKEY for HTML5 & CSS3


timani:
https://pantheon.io/docs/articles/sites/varnish/caching-advancedtopics/

After seeing STYXKEY I can say it is hard for users to digest. Doing things like mobile detection using modules like https://www.drupal.org/project/mobile_switch_varnish are not supported on the plaftform as the VCL can not be edited. It is also no longer the best or easiest way to do this.

### Summary
I vote that we remove the "Content Customization" section for STYXKEY and recommend this be done on the client side using browser detection, orientation or features like aspect ratio using HTML5, CSS3 and JS.

## Modernizr
Users can allow for a wider selection of browser properties in order to make their themes more responsive. This will also allow requests to have the benefit of being saved in Varnish and rendering correctly depending on the requirements.

## Device detection

A full list of the devices and their support for HTML5 is available on https://html5test.com where a comparison can be done for a variety of clients:

Desktop browsers: https://html5test.com/results/desktop.html 
Tablet browser: https://html5test.com/results/desktop.html 
Mobile browsers: https://html5test.com/results/desktop.html 
Other browsers: https://html5test.com/results/other.html 

## Resources

**Modernizr**
Main site: http://modernizr.com
Official documentation: http://modernizr.com/docs
Modernizr browser feature support matrix: http://www.browserscope.org/user/tests/table/agt1YS1wcm9maWxlcnINCxIEVGVzdBib2KQGDA
Detectizr an extension to detect device features: https://github.com/barisaydinoglu/Detectizr

**Drupal**
Modernizr Module - https://www.drupal.org/project/modernizr
Browser feature detection: https://www.drupal.org/node/1913726
Modernizr.load() and Drupal: Basics - https://www.drupal.org/node/1913830
Implementing the Modernizr Test API within a module or theme - https://www.drupal.org/node/2332277

**Wordpress**
Loading Modernizr for use in a plugin: http://wordpress.stackexchange.com/a/62362

nataliejeremy:
See details on PR #671

PR #671, opened 2015-06-30
remove styxkey, add modernizr

nataliejeremy:
This closes issue #556 @timani please review!

### DIFF
--- a/source/docs/articles/sites/varnish/caching-advancedtopics.md
+++ b/source/docs/articles/sites/varnish/caching-advancedtopics.md
@@ -44,17 +44,47 @@ WordPress does not use PHP session cookies; however, some themes and plugins do.
 
 ## Geolocation, Referral Tracking, Content Customization, and Cache Segmentation
 
-A site may need to deliver different content to different users without them logging in or starting a full session (either of which will cause them to bypass the page cache entirely). Pantheon recommends doing this on the client side using browser detection, orientation, or features like aspect ratio using HTML5, CSS3, and JavaScript.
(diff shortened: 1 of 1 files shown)


source/docs/articles/sites/varnish/caching-advancedtopics.md
```diff
@@ -42,35 +42,20 @@ Drupal uses SESS-prefixed cookies for its own session tracking, so be sure to na
 ### WordPress Sites
 WordPress does not use PHP session cookies; however, some themes and plugins do. If you are using a theme or plugin that requires PHP sessions, you can install [Pantheon-sessions](https://wordpress.org/plugins/wp-native-php-sessions/ "Panthon Session WordPress plugin"). It is designed to handle the naming properly.
 
-## Geolocation, Referral Tracking, Content Customization, and Cache Segmentation Using STYXKEY
+## Geolocation, Referral Tracking, Content Customization, and Cache Segmentation
```
nataliejeremy: @timani Does this header still work? If not please suggest a new one.

davidstrauss:
-1 STYXKEY is actively used for capabilities that are unrelated to the replacement methods.

timani:
It is perhaps the biggest source of pain for users implementing this. I have seen almost every case use HTML5 to handle this leave satisfied vs those who continue struggling with this.

We can keep it but we should make it clear mark it the less prefered option. #1 options should be the modern solution built in to most browsers

nataliejeremy:
@davidstrauss @timani Please review and +1 or comment on the changes.

nataliejeremy:
@davidstrauss Can you +1 if you approve the changes?

davidstrauss:
> It is perhaps the biggest source of pain for users implementing this. I have seen almost every case use HTML5 to handle this leave satisfied vs those who continue struggling with this.

STYXKEY was not added for the purpose of fake responsive design; it is no surprise users don't have a good time trying to use it that way. It is around to facilitate page keying for the many other use cases for caching arbitrary page variants. +1 on restoring docs for it, though people should be warned that it's not for fake responsive design.

PR #691, opened 2015-07-07
WP sessions is not a silver bullet

timani:
Needs to be updated on the plugin README


rachelwhitton:
Screenshot:
screen shot 2015-07-08 at 9 33 03 am
+1

PR #699, opened 2015-07-09
update training link

nataliejeremy:
@rachelwhitton please merge. Thanks!


rachelwhitton:
@nataliejeremy Should we redirect them to something like this instead of pantheon.io? https://pantheon.io/resources?field_resource_type_value%5B%5D=Webinar

nataliejeremy:
@rachel no, that's specific to Brian's webinar that used to be monthly but is now weekly.

PR #706, opened 2015-07-10
Adding maintenance mode issue with global redirect

slashzero:
Typo in adminsitration should be administration.


rachelwhitton:
+1 thanks @slashzero!

PR #715, opened 2015-07-13
Use table for unsupported-modules-plugins.md

### LINKED ISSUE #821: Unsupported/problematic modules should include why
Each module on the list should have an explanation of why it is problematic or unsupported. If the module has a related upstream issue, link to it.

https://pantheon.io/docs/articles/sites/code/unsupported-modules-plugins/

timani:
closes #821


bmackinney:
We'll need more content for the issue and solution columns in order to ship it. In the meantime we need to rebase and add a row for mobile tools in the Drupal seciton. Consider also linking to the module and its d.o or wordpress.org issue, if applicable.

ari-gold:
I looked at this PR locally and I'm seeing:


unsupported-table

It needs some work. I don't have time to dig into the formatting. @rachelwhitton do you? Would be great to get why plugins are unsupported out ASAP, as the state of the page right now is very incomplete. Maybe we can just add info without a table for now and add table later? /cc @timani @bmackinney

bmackinney:
A table will not add value here. Use H4's or labels for `Issue` and `Solution` (or `Alternatives`).

rachelwhitton:
Closing this PR in favor of #852

PR #737, opened 2015-07-21
Logs Landing Page

rachelwhitton:
Connects to #359 

- Create logs landing page
- Create logs directory for breadcrumbs
- Moves the following and updates links throughout:
 - source/docs/articles/sites/debugging-sites-with-log-files.md -> source/docs/articles/sites/logs/debugging-sites-with-log-files.md
 - source/docs/articles/sites/downloading-live-error-logs.md -> source/docs/articles/sites/logs/downloading-live-error-logs.md
 - source/docs/articles/sites/code/php-slow-log.md -> source/docs/articles/sites/logs/php-slow-log.md
 - source/docs/articles/sites/database/mysql-slow-log.md -> source/docs/articles/sites/logs/mysql-slow-log.md

Linkchecker has been run with all 404s resolved.

@bmackinney and @nataliejeremy can you review? Leaving the issue open - planning to add individual PR's for each article created.

Issue #359, opened 2015-04-17
Create docs for available logs files like the PHP slow log doc 


nataliejeremy:
Story: As a Developer I want additional docs for available logs files like the PHP slow log doc so I have more information about the performance of my site

- [x] Create `/docs/articles/sites/logs.md` and move php slow to  `/logs/php-slow-log.md`
- [ ] Add all relevant logs articles and stuff like Timani's Drupalcon presentation to the directory.
- [x] Create a new doc for each of the different logs in `/logs`
 - [x] newrelic.log - Summarized in logs.md
 - [x] nginx-acces.log - See [Parsing nginx Access Logs with GoAccess](https://pantheon.io/docs/articles/sites/logs/nginx-access-log/)
 - [x] nginx-error.log - Summarized in logs.md
 - [x] php-error.log - Existing doc not in the logs dir - [PHP Errors and Exceptions](https://pantheon.io/docs/articles/sites/php-errors-and-exceptions/)
 - [x] php-fpm-error.log - See [PHP Slow log](https://pantheon.io/docs/articles/sites/logs/php-slow-log/#analyze-the-php-fpm-error-log)
 - [x] php-slow.log - See [PHP Slow Log](https://pantheon.io/docs/articles/sites/logs/php-slow-log/#analyze-the-php-slow-log)
 - [x] pyinotify.log - Summarized in logs.md
 - [x] watcher.log - Summarized in logs.md
 - [x] mysqld-slow-query.log - See [MySQL Slow Log](https://pantheon.io/docs/articles/sites/logs/mysql-slow-log/)

rachelwhitton:
@timani can you create a draft for a doc from your Drupalcon LA presentation - Making your logs work for you: Drupal escalation and disaster recovery?

timani:
@rachelwhitton there is this outline but let me know what format you need, like an overview or guide. I can grab the notes from the slides too perhaps?

https://events.drupal.org/losangeles2015/sessions/making-your-logs-work-you-drupal-escalation-and-disaster-recovery

rachelwhitton:
@timani a full guide would be great! If you could organize a draft I can build it out. Thanks for the quick response!

PR #741, opened 2015-07-21
nginx access logs 

rachelwhitton:
Connects to #359

- Create new article: **Parsing the nginx Access Log File with GoAccess**

This PR should be reviewed/merged with #737 
@bmackinney @nataliejeremy


rachelwhitton:
@ari-gold can you review? Now that I realize access logs won't contain entries from traffic hitting Varnish this may not be a good topic for a doc. Any info you can provide on how nginx access logs are valuable to users would be awesome!

ari-gold:
Looks like it's still a useful doc. How about just adding a note that any requests served by Edge Cache will not hit nginx, so will not show up in nginx-access.log ? And if there's a separate doc can we delete the current answer to "How can I parse my Nginx access logs?" and just have the new answer be "See <Parsing nginx Access Logs with GoAccess>"

bmackinney:
LGTM :+1:

PR #745, opened 2015-07-22
Add version to stylesheet source to force reload

### LINKED ISSUE #744: Force Browser to clear cache for FF
See http://stackoverflow.com/questions/1922910/force-browser-to-clear-cache.

rachelwhitton:
Tested in Firefox

This update will create a new stylesheet source forcing browsers to reload. 

Closes #744 

@bmackinney please merge

### DIFF
--- a/app/config/sculpin_site.yml
+++ b/app/config/sculpin_site.yml
@@ -5,6 +5,7 @@ title: Pantheon Docs
 subtitle: Information for building, launching, and running dynamic sites on the Pantheon Website Management Platform
 url: /docs
 site_url: https://pantheon.io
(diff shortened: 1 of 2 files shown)


bmackinney:
why 1.x instead of 1.0? I think the point is to update the version number each time we make a change, meaning next time it will be 1.1, and so on.

rachelwhitton:
@bmackinney updated so that version is defined in `sculpin_site.yml` with `assets_version: 1` going forward when CSS changes are made, `asset_version` would need to be updated. You can use 1, 2, 3, etc. or 1, 1.1, 1.2, etc.

PR #753, opened 2015-07-23
Fix cron check failure output

westonruter:
Compare with https://github.com/pantheon-systems/wp_launch_check/blob/e7a861494a46af3195ecdcf5a1abba0a99fd3e68/php/pantheon/checks/cron.php#L34


nataliejeremy:
Thanks! @westonruter

PR #754, opened 2015-07-23
Fix formatting of robots.txt example

westonruter:


nataliejeremy:
@westonruter  Thanks for updating that!

Issue #751, opened 2015-07-23
Add incompatibility with Acquia modules and Pantheon Solr search


slashzero:
If you have acquia modules in your codebase (even disabled) and then install Pantheon's Apache Solr service you will not be able to connect to the solr server.

The solution is to disable Pantheon Solr, delete the acquia module files, clear the cache and then enable Pantheon Solr module.

PR #755, opened 2015-07-24
Added incompatibility with acquia search and Pantheon Apache Solr

slashzero:
Re https://github.com/pantheon-systems/documentation/issues/751


nataliejeremy:
Thanks for adding this! @slashzero

PR #262, opened 2015-03-04
Add syntax highlighting with highlight.js

bmackinney:
Merging this pull request will add syntax highlighting to preformatted code blocks with the [highlight.js](https://highlightjs.org/) library
This is a WIP because language detection seems off. 
This is the output at the configuring settings.php doc, which is accurate:
- [x] Identify all documents with syntax highlighting
- [x] Identify language tagging standards and add to the [style-guide.md](https://github.com/pantheon-systems/documentation/blob/syntax-highlight/style-guide.md)
- [x] Tag languages in code blocks according to style guide

PR #757, opened 2015-07-24
Styling alerts

rachelwhitton:
@nataliejeremy please review and provide feedback. 

Screenshots:
screen shot 2015-07-24 at 10 26 41 am
screen shot 2015-07-24 at 10 26 58 am

The style guide has been updated within this PR request.
`assets_version` updated to 1.2 to force CSS reload on all browsers
Going forward alert styling needs to wrap **Warning** or **Note** in H4 tags instead of strong tags.

### DIFF
--- a/app/config/sculpin_site.yml
+++ b/app/config/sculpin_site.yml
@@ -5,7 +5,7 @@ title: Pantheon Docs
 subtitle: Information for building, launching, and running dynamic sites on the Pantheon Website Management Platform
 url: /docs
 site_url: https://pantheon.io
-assets_version: 1
+assets_version: 1.2
 
 #######################################################################################################
 # The authors. Include your info here, under the "authors" key, using the following template:
--- a/source/docs/articles/drupal/cdn-setting-up-cloudfront.md
+++ b/source/docs/articles/drupal/cdn-setting-up-cloudfront.md
@@ -20,7 +20,8 @@ CloudFront is a pull-only content distribution network. All requests for assets 
 
 The first step in setting-up CloudFront on your Drupal site is to create a new CloudFront distribution. This article will help you create a barebones configuration. If you require a more complicated configuration, refer to the [AWS documentation](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/WorkingWithDownloadDistributions.html#DownloadDistValuesDomainName).
 <div class="alert alert-info">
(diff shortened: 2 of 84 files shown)


nataliejeremy:
@rachelwhitton Awesome!! Let's stick with blue for notes and red for warnings. Yellow is an issue for those who are color blind and it doesn't really convey a "warning" like using red will.

rachelwhitton:
@nataliejeremy updated, good idea to remove yellow!

screen shot 2015-07-24 at 10 38 47 am

nataliejeremy:
@informanddelight As the UX guru, can you +1 if this change improves readability and helps make the content in the notes scannable now? Here's what it currently looks like: 
screen shot 2015-07-24 at 9 48 33 am
screen shot 2015-07-24 at 9 48 06 am

informanddelight:
Notes are a bit more readable with the colored background. Could we actually do styling/use bootstrap alerts?

http://getbootstrap.com/components/#alerts

It'd just be a little easier extending what we're using in the product already.

nataliejeremy:
@informanddelight Aren't those for confirmations and warnings shown at the top of the page? These notes are spread throughout the docs, sometimes 3 or 4 on a page. We're trying to improve readability so they don't get ignored.

rachelwhitton:
Current styling on the live site is implemented with the bootstrap link you posted @informanddelight 
This PR is to replace the bootstrap styling with transparent background and black text. 

As far as aligning with product.. what if we made the notes look similar to what's on the status tab in the dashboard? 

Something like this..
screen shot 2015-07-24 at 12 31 26 pm

Lighter background and black text still accomplishes our objective of increasing readability (removing red text and red background that's currently live). 

@nataliejeremy and @informanddelight thoughts?

nataliejeremy:
+1 @rachelwhitton That looks good.

informanddelight:
True, it is more of a status/system alert rather than a note on article.

+1 Rachel, that balances making the contrast from the text color & background but also makes it a slightly different pattern from the system alerts. Nicely done!

rachelwhitton:
Updated screenshot:
screen shot 2015-07-24 at 12 47 05 pm

Thanks for weighing in @informanddelight! Much appreciated

rachelwhitton:
@informanddelight the black text is exclusively for code snippets, introduced in #262

bmackinney:
:+1: :ship:

PR #758, opened 2015-07-24
fixed path for simplesamlphp in way to be conform with download file …

pauloamgomes:
…and settings.php configuration


rachelwhitton:
Thanks @pauloamgomes for the update! Much appreciated

+1

PR #761, opened 2015-07-27
Remove reference to paid account for ssl

ronan:
I don't believe you need a paid account in Cloudflare for ssl and it contradicts this doc: https://pantheon.io/docs/guides/ssl-with-cloudflare/


nataliejeremy:
@ronan Thanks for the update!

Issue #762, opened 2015-07-27
Remove excess spacing and box at right and bottom of all articles


bmackinney:
@rachelwhitton @informanddelight I just noticed this bit of funk. 

![image](https://cloud.githubusercontent.com/assets/646061/8915826/21c9ed82-345c-11e5-8f3f-f5d4ef19770e.png)

rachelwhitton:
I'm not seeing the same thing @bmackinney - tested on safari, firefox, and chrome.

rachelwhitton:
The only funk I've noticed is that the breadcrumbs implementation generates an empty div container on the home page, which adds some padding above the search bar. 

Adding it to this issue for future CSS work:

<img width="1210" alt="screen shot 2015-07-28 at 9 37 14 am" src="https://cloud.githubusercontent.com/assets/10119525/8934263/73b0c764-350c-11e5-911e-24fe2f89ab8a.png">

bmackinney:
@rachelwhitton Seems to have resolved for me too.

bmackinney:
Or.. #764 will close this

PR #764, opened 2015-07-28
Remove overflow scroll to remove scrollbars from main section

timani:
Before removing overflow:
docs

After removing overflow:
after

Chrome Version 42.0.2311.152 (64-bit)
Linux


bmackinney:
closes #762

PR #765, opened 2015-07-29
Need a lasting place to reference for power users promotion

joshkoenig:
Would love to get this live so I can do another round of Power User outreach! :)


rachelwhitton:
Thank you @joshkoenig! Great write up!

PR #778, opened 2015-08-03
Update reading-pantheon-environment-configuration.md

davidneedham:
The code snippet for Domain Access was missing a leading forward slash before sites/all/modules/domain/settings.inc (which is in conflict with Domain access documentation). 

Drupal was interpreting this as /srv/bindings/foo/codesites/all/modules/domain/settings.inc. After this change it correctly comes out to /srv/bindings/foo/code/sites/all/modules/domain/settings.inc.


rachelwhitton:
+1 thank you @davidneedham! much appreciated!

PR #782, opened 2015-08-04
Implications for downgrading to Personal

ccorpus:
Added implications for SSL connections when downgrading to Personal. Please review @rachelwhitton @nataliejeremy


nataliejeremy:
Thanks! @ccorpus

PR #784, opened 2015-08-05
Add SagePay to IP-Address Based Security Schemes

slashzero:


rachelwhitton:
+1 thanks @slashzero Much appreciated!

I added issue #786 to document what was discussed in the CSE channel earlier today.

PR #787, opened 2015-08-05
Add note about committing Composer dependencies

westonruter:


nataliejeremy:
@westonruter  Thanks for updating that!

PR #791, opened 2015-08-07
Adding notes about the error_log function.

nstielau:
Not sure if this is helpful, or a distraction because it will be hard to get those logs in multiple application container situations.

cc @bmackinney


rachelwhitton:
+1 @nstielau - I think it's worth noting, even if it won't log to multiple app servers. Thanks for the update!

PR #793, opened 2015-08-08
Improve docs and method for installing WP Redis

westonruter:
It's not a good idea to move `object-cache.php` out of the WP Redis plugin because this breaks plugin upgrades.


nataliejeremy:
@westonruter Thanks for the update!

PR #820, opened 2015-08-17
Fix link to FileZilla project

kayakr:


nataliejeremy:
Thanks for the update! @kayakr

PR #837, opened 2015-08-20
Change title to help define Change Management

ari-gold:
Suggested by customer in https://pantheon-systems.desk.com/agent/case/43385


nataliejeremy:
@ari-gold I like it--thanks!

nataliejeremy:
@ari-gold I like it--thanks!

PR #839, opened 2015-08-22
Update url parameter naming details for "Considerations for Google Analytics and PANTHEON_STRIPPED"

angrytoast:
## What? Why?
Proposed documentation details around valid url query parameter naming to trigger `PANTHEON_STRIPPED` normalization on edge caching.

This is after testing on our current Pantheon site and confirming how param naming affects normalization.


bmackinney:
@ajberry can you review?

bmackinney:
@ajbarry can you review? (I pinged ajberry on accident earlier..)

ajbarry:
@angrytoast Thanks for the detective work, and your contribution to the docs. We've determined that this undesirable Varnish behavior so rather than document around it we're going to fix the issue with the cache keys.

I'm going to close out this PR rather than introduce a temporary alteration to the documentation.

Cheers!

angrytoast:
@ajbarry Thanks for the update. Any ETA as to when the Varnish behavior adjustment will go live?

PR #846, opened 2015-08-25
Delete add-cloudflare-free-ssl-to-wordpress-sites.md

timani:
We already have this other WP Cloudflare doc, do we still need this one?

https://pantheon.io/docs/guides/ssl-with-cloudflare/


nataliejeremy:
@bmackinney What do you think?

nataliejeremy:
@bmackinney I agree with Timani. The guide has Drupal and WP info and the WP specific doc isn't really needed.

atjuch:
So sad to see my one of my first docs going to a better place *sniff. Oh
well, more to write! :D

ari-gold:
@rachelwhitton Redirect needed?

nataliejeremy:
@ari-gold I added the redirect.

PR #848, opened 2015-08-26
Add warning that indexing can fail on large batch

slashzero:
https://pantheon-systems.desk.com/agent/case/43410


nataliejeremy:
Thanks for the update @slashzero

ari-gold:
What about the existing section "Check Index and Batch Sizes" ?

Issue #414, opened 2015-04-20
Document how to send HTTP headers that Varnish will care about 


nataliejeremy:
Story: As a website developers I want to understand how to send HTTP headers that Varnish will care about so that I can control my websites cache behavior.

Right now we have people setting cookies as a way to prevent pages from being cached, which is bass-ackwards. We should start teaching people about HTTP headers and how to set them to control downstream caching.

Problem: I want to control cacheability and/or cache-lifetime on a per-page basis
Soution: get smart about how to use the internal page cache, and HTTP headers for downstream (varnish) caching
About HTTP headers
How/why to use both the internal page cache and the downstream page cache
Customizing cache lifetime
Preventing a single page from caching

--submitted by @joshkoenig

bmackinney:
@nataliejeremy work with @timani to refine this issue and get it ready to write.

timani:
I stumbled upon this and just adding as a note: 

https://www.fastly.com/blog/best-practices-for-using-the-vary-header

Maybe having settings.php or wp-config.php examples but it would require some testing.

nataliejeremy:
@timani This is the one we're working on: https://github.com/pantheon-systems/documentation/pull/1079

PR #857, opened 2015-08-27
Add see also to Drupal Caching Settings

ari-gold:
- TODO: add link to WordPress article, once it's created


rachelwhitton:
@ari-gold are you suggesting a new doc on optimizing wp-cache or can this be done via #414 and #858? 

Between [Pantheon Plugins](https://pantheon.io/docs/articles/wordpress/pantheon-plugins/#pantheon-cache) and the suggested http headers doc, I think that covers everything. If not, we can merge this and create a new issue for the WP specific doc.

ari-gold:
I was envisioning a new doc for WordPress (#858) that would be equivalent of the Drupal doc.

rachelwhitton:
+1 

#858 has been updated to suggest new WP specific doc. 

Thanks!

PR #861, opened 2015-08-27
Adding suggestions from HD Made.

erik-pantheon:


nataliejeremy:
@erik-pantheon Thanks for the quick update!

PR #866, opened 2015-08-28
Info on access to custom upstreams

ccorpus:
The notice is to make developers aware that adding to custom upstreams is only available to partners - not Ally. I had a dev go through the trouble of creating a custom upstream only to learn later that it is only available to partners. He would have appreciated that detail in this doc before spending time creating a custom upstream.


ccorpus:
@rachelwhitton @nataliejeremy please review and approve. Thanks!

rachelwhitton:
+1 @ccorpus thanks for the PR! 

We have this documented in [Running a Custom Upstream on Pantheon](https://pantheon.io/docs/articles/organizations/running-a-custom-upstream/), but I agree it should be noted within this doc as well. 

@nataliejeremy can you review and +1?

nataliejeremy:
+1 Looks good--thanks.

PR #870, opened 2015-08-31
Responsive Docs

### LINKED ISSUE #853: Responsive Docs
The footer and nav elements need some cleaning up

- [x] Make footer responsive to media widths
- [x] Nav button does not show on mobile or when browser is less than 768px wide
- [x] Nav menu does not collapse at the correct width (making the menu split into 2 lines from 768px to 990px)
 - [x] Menu should resize automatically before collapse so lines are not split

rachelwhitton:
Closes #853 

@informanddelight @nataliejeremy @bmackinney please review and provide feedback. The only item removed within this PR is the CREATE FREE ACCOUNT primary button in the footer. Since our top nav bar is fixed, seems unnecessary to have the same CTA in the footer. 

# CSS 1.7 
### Absolute footer
**Add marketing icons to SEO links:**
screen shot 2015-08-31 at 1 31 57 pm
SEO links are set to `display:none` when width is below 995px to reduce clutter (IMO this improves UX when multi-tasking). **Copy right text is displayed on all pages, regardless of browser width.** Nav button styled so it's visible, menu is responsive w/smaller text based on media width:
screen shot 2015-08-31 at 2 27 57 pm
**Fixed footer (dark gray) on all pages excluding the docs home page.** CTA to join weekly training set to `display:none` below 725px:
screen_shot_2015-08-31_at_2_00_59_pm
### Search bar
**Combine search bar with TOC, fixed position so it scrolls with the page - excluding landing page and list pages** Responsive. 
Article example: 
screen_shot_2015-08-31_at_2_34_32_pm
List page example: 
screen shot 2015-08-31 at 2 45 40 pm
Responsive list page:
screen shot 2015-08-31 at 2 46 33 pm
### Breadcrumbs
**Combine breadcrumbs with nav** Not responsive, displays at all widths for easy navigation
screen shot 2015-08-31 at 2 48 13 pm

TODO: 
 - [ ] Add fixed position to sidenav (search docs) on list pages such as "All Articles", "Changelogs", "Guides", etc.


nataliejeremy:
+1 I love the responsive footer now--Huge improvement! :)

bmackinney:
:+1: . Let's wait for @informanddelight to comment before we ship.

informanddelight:
Is the search bar now in the right sidebar in all instances?  I like how
it's working towards making the information universally available but am
not a fan of the placement in with the TOC since it almost implies that the
search is parent to the table of contents for that page/article rather than
a universal/side-wide search function.

Also at a loss - are we just trying to take what we currently have in the
footer feel reduced/smaller. That seems like a fine change to me.

rachelwhitton:
@informanddelight thanks for the review! Much appreciated! Regarding the footer - in a way, yes. I'm hoping to make the footer less bulky. Here's how it looks on the live site at the moment when reduced: 
screen shot 2015-08-31 at 4 07 44 pm
I'm also removing non-essential containers (e.g. Live Training CTA and SEO Links for WordPress, Drupal, etc) at lower widths improve ux. In my experience I tend to have a doc open while I follow the steps in another window, the current state doesn't do that very well. 

Regarding the TOC: I had that thought as well. So I put the header "Search Docs" to indicate it was site wide, not parent to the TOC. Any suggestions for making that more clear? We could adjust the title, or move the search bar back to the breadcrumbs container and make it fixed with the top nav bar? Open to suggestions! 

Thanks again for the feedback

rachelwhitton:
ping @bmackinney - thoughts on having breadcrumbs and the search bar combined again and fixed under the nav bar? thinking that will accomplish what i was going for, which was to have the search bar available at all locations throughout the doc, no matter the position on the page.

nataliejeremy:
+1 for leaving the search at the top of the article (combined with breadcrumbs). That keeps it consistent throughout the docs site.

bmackinney:
As much as I would like constant access while in the docs, I don't think it's necessary, and defer to @informanddelight's opinion.

rachelwhitton:
@informanddelight can you take another look? 
## Breadcrumbs and Search
#### Menu
Container margins around the nav elements and color scheme now match panther site. Top menu position is fiexed, but the secondary nav menu with breadcrumbs and search is not. 
screen shot 2015-09-01 at 10 11 24 am
#### Current Page and Hover
The current page font is weight 200px to indicate it is the active element, and the hover functions work the same as the secondary nav on the panther site
screen shot 2015-09-01 at 10 16 32 am

informanddelight:
Awesome. Thanks for keeping the search with the secondary nav. Looks great!

Just to note, font-weight: 400 is "regular". Using 200 gets into the
"light" although we haven't implemented light so it would just default to
400.

bmackinney:
:ship:

PR #888, opened 2015-09-08
clarify TFA is the best authentication method

### LINKED ISSUE #372: Update recommendations on authentication methods
Story:
As a developer I want updated recommendations on authentication methods so that I can use best practices.

[David's comment on a recent PR](https://github.com/pantheon-systems/documentation/pull/302#issuecomment-85672363) states that LDAP is no longer best practice.

We need to:

 - [ ] Identify which docs require an update
 - [ ] Determine the best practice recommendations for authentication methods and SSO

nataliejeremy:
Closes #372 

@bmackinney @rachelwhitton Please review.

### DIFF
--- a/source/docs/articles/sites/code/ldap-and-ldaps.md
+++ b/source/docs/articles/sites/code/ldap-and-ldaps.md
@@ -10,9 +10,11 @@ keywords: ldap, ldaps, ldap and ldaps, using ldap as a provider, ldap provider, 
 
 ## LDAP as a Consumer  
 
-Using LDAP as a consumer of services is supported on the platform and will work at all plan levels, assuming correct configuration. The implementation and configuration details are up to the user as not all instances are supported. For general information about implementing LDAPS, see [https://drupal.org/node/1404368](https://drupal.org/node/1404368) and [https://drupal.org/node/1302032](https://drupal.org/node/1302032).
-
-LDAP provides access and maintenance of a distributed directory storing organized sets of records. LDAP is often used for single sign-on authentication and authentication requests will originate from Pantheon to the LDAP server used (outbound from Pantheon). However, we do not recommend using LDAP for single sign-on authentication as it is not secure.  The recommended approach is SAML (both for sites and the Dashboard) and OneLogin. Learn more about [two-factor authentication](/docs/guides/two-factor-authentication/).
+<div class="alert alert-info" role="alert">
+<h4>Note</h4>
+We do not recommend using LDAP for single sign-on authentication as it is not secure.  The recommended approach for sites and the Dashboard is to use SAML. For details, see <a href="https://pantheon.io/docs/guides/two-factor-authentication/"> Secure Your Site with Two-Factor Authentication</a>.
+</div>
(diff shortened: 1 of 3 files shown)


source/docs/articles/sites/code/sso-and-identity-federation.md
```diff
@@ -11,6 +11,11 @@ Pantheon’s flexible infrastructure does not restrict protocols or ports used f
 
 Use SSL certificates for encrypted secure communication with externally hosted servers for authentication.
 
+## Two-Factor Authentication
+Two-factor authentication (TFA) is a security practice that requires users of your website to provide, along with their standard username and password, an additional form of authentication to log in.
+
+We strongly recommend using SAML (both for sites and the Dashboard) and OneLogin. Learn more about [two-factor authentication](/docs/guides/two-factor-authentication/).
```
davidstrauss: We should strongly recommend using SAML but not OneLogin. There are multiple legitimate options for SAML IdP out there.

source/docs/articles/sites/code/ldap-and-ldaps.md
```diff
@@ -10,9 +10,9 @@ keywords: ldap, ldaps, ldap and ldaps, using ldap as a provider, ldap provider,
...
 
-LDAP provides access and maintenance of a distributed directory storing organized sets of records. LDAP is often used for single sign-on authentication, and many sites on Pantheon use LDAP and LDAPS for authentication. Authentication requests will originate from Pantheon to the LDAP server used (outbound from Pantheon).  
+LDAP provides access and maintenance of a distributed directory storing organized sets of records. LDAP is often used for single sign-on authentication and authentication requests will originate from Pantheon to the LDAP server used (outbound from Pantheon). However, we do not recommend using LDAP for single sign-on authentication as it is not secure.  The recommended approach is SAML (both for sites and the Dashboard) and OneLogin. Learn more about [two-factor authentication](/docs/guides/two-factor-authentication/).
```
rachelwhitton: @nataliejeremy what do you think of reducing this paragraph a bit and wrapping it in the alert styling? Something like this: 
screen shot 2015-09-08 at 3 29 24 pm

nataliejeremy: That gives the impression it's not OK to use, when it is. We just don't recommend it. @bmackinney what do you think?

rachelwhitton: @nataliejeremy I may be misunderstanding the [original comments](https://github.com/pantheon-systems/documentation/pull/302#issuecomment-85672363), but I don't think LDAP for SSO can be used at all unless it's in conjunction with a SaaS provider (such as onelogin). If so, we should update the paragraph to suggest they use SAML with a SaaS provider to bridge authentication over to LDAP. (Current verbiage doesn't mention bridging back to LDAP and mentions OneLogin directly). 

If a user is on this doc hoping to configure some type of SSO system with LDAP, I think a warning alert is fine. The verbiage used within the warning should be enough to inform the reader it doesn't apply to all implementations of LDAP, but just SSO. 

I defer to you and @bmackinney on that though

nataliejeremy: They would need an IdP provider to use LDAP. But that content already existed and I don't think we want to remove it or give them further instructions on using LDAP for SSO. IMO it doesn't warrant the warning treatment.

If I understood the issue correctly, we just want to make it clear we recommend SAML. Maybe @ari-gold or @bmackinney can clarify this.

nataliejeremy: @rachelwhitton Just updated to show the info as a note, per convo with @bmackinney

rachelwhitton:
Awesome, thanks @nataliejeremy! 

+1

PR #966, opened 2015-09-23
Add a simpler step for copying to clipboard

cauerego:
Well, it works at least if you're on a Mac. Based on http://apple.stackexchange.com/questions/15318/using-terminal-to-copy-a-file-to-clipboard


rachelwhitton:
Thanks for the PR @cauerego! Agreed that pbcopy is much easier for mac users

I suggested `pbcopy < ~/.ssh/id_rsa.pub` in place of `cat ~/.ssh/id_rsa.pub | pbcopy`

cauerego:
Awesome @rachelwhitton ! :100: 

:)

PR #969, opened 2015-09-23
Drush Alias update

atjuch:
This PR will address the common question asked about drush aliases being populated in a users site list if they fall under several conditions:

- They are an Administrator of an ORG
- They are not a Team Member on the site

@rachelwhitton and @bmackinney please review?

### DIFF
--- a/source/docs/articles/local/drupal-drush-command-line-utility.md
+++ b/source/docs/articles/local/drupal-drush-command-line-utility.md
@@ -75,7 +75,7 @@ $ drush sa
 ```
 <div class="alert alert-info">
 <h4>Note</h4>
-You must be a <a href="/docs/articles/sites/team-management/#team-management">team member</a> of the site for it to be included within your alias file. Organization administrators will not see all associated sites within their alias file, but will see sites for which they are team members.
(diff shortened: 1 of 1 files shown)


rachelwhitton:
@atjuch thanks! Moved this to a note in the **List Available Site Aliases** section, doesn't feel like a "limitation" but more of a platform specific note. Linked to team management doc. 

bmackinney:
Was this tested against 0.8.0? Reason: https://github.com/pantheon-systems/cli/issues/457#issuecomment-142677854

rachelwhitton:
@bmackinney good catch, hadn't seen that yet!

0.8.0 allows the user to run `terminus drush` commands on org sites, but 0.8.0 does not update their local alias file with all org sites. The note is accurate, but we can add content about using Terminus for running drush commands on sites where users are not direct team members. 

We could remove the note styling and just add it as content in the same section. Thoughts?

bmackinney:
Please add an issue to /cli for this.  

We should also consider adding a caveat link to all terminus commands in the repo that indicates it is is still pre-release and commands may change in the repo/not all functionality will work with all use cases.

rachelwhitton:
@bmackinney @nataliejeremy updated - pre-release caveat seems a bit out of scope for this PR so another issue has been created to address that. 

+1

PR #973, opened 2015-09-25
Fix to the "Getting Support" link.

twooten:
Removed the leading '/' which was causing the link to render as "https://pantheon.io/https://pantheon.io/docs/articles/getting-support"


nataliejeremy:

PR #1013, opened 2015-10-05
Create D8 Config Managment doc

### LINKED ISSUE #921: Create Drupal 8 Configuration Management article
- [x] Build the doc from Greg's blog post: https://pantheon.io/blog/configuration-workflow-drupal-8-sites
- [x] Link to post, link post to doc with notice that current information will be found here

Path: /docs/articles/drupal/

nataliejeremy:
Closes #921 
@bmackinney @rachelwhitton Please review.


nataliejeremy:
@greg-1-anderson Can you review this new doc? It was created based on your blog post. Thanks!

greg-1-anderson:
Hmm, I don't know.  Not to hate on my blog post or anything, but it would be much more compelling if we could document this based on the (not yet available) dashboard import / export buttons.   The local development workflow is less convenient; do we want to document it like this as an interim workflow, or hold off until the Pantheon workflow is available?

If we do go with this approach, the local development stuff should perhaps stated upfront, as a base assumption.  The blog was written as a generic article about Drupal, where local development is assumed.  On Pantheon, I think we should assume the multidev workflow is in use, and call out the difference when it isn't.

rachelwhitton:
+1 to adding an intro statement that warns the doc applies to local development but that a new Pantheon Dashboard "export configuration button" is coming soon for Drupal 8 sites. 

This as an interim doc seems okay while the export button is finalized. Unless we can expect that new dashboard feature with the D8 product release, is that the case @greg-1-anderson? If so, we should probably redraft to cover the new feature, then link to the blog for local development instructions.

greg-1-anderson:
The export button is currently not part of the D8 MVP, and is blocked on a UI review.  As far as the code is concerned, 'import' and 'export' work, but 'merge' needs work.  Not sure when this will be available; probably not too much after the D8 MVP.

greg-1-anderson:
Drupal core just renamed the "staging" configuration store to the "sync" configuration store. This document, and the blog post that it was based on, are both general enough that they somehow managed to avoid mentioning the "staging" store altogether. The repositories mentioned (example repo and config-extras) both need to be fixed to account for this now, though.

bmackinney:
@greg-1-anderson can you take another look?

greg-1-anderson:
Looks good to me!

PR #1017, opened 2015-10-06
Set cron job doing_wp_cron param to true.

zepner:
Client reports that Cron doesn't run unless the doing_wp_cron GET param is True.
https://pantheon-systems.desk.com/agent/case/45868


nataliejeremy:
Thanks! @zepner

PR #1022, opened 2015-10-07
update cloudflare known limitations

nataliejeremy:


source/docs/articles/sites/cloudflare-cdn-and-dns.md
```diff
@@ -52,5 +52,28 @@ In Drupal, no modules are required to use CloudFlare, but there is an optional [
...
+X-Cf-Url: http://www.domain.com/sites/all/modules/admin_menu/admin_menu.js?nvql2t
+X-Cf-Status: 403
+Content-Transfer-Encoding: binary
+X-Cf-Max-Age: 7200
+X-Cf-Origin-Status: 403
+X-Cf-Cache-Status: EXPIRED
+Content-Type: text/html; charset=UTF-8
+X-Cf-Error: true
+```
+
```
rachelwhitton: @nataliejeremy should we add that the solution to this error is to disable the Rocket Loader feature in CloudFlare?

nataliejeremy: @rachelwhitton +1 if that's the best solution/suggestion

rachelwhitton: @natalie - Suggestion based on the [linked CF doc](https://support.cloudflare.com/hc/en-us/articles/200169456-Why-is-JavaScript-or-jQuery-not-working-on-my-site) - so the user doesn't need to load a new doc/site to find a solution. 

Looking again, we probably don't need to include the CF link at all - it just says to disable the feature or to report the bug to CF engineers.

I defer to you and @bmackinney for final edits -- :+1: from me!

nataliejeremy: Thanks--sounds good to me @rachelwhitton

PR #1024, opened 2015-10-07
Added connection method information

newtoid:
Suggested in support case 46399 as user was attempting connection via SSH


rachelwhitton:
Thanks @newtoid! 

+1

PR #1028, opened 2015-10-10
More information about SendGrid

kappaluppa:
Because SendGrid does not automatically approve everyone for use of their services there could be a delay in setting up email. This could be a real blocker for a newly launched site if email is critical, such as a member registration site. I did not know this and my launch is now being held up.


rachelwhitton:
+1

Thanks for the contribution @kappaluppa! 

@nataliejeremy can you review? Thinking we might want to have the new info styled in our note alert class, but I defer to you on that. Thanks!

PR #1041, opened 2015-10-15
Correct "Creating a services.yml File for Drupal 8"

### LINKED ISSUE #1039: services.yml documentation needs to be reworked.
The advice I provided on the global services.yml file was not accurate; the purpose of this file is different than I supposed, and this difference has a profound impact on how this file should be managed.  @rachel.jaro pointed out that sites/*/services.yml is in our .gitignore file (inherited from Drupal 8's example.gitignore file), and @ari-gold provided the Drupal issue that explains this:

https://www.drupal.org/node/2251113

I had not read this background information previously, so I did not realize that the purpose of this file is to store environment-specific settings, such as twig debug mode. These settings should *not* be committed to the git repository, and should *not* migrate from one environment to another when pushing code.

The [services.yml documentation](https://pantheon.io/docs/articles/drupal/8/create-services-yml-file/) should be rewritten to advise users to manage this file in SFTP mode, and to use features of their SFTP client to make it unmodifiable once they are done.  It should remain in the .gitignore file, so that it is not deployed to other environments.  The existing reference to module-specific services.yml files should be maintained.

rachelwhitton:
source/docs/articles/drupal/8/create-services-yml-file.md
Closes #1039

- Remove recommendations to create/modify services.yml via Git, as the file is ignored and should not be migrated between environments when pushing code.
- Redraft includes warning note stating why Git cannot be used.
- Link to resource on environment-configurable settings issue on drupal.org
- Link to SFTP doc for general instructions on modifying files via SFTP
- Move module note to more appropriate section (file creation step)

@greg-1-anderson and @bmackinney can you review?

### DIFF
--- a/source/docs/articles/drupal/8/create-services-yml-file.md
+++ b/source/docs/articles/drupal/8/create-services-yml-file.md
@@ -6,20 +6,25 @@ category:
   - drupal
 keywords: drupal, services.yml, yml file
 ---
-Drupal 8 enables developers to define services used by core within the `sites/default/services.yml` configuration file, eliminating the need for hacks to swap functionality. Creating or modifying this file is not required when installing a new Drupal 8 site.
+Drupal 8 allows users to easily define core services and environment-specific settings within the `sites/default/services.yml` configuration file, eliminating the need for hacks to swap functionality. Creating or modifying this file is not required when installing a new Drupal 8 site.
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Looks good.  I would say "should not be created" instead of "cannot be created", because it is physically possible to put the file into the Git repository; it would cause problems to do so, though, as the documentation already correctly explains.

rachelwhitton:
Thanks for the issue and review @greg-1-anderson! I rephrased the warning note

nataliejeremy:
+1

PR #1051, opened 2015-10-20
deploy-multidev.sh revision - single file deployments

### LINKED ISSUE #1049: rsync - argument for filepath
modify `deploy-multidev.sh` so that file path can be specified if all docs do not need to be deployed.

rachelwhitton:
Revise deploy-multidev.sh so that work is not overwritten on the multidev test environment
 when QA'ing deployments.

Closes #1049

- Ask user for file path
- Check file path, if it does not exist - kill script.
- No backups, log file created locally but no
t sent to remote filesystem

### DIFF
--- a/scripts/deploy-multidev.sh
+++ b/scripts/deploy-multidev.sh
@@ -4,23 +4,24 @@
 # Define env and site UUID
 export ENV=docs-files
 export SITE=72e163bd-0054-4332-8bf8-219c50b78581
-
-
-# Generate docs production files
-rm -rf output_prod/docs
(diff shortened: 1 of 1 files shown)


rachelwhitton:
+1 thanks for editing @bmackinney!

bmackinney:
Thinking maybe we should remove the yes/no and default to prompt for directory/doc always. Can use another script for reset/catch up to master. thoughts @rachelwhitton ?

rachelwhitton:
@bmackinney updated, ready for you to review/test.

PR #1054, opened 2015-10-22
replacing base_path with base_url for consistency

### LINKED ISSUE #1053: Replace Links Referencing IP:PORT in Drupal Doc Update
Re: [Replace Links Referencing IP:PORT in Drupal](https://pantheon.io/docs/articles/drupal/fix-broken-links-that-reference-ip-port-instead-of-domain-name)
## Issue Description:
Doc uses  `$base_url` and `$base_path` interchangeably, which is a bit confusing. 
## Suggested Resolution 
Use `$base_url` exclusively (`$base_path` is used to set path for core - not covered in this doc).

hengkit:
Closes #1053


rachelwhitton:
Thanks @hengkit!

+1

PR #1056, opened 2015-10-22
Adds additional GA parameter anti-patterns

ajbarry:
Warns users about code generated redirects.
Warns about overloading GA parameter namespace.
Placeholder for curl/wget tests.


bmackinney:
:rocket: :+1:

PR #1072, opened 2015-10-28
Fix terminus backups get command --to param

tauno:
Previously --to-directory, now appears to be --to


nataliejeremy:
Thanks for the update! @tauno
@slashzero Please review

slashzero:
Confirmed as correct.

nataliejeremy:
Thanks for the update! @tauno

PR #1073, opened 2015-10-28
Document cookie-stripping on URLs ending in .html and .htm

jeffesquivels:
Editing the documentation on debugging the cache, to include in the Cookies & Varnish section the information regarding how cookies are stripped on URLs that end in .html and .htm (as per this conversation[0]) so that anyone experiencing this behaviour know it's expected and not a malfunction. Also adding the recommended way to deal with the scenario where a .html or .htm extension must be used to satisfy backwards compatibility with an old site that used to have those extensions (as stated here[1]).

[0] https://twitter.com/DavidStrauss/status/659063304636116992
[1] https://twitter.com/DavidStrauss/status/659064986778537984


nataliejeremy:
Thanks for updating that! @jeffesquivels

jeffesquivels:
@nataliejeremy you're welcome!

nataliejeremy:
@jeffesquivels Our engineering team has created a fix for this, so the new content isn't needed anymore. :) I'll go ahead and remove it. Thanks for contributing!

jeffesquivels:
@nataliejeremy that's great to hear (read? :), thanks!

PR #1075, opened 2015-10-29
Update filezilla.md

ttrowell:
Users need to use an alternate ftp client when uploading files.  Transmit works for Mac, but we also should probably identify one that works on a PC - if anyone has one that we can test on!
Also, my formatting stinks.  Help!  @nataliejeremy


nataliejeremy:
Thanks for the update @ttrowell

PR #1080, opened 2015-11-03
Add google tag manager to default.html

rachelwhitton:
ping @sosnovka


sosnovka:
Hi Rachel, it needs to be immediately after the opening <body> tag according to google:
https://developers.google.com/tag-manager/quickstart


Just like on pantheon.io :
view-source:https://pantheon.io/agencies

Best,
Nikita

rachelwhitton:
Thanks @sosnovka - must have missed that link before! Updated.

PR #1081, opened 2015-11-03
Update reading-pantheon-environment-configuration.md

ttrowell:
I keep needing to compare these values against what applications are using.  I am proposing that we expose it to clients.  Please review.  @bmackinney @nataliejeremy 
Thanks!

### DIFF
--- a/source/docs/articles/sites/code/reading-pantheon-environment-configuration.md
+++ b/source/docs/articles/sites/code/reading-pantheon-environment-configuration.md
@@ -41,6 +41,148 @@ Database credentials, Redis authentication, and other configuration data is prov
 
 If you are using a common CMS framework, the code you need to load this configuration and boot your app should already be pre-configured. However, if you need to do something custom, you can work with environmental configuration directly.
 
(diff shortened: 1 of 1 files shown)


bmackinney:
I like it. @joshkoenig is this :+1:  to share?

joshkoenig:
In general I support this. However:

1. Let's encourage people to use `$_ENV` rather than `$_SERVER` as that will be around on the CLI too.
2. We should tell people to be cautious about this like they would leaving a stay `phpinfo()` file out there. This contains sensitive data!

ttrowell:
Clarifications and warnings added to the document for review @nataliejeremy .  Thanks!

nataliejeremy:
@joshkoenig Can you take one last look? It's been updated. Thanks!

joshkoenig:
This still uses `$_SERVER`. Change to `$_ENV` and it should be good to go!

rachelwhitton:
Thanks @joshkoenig - can you review c14119d? I replaced `$_SERVER` with `$_ENV` in @ttrowell's contribution as well as throughout the doc.

joshkoenig:
Yup looks good!

PR #1085, opened 2015-11-05
Update migrate.md

zepner:


nataliejeremy:
Thanks for adding that @zepner

PR #1089, opened 2015-11-09
Remove duplicated sentence

twooten:


nataliejeremy:
Thanks for updating that! @twooten

PR #1090, opened 2015-11-09
fix dead link

sukottokun:


bmackinney:
This shouldn't be dead. We should get to the bottom of where it went before replacing

nataliejeremy:
I tried it and http://varnishcheck.pantheon.io and neither worked. Do you not agree with Scott's alternative?

sukottokun:
Natalie, Brian and I spoke about it, after submitting the PR, I debated
submitting another. I vote for removing the link to any varnish checking
site.

PR #1104, opened 2015-11-11
Update drupal-drush-command-line-utility.md

kappaluppa:
With regards to the settings.php file, these instructions only give half the story. We know we need to have the settings.php file added so we can use Drush, but there is no info on how to go about getting the file uploaded:
Switch to SFTP mode and add the file. You can find your code in the ~/code directory once you connect.
https://pantheon.io/docs/articles/sites/code/developing-directly-with-sftp-mode/


rachelwhitton:
Thanks for the PR @kappaluppa!

PR #1116, opened 2015-11-16
Cron doc: remove conflicting/confusing note

### LINKED ISSUE #1034: Cron for Drupal Doc Update
Re: [Cron for Drupal](https://pantheon.io/docs/articles/drupal/cron)
## Issue Description:

Under "Pantheon Cron Execution", the documentation says:

>Note
>
>Setting the value to "Never" will be ignored; cron will always run at least hourly.

Under "Troubleshooting Cron" the documentation says:

>Can I prevent Drupal Cron from running?
>
>Yes, from within Drupal at admin/system/config/cron, select Never from the "Run cron every" drop-down menu, then save the configuration. 

## Suggested Resolution 

Figure out which statement is correct, and fix the other one

nataliejeremy:
Closes #1034 
@newtoid Please review and +1 the change. Thanks!

### DIFF
--- a/source/docs/articles/drupal/cron.md
+++ b/source/docs/articles/drupal/cron.md
@@ -18,10 +18,7 @@ For every site environment, Pantheon executes cron at the top of each hour to al
 This bootstraps your site and invokes [drupal\_cron\_run](https://api.drupal.org/api/drupal/includes!common.inc/function/drupal_cron_run/7).
 
 There is no way to configure when Pantheon executes Drupal cron.
-<div class="alert alert-info" role="alert">
-<h4>Note</h4>
-Setting the value to "Never" will be ignored; cron will always run at least hourly.
-</div>
+
 ## Managing Cron
 
 You can manage cron via Drupal's admin interface at `admin/config/system/cron`.
--- a/source/docs/articles/drupal/cron.md
+++ b/source/docs/articles/drupal/cron.md
@@ -17,7 +17,12 @@ For every site environment, Pantheon executes cron at the top of each hour to al
 
 This bootstraps your site and invokes [drupal\_cron\_run](https://api.drupal.org/api/drupal/includes!common.inc/function/drupal_cron_run/7).
 
-There is no way to configure when Pantheon executes Drupal cron.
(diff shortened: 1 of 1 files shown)


source/docs/articles/drupal/cron.md
```diff
@@ -18,10 +18,7 @@ For every site environment, Pantheon executes cron at the top of each hour to al
 This bootstraps your site and invokes [drupal\_cron\_run](https://api.drupal.org/api/drupal/includes!common.inc/function/drupal_cron_run/7).
 
 There is no way to configure when Pantheon executes Drupal cron.
-<div class="alert alert-info" role="alert">
-<h4>Note</h4>
-Setting the value to "Never" will be ignored; cron will always run at least hourly.
-</div>
+
```
rachelwhitton: The text in line 20 seems a little confusing given the FAQ on line 84
>  There is no way to configure when Pantheon executes Drupal cron.

Suggest a revision which states there's no a way to configure or disable Pantheon's system cron execution, but Drupal cron can be disabled from within `admin/system/config/cron` 

^ CSE and/or @newtoid should confirm that first though.

nataliejeremy: This info here is accurate and under the header "Pantheon Cron Execution". They cannot configure Pantheon's cron at all; only Drupal cron. @newtoid can confirm.

source/docs/articles/drupal/cron.md
```diff
@@ -18,10 +18,7 @@ For every site environment, Pantheon executes cron at the top of each hour to al
...
 ## Managing Cron
 
 You can manage cron via Drupal's admin interface at `admin/config/system/cron`.
```
newtoid: Platform cron cannot be disabled and will run Drupal cron every hour even if cron is disabled in the application. Using a module like Elysia Cron will allow more granular control of what happens when Drupal cron runs.

nataliejeremy:
@bmackinney can you respond to Rachel's question and +1 David's response? Thanks.

bmackinney:
No, I can't +1 it because I don't know/haven't tried. If it is, then the paragraph that was deleted needs to be restored, and the other conflicting statement needs to be fixed.

nataliejeremy:
@rachelwhitton @newtoid I incorporated your feedback...can you take another look?

rachelwhitton:
+1 - looks good to me, thanks @nataliejeremy!

PR #1134, opened 2015-11-23
Feature testing

craychee:
This PR builds on @rachelwhitton's awesome work with behat.

It adds:
- A grunt task that runs [a11y](https://github.com/addyosmani/a11y) testing.
- A Rake task that runs [htmlproofer](https://github.com/gjtorikian/html-proofer).
- A script to look for remnants of a bad merge.
- It swaps out the Selenium Driver for the (faster, easier to manage) Ghost Driver.
- Option (preferred) to work on documentation locally with vagrant.

TODO:
- The regression tests run by behat need better definition. (These are @rachelwhitton's first pass.)
- Get feedback on testing made available here and iterate over it.


README.md
```diff
@@ -25,32 +25,45 @@ documentation.
...
+
+### Option 1: Use Vagrant (preferred)
+1. Install requirements:
+* [virtualBox](https://www.virtualbox.org/wiki/Downloads) >= 4.3.x
+* [ansible](http://docs.ansible.com/ansible/intro_installation.html#installing-the-control-machine) >= 1.8.x
+* [vagrant-hostmanager](https://github.com/smdahlen/vagrant-hostmanager)
+2. From inside the project root, run `vagrant up`
+3. You will be prompted for the administration password on your host machine. Obey.
+4. Visit <http://docs.local:8000/docs> in your browser of choice.
+
```
rachelwhitton: @craychee I'm not able to complete using Option 1

I get the following output from `vagrant up`: https://gist.github.com/rachelwhitton/6050a769a520481dc746#file-vagrant_up

Virtualbox version 5.0.2
ansible 1.9.4
vagrant-hostmanager 1.6.1

Any tips?

craychee:
@rachelwhitton `vagrant plugin install vagrant-auto_network` will solve the issue.
I updated the `Vagrantfile` and the README to reflect this new requirement.

rachelwhitton:
@craychee thanks, I installed `vagrant-auto_network` and I'm getting the following output for `vagrant up`:
```
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

hostname -f

Stdout from the command:



Stderr from the command:

sudo: unable to resolve host vagrant-ubuntu-trusty-64
stdin: is not a tty
hostname: Name or service not known
```

Full output of `vagrant up`: https://gist.github.com/rachelwhitton/62a51950794eb0119e54

I'm not receiving a prompt for my password like before.

craychee:
@rachelwhitton following the `hostname: Name or service not known vagrant` issue, I just made a change that will (I hope) resolve this issue for you. If not, we can just remove that line for now.

rachelwhitton:
Just pulled this down again and tested, had to install the vagrant-hostmanager plugin again because I believe we uninstalled it in our troubleshooting. Unfortunately, vagrant up still produces the same error :/ 


```
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

hostname -f

Stdout from the command:



Stderr from the command:

sudo: unable to resolve host vagrant-ubuntu-trusty-64
stdin: is not a tty
hostname: Name or service not known
```

craychee:
@rachelwhitton great! I changed the `composer.json` to put executables inside "bin". I hadn't thought about this but that means if you are building from an existing project, composer doesn't know to rebuild your executables. I don't have a graceful way around this other than to blow away `vendor`. Once you do, run `vagrant provision` to re-run the provisioning.

craychee:
@rachelwhitton I updated the basebox here so you can run the tests locally after building the vagrant box. Please pull down changes that I made to this branch, destroy your environment (`vagrant destroy`), and run `vagrant up` again. You should be able to log in (`vagrant ssh`), navigate to the project (`cd /vagrant`) and run the tests (`rake` and `grunt`, along with behat).

Please update this thread too when you get the chance.

rachelwhitton:
@craychee thank you!! Everything seems to be working as expected - I can access the docs site at http://docs.local:8000/docs and run tests on the local environment. I do have a few questions: 

**1. Do you have tool recommendations for reviewing the audit/test results outside the terminal?**

**2. Can you validate that the output provided is expected?**
 _Curious about the `rake aborted!` occurrence, in addition to using `grunt --force`?_

 - **HTML-Proofer output, `rake`:** 
 ```
 rake aborted!
 HTML-Proofer found 824 failures!
 /var/lib/gems/2.0.0/gems/html-proofer-2.5.2/lib/html/proofer.rb:199:in `print_failed_tests'
 /var/lib/gems/2.0.0/gems/html-proofer-2.5.2/lib/html/proofer.rb:100:in `run'
 /vagrant/Rakefile:8:in `block in <top (required)>'
 Tasks: TOP => default => htmlproofer
 (See full trace by running task with --trace)
 ```
 (This is not the entire output, just the end)

 - **Output from a11y:dev testing, `grunt`:**
 ```
 Running "a11y:dev" (a11y) task
 Warning: spawn ENOMEM Use --force to continue.
 
 Aborted due to warnings.
 ```
 
 `grunt --force` is a bit of a slow moving test, but seems to be working as expected (still running an audit but has not moved forward in a few minutes)

craychee:
@rachelwhitton 
I can.
1. I don't, I am afraid. The tools here are the same tools that circleci can use.
2. 
a.) Rake by default will stop on a failure. Since html-proofer is failing, rake halts the run of additional tasks. Fortunately, we do not have other tasks. I could allow failures but there isn't really a use case at the moment. It's better that we either resolve html-proofer's issues or we configure html-proofer to ignore errors that you and @bmackinney agree to allow.
b.) the current a11y run is enormous. You might occasionally get that error because there are so many errors, the system runs out of space. \o/ Better to clean up the errors or just focus on running the suite against a limited part of the code. (Again, this is something you and @bmackinney should decide.)

Vagrantfile
```diff
@@ -0,0 +1,37 @@
...
+
+  config.ssh.forward_agent = true
+  config.ssh.insert_key = false
+
+  config.vm.provision :shell, inline: <<SCRIPT
+  set -ex
+  phantomjs --webdriver=8643 &> /dev/null &
+  su vagrant -c 'cd #{path} && composer install && bundle && npm install;
+  bin/sculpin generate && ln -sf #{path}/source #{path}/output_dev/.
+  bin/sculpin serve &
```
rachelwhitton: @craychee - can we adjust the sculpin script so that the `--watch` flag is included? I see that the project dir is synced to the vm, but sculpin isn't watching for changes and regenerating. I've tried it on Line 33 and 34, but saw some strange behavior from `vagrant up` and `vagrant provision` after adding the flag.

craychee: @rachelwhitton you can do that. You would need to remove the `&` and the `disown`. If change this to `bin/sculpin serve --watch`, vagrant will just hang. This creates two problems:
1. You can't `vagrant ssh` into the machine in the same command window since it's a rolling log. Instead, you would need to open a new window.
2. If you kill the output (via ctrl-c), you won't be able to halt the vagrant machine because although you killed the output, you didn't kill the process. You would need to ssh in to kill the process before shutting down.

My recommendation is to leave it as is but update the docs: `vagrant up` will build the site for you. To work on it, you should `vagrant ssh` in, navigate to `/vagrant` and run `bin/sculpin serve --watch`. This is more aligned with vagrant's usefulness anyway: it's really good at building an environment but isn't intended to have an opinion about how you actively develop on it.

rachelwhitton:
@craychee - I added wraith as a rake task and left htmlproofer as the default, thinking circle would just run the default.. but circle failed on my last commit after rake aborted due to a wraith error so that assumption was not correct. 

I can remove the wraith task and update the PR, just wanted to ping you for review/feedback first.

craychee:
@rachelwhitton circle is throwing an error because wraith needs to be installed (its part of the requirements).
However, I don't think using [wraith's Rakefile](https://github.com/BBC-News/wraith/blob/master/Rakefile) is a good fit for us. We will want to configure our own workflow.

rachelwhitton:
Thanks @craychee! Mind taking another look? 

I think we're getting close, but I'm not sure about the behat definitions, @bmackinney can you think of additional scenarios we should test? Can we set aside some time to go through the [results](https://circleci.com/gh/pantheon-systems/documentation/3289) and strategize a debugging task?

PR #1123, opened 2015-11-18
Adding Pantheon WP Varnish Caching article and screenshot. - EHM

rachelwhitton:
Restore prematurely merged PR #1115 

Closes #1061

PR #1135, opened 2015-11-23
WordPress Pantheon Cache Plugin Configuration - update

### LINKED ISSUE #1061: How WP caches are cleared at the edge
WP cache clears differently than drupal, and it affects how we rate limit cache clears at the edge. We should edit the existing doc, [WordPress Pantheon Cache Plugin Configuration](https://pantheon.io/docs/articles/wordpress/wordpress-pantheon-cache-plugin-configuration/) so CSE and users understand this and provide workarounds for cases like: https://pantheon-systems.desk.com/agent/case/46771 Nick says Sara can assist.

Todo: 

- [ ]  Using the info present in #1123 create a PR to the existing doc
- [ ]  Test 0 seconds settings, note from Ari below
 - Test using curl, as explained in [Testing varnish with curl](https://pantheon.io/docs/articles/sites/varnish/testing-varnish/#test-varnish-with-curl)

Note from Ari:
Have you tested 0 seconds? I don't think that works currently. I think current behavior is wonky, allowing minimum of 60 seconds, and if you try to set it to less than 60 seconds it will set it the TTL to 600 seconds. See current version https://github.com/pantheon-systems/WordPress/blob/6b1eefb5269a266cca3db8ceb6372899e844a41c/wp-content/mu-plugins/pantheon/pantheon-cache.php#L204

Work being done on release-4.4 branch that will change this behavior: pantheon-systems/WordPress#34

nataliejeremy:
Closes #1061 
This PR will replace and close PR #1123 as the content from both docs as been combined into the existing WP Cache doc.
I tested per Ari's comment and you cannot set Pantheon Cache to 0. If you try, it changes it to 600. I removed that content from the previous PR.

@erik-pantheon @rachelwhitton @ari-gold Please review.

### DIFF
--- a/source/docs/articles/wordpress/wordpress-pantheon-cache-plugin-configuration.md
+++ b/source/docs/articles/wordpress/wordpress-pantheon-cache-plugin-configuration.md
@@ -26,42 +26,42 @@ You'll want to strike a balance between freshness of content and speed. We recom
 
 Within the `pantheon-cache.php` file that houses the Pantheon Cache plugin code, there are three functions that are useful to developers. You can call them from within your own custom code using various WordPress hooks, such as [save_post()](https://codex.wordpress.org/Plugin_API/Action_Reference/save_post). Currently, the [limit on the number of paths](https://github.com/pantheon-systems/WordPress/issues/24) that can be cleared in a single call is 10.
 
-**flush_site**  
+### flush_site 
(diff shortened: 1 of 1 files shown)


rachelwhitton:
+1 Looks great! Thanks @nataliejeremy - I made a few edits in commit 3083db2

PR #1148, opened 2015-12-03
Update redirect-incoming-requests.md

jaypan:
The documentation for redirecting to HTTPs was incorrect, as it was checking if HTTPS was 'ON', when it should be checking if HTTPS is 'off'.

The documentation for redirecting specific pages to https may also be incorrect - I only checked the site-wide redirect.


nataliejeremy:
Thanks for the update! @jaypan

newtoid:
+1

rachelwhitton:
Thanks @jaypan - much appreciated!

timani:
+1

PR #1151, opened 2015-12-03
Update drush-versions.md

ttrowell:
The section on using a policy file locally is confusing.
These edits try to clarify if the default is Drush 5.x, but the Drush 8.x part is also cryptic IMHO.

This PR is technically not ready.  :) Drush 8 needs to also be addressed.  @rachelwhitton

### DIFF
--- a/source/docs/articles/local/drush-versions.md
+++ b/source/docs/articles/local/drush-versions.md
@@ -28,7 +28,7 @@ For sites running 5.x, create a file called `policy.drush.inc` and place it in t
 The Pantheon platform executes hooks as they are written. If a policy file has incorrect logic (e.g. using Drush 7 with Drupal 8), the operation will fail.
 </div>
 
-The following example uses Drush 8 instead of 5.x for targets on the Pantheon platform. Our `hook_drush_sitealias_alter` function looks like this:
+The following example uses Drush 8 instead of 5.x for targets on the Pantheon platform. Our `hook_drush_sitealias_alter` function within the `policy.drush.inc` file looks like this:
 ```php
 function policy_drush_sitealias_alter(&$alias_record) {
   // Fix pantheon aliases!
@@ -38,7 +38,7 @@ function policy_drush_sitealias_alter(&$alias_record) {
   }
 }
 ```
-Commit the policy file to your site's repository and clear Drush cache by running `drush cc drush`, then check Drush version:
+Save the file and clear Drush cache by running `drush cc drush`, then check Drush version:
 ```
 $ drush @pantheon.SITENAME.ENV version
 drush version 8.x


rachelwhitton:
Thanks @ttrowell! 

@greg-1-anderson should we add an example hook for changing the remote drush version to 5.x for sites currently running Drush 8.x on Pantheon? If so, can you provide it? [Existing example](https://pantheon.io/docs/articles/local/drush-versions/#current-default-is-5.x) is for changing remote drush to 8.x for sites running 5.x

greg-1-anderson:
If you change your remote Drush to anything less than 8.x for Drupal 8.x sites, then Drush will stop working; only Drush 8 supports Drupal 8.

rachelwhitton:
@greg-1-anderson right, thanks for the info on D8 - forgot that it requires Drush8! How about new D7 sites created on Pantheon? Default for those are Drush8, correct? Would it be possible, or would it make sense, for those users to change the remote version to something else, like Drush 5?

greg-1-anderson:
Drush 5 has not been supported for more than a year. Drush 6 was never fully rolled out on Pantheon; we are still running Drush 6.2; the most recent version is 6.7, and Drush 6 is also EOL.  Moshe calls Drush 7 "the ignorable release" -- it is supported, but very similar to Drush 8. The one main difference is that Drush 7 works with PHP 5.3, but that should not be an issue for new Pantheon sites, which will be running  PHP 5.5.

I would tend to think we should not encourage folks in the docs to go back to old versions for new sites. If you have any place in the docs where you tell people to use PHP 5.3 (e.g. importing an old site with legacy code, maybe?), then maybe at that spot point them to this doc and say "put in 'drush7' instead of 'drush8' to use a version of Drush that works with PHP 5.3.

If you think this is at all a common scenario, then perhaps the platform should automatically select Drush 7 instead of Drush 8 when  PHP 5.3 is selected.

cc @joshkoenig

rachelwhitton:
@bmackinney what do you think about documenting incompatibility with default drush version (Drush8) for D7/6 sites migrating to the platform which require PHP 5.3? See comments from Greg above. Curious what doc we should add it to?

joshkoenig:
> If you think this is at all a common scenario, then perhaps the platform should automatically select Drush 7 instead of Drush 8 when PHP 5.3 is selected.

We have an active bug to work on this I believe.

source/docs/articles/local/drush-versions.md
```diff
@@ -38,7 +38,7 @@ function policy_drush_sitealias_alter(&$alias_record) {
   }
 }
 ```
-Commit the policy file to your site's repository and clear Drush cache by running `drush cc drush`, then check Drush version:
+Save the file and clear Drush cache by running `drush cc drush`, then check Drush version:
```
bmackinney: Is this clearer? maybe we say "save and commit"?

greg-1-anderson: The policy file is stored locally, usually in `~/.drush/policy.drush.inc` or similar; it is usually not committed, and when committed, it would be in some repository other than the Pantheon site repository.

greg-1-anderson:
LGTM.   +1

PR #1160, opened 2015-12-09

ttrowell:
This PR adds a possible solution to using Views Data Export.
@timani Please review for @rachelwhitton 
Thanks!

### DIFF
--- a/source/docs/articles/sites/code/unsupported-modules-plugins.md
+++ b/source/docs/articles/sites/code/unsupported-modules-plugins.md
@@ -120,7 +120,7 @@ Additionally, the variable can be set using drush:
 ````
 drush vset views_data_export_directory 'public://'
 ````
-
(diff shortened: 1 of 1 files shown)


timani:
I think if we add the note Sheldon put in about using the DB streamwrapper for edge cases with a disclaimer it may/may not impact performance and I think that should do it. 

As is this will take care of 95% of all users but there are about 5% who may need the DB streamwrapper (it is a sandbox project but a report of success)

bmackinney:
@timani do you have the link?

timani:
https://www.drupal.org/sandbox/jim/2352733

ari-gold:
I like adding a Solution section. Should the link to the patch be moved to the solution section? Do we need multiple solutions?

bmackinney:
@timani good enough?

timani:
LGTM

PR #1163, opened 2015-12-10
Drush limitation with value of ['uri'] in pantheon.aliases.drushrc.php

timani:
Sanitized the info from the commands, the formatting may need some work.

### DIFF
--- a/source/docs/articles/local/drupal-drush-command-line-utility.md
+++ b/source/docs/articles/local/drupal-drush-command-line-utility.md
@@ -193,7 +193,7 @@ If you have successfully set up [Terminus](/docs/articles/local/cli/), the Panth
 ## Drush Alias Strict Control
 Create a file called `policy.drush.inc`, and place in in the `.drush` folder of your home directory.  You can create a new file or use the example policy file in Drush’s `examples` folder to get started.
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Maybe we should recommend using a policy file here, so the user does not have to re-edit on every download of their Pantheon aliases.

timani:
I am all for that business of adding a policy file too. Initial recommendation in the card was to edit but that is a better option

greg-1-anderson:
Someone wrote a blog post: https://pantheon.io/blog/fix-drush-site-aliases-policy-file  :smile:

timani:
Bloggers!

greg-1-anderson:
The policy file is already discussed (complete with a link to the blog post).  Maybe the link should point to #drush-alias-strict-control?

Also, the existing examples show changing the `drush` script. Maybe it would be helpful to have another example here that says `if alias uri == wrong.example.com then alias uri = example.com`.

greg-1-anderson:
@rachelwhitton: Should I literally use `wrong.example.com` and `example.com`?  Can you give some real-world examples of wrong and right domain names?

greg-1-anderson:
See https://github.com/pantheon-systems/documentation/pull/1188 for proposed policy file documentation change.

rachelwhitton:
+1 thanks @timani and @greg-1-anderson - @nataliejeremy can you review and merge?

PR #1164, opened 2015-12-10
Removed reference to W3 Total Cache plugin

newtoid:
This is listed as an unsupported module here https://pantheon.io/docs/articles/sites/code/unsupported-modules-plugins/#caching-plugins-(e.g.-batcache,-w3-total-cache,-or-wp-super-cache), and here https://pantheon.io/docs/articles/wordpress/launch-check-wordpress-performance-and-configuration-analysis/#unsupported-plugins

Referring to it as a beast of a plugin might encourage people to try and use it.


ari-gold:
:+1:

PR #1173, opened 2015-12-15
add quicksilver.md

bmackinney:
@joshkoenig 

First pass
needs:
 - [ ] terminus interaction


joshkoenig:
This is a good start. I'll add a little more of an introduction.

We should also add a section outlining the terminus commands, and remind people to get an up to date version of terminus.

PR #1181, opened 2015-12-16
Add note about using Drush SQL Sync Pipe

colans:
Users should be aware that, although "drush sql-sync" doesn't work on Pantheon, the alternative [Drush SQL Sync Pipe](https://www.drupal.org/project/drush_sql_sync_pipe) does work on Pantheon.  It should be the recommended approach rather than the inefficient method of dumping and importing.

### DIFF
--- a/source/docs/articles/local/drupal-drush-command-line-utility.md
+++ b/source/docs/articles/local/drupal-drush-command-line-utility.md
@@ -146,7 +146,7 @@ drush sql-dump --gzip --result-file=$HOME/Desktop/db_$(date +"%Y-%m-%d").sql
 ```
 ## Transfer Database Content Using Drush on Pantheon
 
-You cannot use `drush sql-sync` on Pantheon. Instead, download a copy of the database from a backup and save it to your local disk, then use `gunzip` and `mysql` to decompress and import the dump.
+You cannot use `drush sql-sync` on Pantheon, but you can use [drush sql-sync-pipe](https://www.drupal.org/project/drush_sql_sync_pipe). If that option doesn't appeal to you, download a copy of the database from a backup and save it to your local disk, then use `gunzip` and `mysql` to decompress and import the dump.
 ```sql
 gunzip < database.sql.gz | mysql -uUSER -pPASSWORD DATABASENAME
 ```


bmackinney:
@greg-1-anderson can you +1?

greg-1-anderson:
IIRC, SQL Sync Pipe does not work with Drush 8, which is now the default on Pantheon.  Their project page only mentions Drush 4-7, so not sure here.

Also, I think that sql-sync should work on Pantheon. I'll run a couple of tests, and see if we can't just update this to explain how to use sql-sync.

colans:
@greg-1-anderson Over lunch at DrupalCon, I mentioned that sql-sync didn't work on Pantheon, and you didn't believe me, but it's right there in the docs. ;)

There was a D8 branch for SQL Sync Pipe, but I just cut a [beta1 release](https://www.drupal.org/node/2637074) and updated the [project page](https://www.drupal.org/project/drush_sql_sync_pipe).  So yes, it definitely works with Drush 8, except for the sanitization option which is actually [a Drush bug](https://github.com/drush-ops/drush/issues/1869) (also affecting Drush core's `sql-sync --sanitize`), not a Drush SQL Sync bug (even though we have [one open for it](https://www.drupal.org/node/2537146)).  If you can offer any info/help there, it would be appreciated.

And if you can get sql-sync working instead, even better!

greg-1-anderson:
It doesn't have to be "instead"; we can document the correct use of all working options, sql-sync and sql-sync-pipe alike.

bmackinney:
@greg-1-anderson I think we can merge this PR and then create an issue to "document the correct use of all working options, sql-sync and sql-sync-pipe alike." 
+1 to that?

greg-1-anderson:
I have no objections, but I have not tested sql-sync-pipe.

markcarver:
FWIW and to give a little back history, the primary reason I originally created drush-sql-sync-pipe was because of Pantheon. At the time (not sure if it's still the case with current infra/terminus), SSH-ing into a Pantheon site was restricted. This in turn prevented Drush core `sql-sync` to not work.

So, since Pantheon allowed direct DB connections, I decided to come up with this solution to directly sync a DB via pipes. At one point even, Pantheon's docs had said to use this command over Drush core's (I briefly had conversations with @joemiller, perhaps @joshkoenig too, about it). It's also why Pantheon adds the `db-allows-remote` property in their site alias files (or used to). This was to inform the this command that it could connect directly to the database server instead of having to SSH in first.

Since this command was originally created, I actually haven't used it for Pantheon sites much anymore. So I really haven't kept up to date with what y'all are or have been doing with infra/terminus.

Anyway, hope this helps explain a bit of history since I cannot seem to find the original issues around it.

colans:
Looks like this got reversed in #1203 .

greg-1-anderson:
So, to return the the DrupalCon discussion, yes, you were right, at that time sql-sync did not work with Pantheon with default options in either direction, and could be made to work with certain options in one direction (Pantheon to local, which is what I was doing), but not the other direction.  It now works in both directions.

As mentioned in #1203, problems were encountered with sql-sync-pipe on the platform, so we removed the documentation about it and marked it unsupported. This could change in the future if the problems are fixed.

In the meantime, poor-man's sql-sync-pipe:

`drush @pantheon.source.dev sql-dump | $(drush @pantheon.destination.dev sql-connect)`

PR #1188, opened 2015-12-16
Related to #1163: Set site URI via policy file.

greg-1-anderson:
If the user has a site with multiple domains, this article modification will help the user select which domain should be used with their Drush alias for the specific site and environment.


rachelwhitton:
Thanks @greg-1-anderson! Pulled this work into the existing PR, we'll finish it out there.

PR #1195, opened 2015-12-17
remove CTA for weekly training

bmackinney:


bmackinney:
@rachelwhitton :rocket: with previous

PR #1199, opened 2015-12-18
Use http://portquiz.net:2222/ to test for blocked port

ari-gold:
Suggest using http://portquiz.net:2222/ to test for blocked ports


rachelwhitton:
+1 Thanks @ari-gold!

PR #1207, opened 2015-12-21
Update update-core-for-sites-running-outdated-versions-custom-distrib…

bgonz:
…utions.md

Removed additional "git pull" from the D6 code.


nataliejeremy:
Thanks for the update! @bgonz

PR #1210, opened 2015-12-22
Standardize on "replica" in hyperdb.md

stevector:
This doc and the code examples primarily use "replica." I think we should be consistent in that usage as many open source projects have made specific efforts to use that terminology over "slave."


nataliejeremy:
Thanks for the change! @stevector

PR #1211, opened 2015-12-22
Update accessing-mysql-databases.md

levonlee:
I don't get ERROR 2013 but instead I got the error in the edit.


ari-gold:
Thanks for the contribution @levonlee!

:+1:

PR #1230, opened 2016-01-06
Workflow Export doesn't contain code

slashzero:
Make this explicit as users are assuming it does.


rachelwhitton:
+1 Thanks for the PR @slashzero! Much appreciated

PR #1234, opened 2016-01-07
Suggest clearing caches before creating multidev

slashzero:
This has come up on numerous tickets recently, latest was https://pantheon-systems.desk.com/agent/case/52315


bmackinney:
@slashzero Thanks. We can merge/deploy. I think this should also be an engineering issue, as it will be more effective to add it as an option on the env-create workflow, with `--cc` in terminus or via a checkbox on the dashboard.

slashzero:
@bmackinney Yep - agreed

rachelwhitton:
+1 Thanks @slashzero!

PR #1246, opened 2016-01-11
Bots and Indexing update

### LINKED ISSUE #977: Bots and Indexing on Pantheon Doc Update
Re: [Bots and Indexing on Pantheon](https://pantheon.io/docs/articles/sites/code/bots-and-indexing)
## Issue Description:
Needs more detail
## Suggested Resolution 
Add custom base domains to .pantheon.io serving our default robots.txt. Also state more clearly that adding a custom domain to a dev- or test- environment is not a workaround  (currently says only custom domains on live, should also say not available on dev/test).

nataliejeremy:
Closes #977 
@bmackinney Please review


bmackinney:
@nataliejeremy reviewed.

PR #1251, opened 2016-01-11
Restrictions on switching upstreams

ccorpus:
Highlight restrictions regarding switching upstreams. Referring case: https://pantheon-systems.desk.com/agent/case/52578. Thanks!


nataliejeremy:
Thanks @ccorpus

PR #1258, opened 2016-01-14
Update Drupal 6 sites use case

rachelwhitton:
No Redis module is available for Drupal 6 - need to use cache backport. Current doc suggest you need to use Redis AND cache backport for D6 sites.


rachelwhitton:
@slashzero Do you have anything to add on this one? Saw the branch with no PR, but seemed complete.

+1

PR #1263, opened 2016-01-18
Add detection for drush when redirecting

### LINKED ISSUE #1289: Redirect Incoming Requests Doc Update
Re: [Redirect Incoming Requests](https://pantheon.io/docs/articles/sites/code/redirect-incoming-requests)
## Issue Description:
The provided code snippets doesn't check if the request comes from the cli. In our case we've got the problem that the cron wasn't executed because we redirect every non-https visitor to a https version-

## Suggested Resolution 
Check if the request comes from CLI by wrapping the following condition around the code snippets:
```
if (!(!isset($_SERVER['SERVER_SOFTWARE']) && (php_sapi_name() == 'cli' ||
   (is_numeric($_SERVER['argc']) && $_SERVER['argc'] > 0)))) {
   // ...
}
```

The condition was taken from [drupal_is_cli()](https://api.drupal.org/api/drupal/includes!bootstrap.inc/function/drupal_is_cli/7)

slashzero:
closes #1289 Suggested by https://pantheon-systems.desk.com/agent/case/45595

### DIFF
--- a/source/docs/articles/sites/code/bots-and-indexing.md
+++ b/source/docs/articles/sites/code/bots-and-indexing.md
@@ -47,14 +47,14 @@ The pantheon.io domains are ONLY intended for development use and cannot be used
 
  You can index your site under your production domain. There are many contrib module options available for creating sitemaps for Drupal, including [XMLSiteMap](https://drupal.org/project/xmlsitemap) and [Site\_Map](https://drupal.org/project/site_map). WordPress users can install the [Google XML Sitemaps plugin](http://wpcrux.com/collectives/wordpress-xml-sitemap-plugins/), which will maintain sitemap updates automatically once the initial build has been completed. It is up to you to configure the extensions to work as you desire. Pantheon does not offer support for Drupal modules or WordPress plugins.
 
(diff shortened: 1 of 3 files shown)


nataliejeremy:
Thanks @slashzero 
Can you +1 @newtoid

nataliejeremy:
@greg-1-anderson Can you take a look and +1?

greg-1-anderson:
Usually this is done via `if (php_sapi_name() != "cli") {`.  I think this is better to use the general-purpose test consistently in all examples, rather than depending on a Drupal-specific function for Drupal-specific examples.

Speaking of which, there are a lot of other examples in this same file of redirecting via `header` + `exit`.  All of these should be protected with a `php_sapi_name()` check.

bmackinney:
@greg-1-anderson does the snippet at https://github.com/pantheon-systems/documentation/issues/1289 satisfy?

greg-1-anderson:
I don't know that the `argc` check is necessary; seems redundant with the `php_sapi_name()` check. Maybe this is from older versions of PHP? Anyone know of a use-case where `php_sapi_name()` fails, but the `argc` check detects that the cli is being used?

greg-1-anderson:
I added another commit that just calls `php_sapi_name()` directly. This is simplest, and correct. There were a bunch of other places in the docs that referenced different kinds of redirects that used `exit()`; I fixed all of these here as well.

PR #1264, opened 2016-01-20
Improve instructions for merging existing code.

benjifisher:
I propose three changes to the instructions for merging an existing code base with Pantheon's:

1. Switch the order of the `git remote add` and `git pull` steps so that the new remote can be used instead of the copied `ssh_url` in the pull.
2. Remove the `git add -A` line before doing `git commit`. The files we want are already staged after the `git pull` command, and who knows what files might be lying around.
3. (Drupal only) Move the existing `settings.php` to `settings.local.php`. All sorts of things go wrong if you skip this.

### DIFF
--- a/source/docs/articles/sites/migrate/manual-site-import.md
+++ b/source/docs/articles/sites/migrate/manual-site-import.md
@@ -48,10 +48,24 @@ As long as you've chosen the same codebase (Drupal 7, Commerce Kickstart, etc.) 
  ssh://codeserver.dev.{site-id}@codeserver.dev.{site-id}.drush.in:2222/~/repository.git
  ```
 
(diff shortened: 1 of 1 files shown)


benjifisher:
BTW, I set my local branch to track `pantheon/master`.  This further simplifies the `git pull` and `git push` commands, but perhaps not everyone will want to do that.

After `git pull --no-rebase --squash -Xtheirs ...` I found that I could not push my local branch to the remote (pantheon).  I think we can solve this with an additional `git pull pantheon master` before the `git push` step.

rachelwhitton:
Thanks @benjifisher! Awesome contribution - I tested it and included the additional `git pull pantheon master` step. :+1: :ship:

PR #1274, opened 2016-01-22
If you downgrade to sandbox

ccorpus:
What happens if you downgrade to sandbox.

@nataliejeremy @rachelwhitton Please review. Thanks!


nataliejeremy:
@ccorpus Thanks for adding that.

PR #1278, opened 2016-01-22
Rsync optimize

rachelwhitton:
- Adjust paths in rsync commands so that only output_prod/docs/* are deployed
- Revise Grunt task so that only output_dev is tested for a11y instead of output_*
- Correct nokogiri command to use system libraries


bmackinney:
:+1:

PR #1283, opened 2016-01-25
fixed typo in core multisite-install command

aeligature:


nataliejeremy:
Thanks for updating that! @aeligature

PR #1300, opened 2016-01-28
exchanged gist for WP DEBUG to expanded form

ttrowell:
@bmackinney Please review this PR to enhance the WP DEBUG snippet.


bmackinney:
@ttrowell or @nataliejeremy let's change all of the gists to gist-it embeds, especially our `wp-config.php`, which should embed the file from pantheon-systems/wordpress.l

PR #1306, opened 2016-01-29
Order of operations when setting up domains with SSL

greg-1-anderson:
Walking through the domains setup instructions, it was a little unclear whether DNS or Cloudflare should be set up first. If you guess Cloudflare, then you end up in a loop, as you need to set up your DNS before you can set up Cloudflare.

At least, that's my understanding at the moment. I'll come back here and add more edits if needed once I complete the process.

### DIFF
--- a/source/docs/articles/sites/domains.md
+++ b/source/docs/articles/sites/domains.md
@@ -14,10 +14,9 @@ Configuring your domain's DNS is required to route traffic to your Pantheon site
 
 ## Step 1: Determine the URL to Serve From
 
-We recommend using the [HTTPS protocol](https://en.wikipedia.org/wiki/HTTPS) and the www subdomain prefix for all sites. See [this article](http://www.yes-www.org/why-use-www/) for information on why www is recommended with modern platform providers. 
+We recommend using the [HTTPS protocol](https://en.wikipedia.org/wiki/HTTPS) and the www subdomain prefix for all sites. See [this article](http://www.yes-www.org/why-use-www/) for information on why www is recommended with modern platform providers.
 
-If your site will use HTTPS, either [enable HTTPS with your own certificate](/docs/articles/sites/domains/adding-a-ssl-certificate-for-secure-https-communication/) or use [Cloudflare's free SSL service](/docs/guides/ssl-with-cloudflare/) before continuing.
-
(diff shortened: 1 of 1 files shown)


rachelwhitton:
@greg-1-anderson thanks! Do you have any additional updates for this PR?

source/docs/articles/sites/domains.md
```diff
@@ -14,11 +14,11 @@ Configuring your domain's DNS is required to route traffic to your Pantheon site
 
 ## Step 1: Determine the URL to Serve From
 
-We recommend using the [HTTPS protocol](https://en.wikipedia.org/wiki/HTTPS) and the www subdomain prefix for all sites. See [this article](http://www.yes-www.org/why-use-www/) for information on why www is recommended with modern platform providers. If your site will use HTTPS, either [enable HTTPS](/docs/articles/sites/domains/adding-a-ssl-certificate-for-secure-https-communication/) or use [Cloudflare's free SSL](/docs/guides/ssl-with-cloudflare/) before continuing.
+We recommend using the [HTTPS protocol](https://en.wikipedia.org/wiki/HTTPS) and the www subdomain prefix for all sites. See [this article](http://www.yes-www.org/why-use-www/) for information on why www is recommended with modern platform providers. If your site will use HTTPS, either [enable HTTPS with your own certificate](/docs/articles/sites/domains/adding-a-ssl-certificate-for-secure-https-communication/) or use [Cloudflare's free SSL](/docs/guides/ssl-with-cloudflare/) before continuing.
```
rachelwhitton: Suggest removing the last statement here, since it's repeated below in the warning statement. Otherwise, update so they are aligned. @nataliejeremy thoughts?

nataliejeremy: It looks like the note should be removed imo. I think that information should be presented in Step 1 and not Step 2.

rachelwhitton: +1 good plan @nataliejeremy

greg-1-anderson: The reason I repeated the phrase "with your own certificate" is that, strictly speaking, you are "using the HTTPS protocol" when you are using Cloudflare's service too.  There needs to be some way to name these two services to avoid confusion without sounding redundant.

greg-1-anderson:
I was successful here, and I don't think I have anything further to add, providing that the namesmithing issue is resolved satisfactorily.  I don't have any strong opinions on how this should go, except that whatever naming conventions we choose should avoid ambiguity.

rachelwhitton:
Thanks @greg-1-anderson! 

@nataliejeremy - I revised your last edit so it's clear there's a different order of operations for those using CloudFlare's free Universal SSL service. Can you review and merge?

:+1: :shipit:

PR #1307, opened 2016-01-30
Unpublish Okta doc until we update to new config

ari-gold:


nataliejeremy:
I created issue #1308 for the config update.

PR #1312, opened 2016-02-02
Capitalization of Terminus

millemr:
Capitalized Terminus in second bullet of benefits section.


nataliejeremy:
Thanks! @millemr

PR #1313, opened 2016-02-02
Add warning about deleting default.settings.php

slashzero:
Git will consider the directory to be empty and the symlink is lost leading to a broken site

### DIFF
--- a/source/docs/articles/drupal/configuring-settings-php.md
+++ b/source/docs/articles/drupal/configuring-settings-php.md
@@ -13,7 +13,7 @@ Drupal 8 sites on Pantheon run an unmodified version of core, bundled with a cus
 For Drupal 6/7, Pantheon uses a variant of Pressflow Drupal to allow the server to automatically specify configuration settings, such as the database configuration without editing settings.php. Permissions are handled automatically by Pantheon, so you can customize settings.php like any other site code.
 
 <div class="alert alert-danger" role="alert"><h4>Warning</h4>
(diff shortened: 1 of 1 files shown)


nataliejeremy:
Thanks for the update @slashzero

source/docs/articles/drupal/configuring-settings-php.md
```diff
@@ -13,6 +13,9 @@ Drupal 8 sites on Pantheon run an unmodified version of core, bundled with a cus
 For Drupal 6/7, Pantheon uses a variant of Pressflow Drupal to allow the server to automatically specify configuration settings, such as the database configuration without editing settings.php. Permissions are handled automatically by Pantheon, so you can customize settings.php like any other site code.
 
 <div class="alert alert-danger" role="alert"><h4>Warning</h4>
+You should not delete the <code>default.settings.php</code> file as this will render the symlink to your <code>sites/default/files</code> directory invalid.</div>
+
+<div class="alert alert-danger" role="alert"><h4>Warning</h4>
```
rachelwhitton: Thinking we can find a better place for the note? Also, should we include that this is for D7 sites and revise the statement to say the directory cannot be empty? @nataliejeremy @slashzero thoughts?

nataliejeremy: go for it @rachelwhitton

rachelwhitton:
@nataliejeremy I moved the warning to the FAQ section to avoid overuse of the alert styling. Can you review and merge?

PR #1321, opened 2016-02-04
Notify users that twenty- themes and akismet will not be updated

bmackinney:
@greg-1-anderson please review.


bmackinney:
@nataliejeremy does my change belong in a `div alert`?

nataliejeremy:
@bmackinney I think body text is fine; no need for an alert.

bmackinney:
@greg-1-anderson :+1: ?

greg-1-anderson:
+1

PR #1323, opened 2016-02-05
Integrating with Deploybot -- new doc

### LINKED ISSUE #905: Doc/Guide Request: Pantheon + Deploybot
A lot of customers make use of DeployBot (aka deploy.io, aka beanstalk). I think it should be pretty easy to integrate with Pantheon, and if so we'd definitely benefit from having a doc page/guide on how to set this up.

nataliejeremy:
Closes #905 

Working with David N to add/update:
- [x] Use cases and more benefits to using deploybot
- [x] Complete the Add a Server instructions. I was blocked here.
- [x]  Add the steps and info on deploying with deploybot
- [x] screenshot (if needed)


nataliejeremy:
@newtoid I sent you an invite to finish this up today from 1-3. Think you can make it happen?

nataliejeremy:
@newtoid How's this coming along? When do you think it will be finished?

allenfear:
@newtoid update?

newtoid:
@allenfear @nataliejeremy Using deploybot is possible, however it requires absolute path to the site, which can change. Users can get this via env variables, but then populating the path would require a manual step each time the binding info changed. Happy to document that with the caveat, what do you think?

allenfear:
@newtoid that caveat sounds good. Having this doc with the caveat would at least help folks who are trying to figure out why deploybot suddenly stops working for them.

source/docs/articles/sites/deploybot.md
```diff
@@ -0,0 +1,34 @@
...
+title: Integrating Deploybot with Your Pantheon Site
+description: Learn how to set up and use Deploybot on Pantheon.
+category:
+  - developing
+keywords: deploybot, sites
+---
+
+[Deploybot](http://deploybot.com/) allows you to build and deploy code from each deployment environments/branches to one or many servers simultaneously.
+
+One of the features of Deploybot is the ability to run a build process, like Grunt, to compile CSS before deploying. See their [guide](http://deploybot.com/guides/building-assets-with-grunt-or-gulp-during-deployment) for detailed information.
```
rachelwhitton: Explain why you'd want a build process to compile assets instead of committing them to version control. Mention DeployBot's friendly UI for managing configuration and build steps, as opposed to managing them in `yml` files like other CI service providers. Move guide link to the bottom.

rachelwhitton:
@ataylorme I took another pass at this doc after our last conversation, can you review? cc @newtoid

ataylorme:
@rachelwhitton sorry for the delay. I think the new doc looks good. It would be awesome if DeplyBot worked with Git so users don't have to check for changing SFTP credentials but last time I checked our Git repo URLs we're too long for their system.

rachelwhitton:
@ataylorme can you take one more look? I was able to connect using our git repo url but I'm not sure the deployment configuration is correct here for the purposes of compiling CSS and JS assets. I'm still suggesting SFTP with the destination set to `~/files` but I think that might not be the best way to go. When I test it this way it seems to send code to the filesystem which isn't what we want.

ataylorme:
>  I'm still suggesting SFTP with the destination set to ~/files but I think that might not be the best way to go. When I test it this way it seems to send code to the filesystem which isn't what we want.
This should probably be `~/code` as DeployBot ships the entire built site. Files will be included when `wp-content/uploads` is synced.

> I was able to connect using our git repo url
Have you tried deploying via git? I think if it's been fixed on DeployBot's end and works now we should recommend git over SFTP. With an SSH key added to DeployBot the credentials shouldn't change with git, like they do with SFTP.

rachelwhitton:
I didn't get the issue with a symlink we saw in the nested doc example @ataylorme, but this method does track `.revision` unexpectedly. Doesn't seem to be problematic. Ready for one final review, then it can go to @alexfornuto for a style pass.

ataylorme:
@rachelwhitton the new version looks great! Good job on this 👍

PR #1326, opened 2016-02-08
Match syntax in wp-config.php for else statement

bmackinney:
First reported in: https://pantheon-systems.desk.com/agent/case/54769
for reference: http://php.net/manual/en/control-structures.alternative-syntax.php

The block starting on line 144 also may need to change. I didn't touch the Drupal syntax, because I haven't looked at the injected `settings.php` file

@pantheon-systems/documentation please review


suzannealdrich:
That doesn't actually fix the main bug with this syntax, which is that the lack of `elseif` in the conditional blocks results in the base domain always being `HTTP_HOST` for all non-live environments. Please see PR #1335 for a complete fix, including syntax.

PR #1335, opened 2016-02-09
Free SSL with CloudFlare Guide Fixes

suzannealdrich:
- Added note about using the value `edge.live.getpantheon.com` for the CNAME flattened root with legacy gotpantheon.com development domains.
- Fixed bug in logic by adding correct `else if` statements. The conditional for setting `$base_url` and `WP_HOME` resulted in using the fallback value of `HTTP_HOST` in the `else` statement for all non-live hosts. Tested this fix to verify correct base domain functionality on all environments of WordPress, Drupal 6, and Drupal 7.
- Adjusted code style for WordPress and Drupal configuration files to match more closely with existing usage.

### DIFF
--- a/app/config/sculpin_site.yml
+++ b/app/config/sculpin_site.yml
@@ -104,3 +104,12 @@ authors:
     github: https://github.com/eeeschwartz/
     linkedin: https://www.linkedin.com/in/eeeschwartz
     bio: Civic developer, human centered!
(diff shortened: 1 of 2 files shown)


zepner:
+1

nataliejeremy:
Thanks for the update! @suzannealdrich

PR #1343, opened 2016-02-09

aeligature:
just a very slight clarification of the steps to solution


nataliejeremy:
Thanks @aeligature

Issue #1322, opened 2016-02-05
Considerations for Google Analytics and PANTHEON_STRIPPED Doc Update


rachelwhitton:
Re: [Considerations for Google Analytics and PANTHEON_STRIPPED](https://pantheon.io/docs/articles/sites/varnish/pantheon_stripped-get-parameter-values)
## Issue Description:
- [x] Document possible edge cases where PANTHEON_STRIPPED is passed as a utm parameter 

## Suggested Resolution 
- [x] Include diagram to demonstrate Varnish handling a common Google Analytics scenario
- [x] Work with @davidstrauss to provide a code snippet that removes undesired utm parameters before URLs are integrated into a response by the site's framework.

rachelwhitton:
@davidstrauss, thoughts?
```
// Remove query strings and tracking parameters from URLs
$strip = array('/&?__.+?(&|$)$/', '/&?utm_.+?(&|$)$/');
$_SERVER['REQUEST_URI'] = preg_replace( $strip, '', $_SERVER['REQUEST_URI'] );
```

rachelwhitton:
@davidstrauss can you review the code snippet?

davidstrauss:
This looks good! We should also document where to drop it (which would be `settings.php` for Drupal and elsewhere for WordPress).

PR #1355, opened 2016-02-12
Updating PANTHEON_STRIPPED doc

joshkoenig:
1. Don't tell people to use AJAX to pass utm_params to PHP. That will blow up scalability.
2. Adding proper explanation and proof for how our system actually works.


rachelwhitton:
Thanks @joshkoenig! Can you take a look at related issue #1355 - is that something we should work into this update?

joshkoenig:
@rachelwhitton I think this is issue #1355... did you mean to link elsewhere?

rachelwhitton:
Yep, sorry about that. It's #1322

joshkoenig:
Thanks! Let's get David's input on it. The current docs are not correct, so my update is good. If we can get more in-depth or a diagram that's good, but we may not need to block on it.

rachelwhitton:
:ship:

PR #1359, opened 2016-02-17
Update local.md

oseldman:
Correct terminus db and files backup commands.


rachelwhitton:
Hey @oseldman, thanks for the PR! The existing content is up to date with the most recent release of Terminus (0.10.3), can you update and try the commands again? The `--element` argument accepts both `db` and `database`

PR #1368, opened 2016-02-18
Comparing Multidev Environments Locally

### LINKED ISSUE #629: Document git commands used by multidev branch comparisons
Please work at @Hornswoggles to document the git commands that will generate the ahead/behind commits for multidev branch comparison.  Some customers are having issues with this, and we might as well document this for both customers and CSE so we have a better understanding of what is happening.

If it turns out that there is more info needed than the customer can get easily, ping me and we can figure out next steps.  

Thanks!

rachelwhitton:
Closes #629 

@nstielau can you review? I'm not sure that this is the same method used on the platform to generate the ahead/behind comparison but, seems to be a solid suggestion for comparing locally using git.

@slashzero can you review and +1?


nstielau:
+1

We are currently ordering the commits by time, which doesn't always work right (i.e. if an old commit was cherry-picked to more-recently).  Not sure if that is worth documenting.

rachelwhitton:
Thanks @nstielau! Updated 

@nataliejeremy can you copy edit this and merge? :ship:

PR #1375, opened 2016-02-19
Fix "pantheom" typo

hugovk:


nataliejeremy:
Thanks for the update @hugovk

PR #1378, opened 2016-02-19
redis docs

newtoid:
removed recommendation for specific module version


nataliejeremy:

PR #1389, opened 2016-02-23
IMCE Update: Use Backend Caching

### LINKED ISSUE #864: Test Patches for IMCE module
Re: [Unsupported/Problematic Modules and Plugins](https://pantheon.io/docs/articles/sites/code/unsupported-modules-plugins/)

Surfaced in https://github.com/pantheon-systems/documentation/pull/852#discussion-diff-38116432

 - https://gist.github.com/tauno/4236123
 - https://www.drupal.org/node/974568

Test patches and update doc to recommend a better solution than "split up your directories to avoid timeouts".

rachelwhitton:
Closes #864 

## Effect
- Modify IMCE so that backend caching is used for operations on highly populated directories

@alexdicianu can you review?


alexdicianu:
hey @rachelwhitton! Yep that's the patch we were using! looks good.

PR #1395, opened 2016-02-24
Remove href and clarify ignored dirs on backups

rachelwhitton:
Directories ignored when creating backups refers to a `.gitignore` file on the platform level, not application. 
- Remove hrefs since they're not relevant in this context
- Simplify text to easily communicate which dirs are not included in backups


nataliejeremy:
+1

Issue #1411, opened 2016-02-29
Performance testing of the compiled site


stevector:
## Issue Description

The front-end performance of the docs site could be improved in a couple of ways to speed up page load time. Here is a report on a guide page from sitespeed.io: https://results.sitespeed.io/481f-2016/1/29/995094a0-fd1e-4161-be8e-ffefb3a1ab76/index2.html

## Suggested Resolution

This site should have a concrete performance budget that can be tested with each pull request and tracked over time. I have a start on the mechanism over in https://github.com/stevector/documentation/pull/1 I will move that PR over to the main repo rather than a fork.

There are a couple of key questions to be answered:

- What should be the budget numbers for each metric measured? Or perhaps do we only care about total page load time and don't worry about some single metrics remaining red.
- How do we phase in the testing mechanism? We could set a strict budget and not merge in the testing until the strict budget is completely met. We could see a really easy budget just to get the PR testing merged and then progressively make the budget stricter as the performance is improved. Personally, I'd like to do something in the middle: merge in the testing itself along with whatever are the easiest/biggest fixes. We could then have follow-up issues for more complex/time consuming improvements.
- Can we graph changes over time? Sitespeed can integrate with Grafana but I don't know if it is worth our time to spin-up additional infrastructure.
- What URLs should be tested? sitespeed.io can crawl a site from an arbitrary starting point or it can take an exact set of URLs. I lean toward specifying an exact set of representative URLs.

ataylorme:
> How do we phase in the testing mechanism? ... I'd like to do something in the middle: merge in the testing itself along with whatever are the easiest/biggest fixes. We could then have follow-up issues for more complex/time consuming improvements.

I agree with finding the low hanging fruit for the initial integration. 

I think we should define what these initial items and break them out into separate issues.

For example, concatenation would correct _Number of JS files per page_ and _Number of CSS files per page_. This can be it's own issue.

Once all of the blocking issues are resolved then we run a few test to establish a baseline and merge this in.

> What should be the budget numbers for each metric measured? Or perhaps do we only care about total page load time and don't worry about some single metrics remaining red.

Steve we discussed this briefly yesterday but an alternative approach would be to allow for a certain deviation, such as -5%, and any build that decrease the site speed score more than the allowed difference from the previous test are rejected.

> What URLs should be tested? sitespeed.io can crawl a site from an arbitrary starting point or it can take an exact set of URLs. I lean toward specifying an exact set of representative URLs.

[...]

PR #1412, opened 2016-02-29
Sitespeed.io testing via Circle CI

stevector:
This PR is a start on https://github.com/pantheon-systems/documentation/issues/1411. This PR is not ready for review yet.

## Effect
- Introduces sitespeed.io testing to Circle CI.

## Remaining Work

The amount of work needed for this individual PR will be determined by discussion in https://github.com/pantheon-systems/documentation/issues/1411

This PR will likely need to be rebased as most of the commits so far (and more to come) are simply troubleshooting sitespeed.io/Circle CI integration.

### DIFF
--- a/budget.json
+++ b/budget.json
@@ -1,0 +1,16 @@
+{
+  "rules": {
+    "default": 90
+  },
+  "timings": {
+    "serverResponseTime": 300,
+    "domContentLoadedTime": 700,
+    "speedIndex": 1000
+  },
+  "wpt": {
+    "SpeedIndex": 1000
+  },
+  "gpsi": {
+    "score": 90
+  }
+}
--- a/budget.json
+++ b/budget.json
@@ -1,6 +1,6 @@
 {
   "rules": {
-    "default": 90
+    "default": 60
   },
   "timings": {
     "serverResponseTime": 300,
--- a/circle.yml
+++ b/circle.yml
@@ -19,8 +19,7 @@ dependencies:
     - bin/sculpin generate
     - cd output_dev && ln -s ./ source
     # Deploy Multidev environment
-    # uncomment this line once the PR is made to pantheon-systems/documentation
-    #- bash scripts/deploy-multidev.sh
+    - bash scripts/deploy-multidev.sh
 test:
   pre:
     # Start Ghost Driver.
--- a/circle.yml
+++ b/circle.yml
@@ -36,7 +36,7 @@ test:
     - scripts/merge_conflicts.sh
     # Run htmlproofer.
     - bundle exec rake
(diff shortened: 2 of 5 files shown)


circle.yml
```diff
@@ -31,6 +37,12 @@ test:
     - scripts/merge_conflicts.sh
     # Run htmlproofer.
     - bundle exec rake
+    # @todo, Switch the test to http://"$normalize_branch"-static-docs.pantheon.io
```
stevector: @rachelwhitton Do you know if/how I can get this variable available in this yml file? Or will the running of this test have to move to a script where the multidev URL can be constructed?

stevector:
@rachelwhitton, @bmackinney, Will I need to spin up a alternate Pantheon site in order to get the deploy to Pantheon working? PRs from forks don't build: https://github.com/pantheon-systems/documentation/blob/master/scripts/deploy-multidev.sh#L4

I don't mind doing so, I just want to ensure I do it right.

circle.yml
```diff
@@ -3,8 +3,14 @@ machine:
   php:
     # https://circleci.com/docs/environment#php
     version: 5.5.11
+  pre:
+    # @todo look for a better way to install phantomjs.
+    - sudo curl --output /usr/local/bin/phantomjs https://s3.amazonaws.com/circle-downloads/phantomjs-2.1.1
```
joemiller: The common pattern for doing things like this on circle seems to be:

1. download the artifact into a `~/deps`, add that dir to the circle.yml cache_directories so that it is available on future builds. If it's already there (and matches your expected version), don't re-download it.
2. install from `~/deps` to somewhere in the $PATH. You could put it into /usr/local/bin or put it in `~/bin` and ensure ~/bin is in the PATH (you can set PATH in circle.yml)

We have several examples throughout our codebase of this. Here is one I recently used in a non-pantheon project:

https://github.com/joemiller/creds/blob/master/test/circle_deps.sh#L9-L18
https://github.com/joemiller/creds/blob/master/circle.yml#L3-L7

stevector: Thanks @joemiller, I was doing this install here because the phantomjs version pre-installed with CircleCI wasn't high enough for sitespeed.io so I was getting an error on `npm install -g sitespeed.io`. I have now moved the installation of sitespeed.io to package.json and it is also finding the correct version of phantomjs.

budget.json
```diff
@@ -0,0 +1,16 @@
+{
```
stevector: This file is currently just a slight modification of the example budget file from sitepeed.io. The values in this file should be evaluated.

stevector:
I now have write access to this repo. I'm going to close this cross-repo PR and replace it with a PR within this repo.

PR #1447, opened 2016-03-16
Update manual-site-import.md

twooten:
## Effect
- changing "uploading" to "downloading"

## Remaining Work
- I have written a modified version of this script which allows uploading and downloading. Can I have someone review and test this?

Script is located here - https://gist.github.com/twooten/864eb120659247c9c52b

@rachelwhitton please test

The description for this script says that it uploads files but the command is for downloading files.

### DIFF
--- a/source/_docs/manual-import.md
+++ b/source/_docs/manual-import.md
@@ -113,7 +113,7 @@ fi
 
 while [ 1 ]
 do
-sshpass -p "$PASSWORD" rsync --partial -rlvz --size-only --ipv4 --progress -e 'ssh -p 2222'  $ENV.$SITE@appserver.$ENV.$SITE.drush.in:files/* --temp-dir=../tmp/  ./files/
+sshpass -p "$PASSWORD" rsync --partial -rlvz --size-only --ipv4 --progress -e 'ssh -p 2222' ./files/ --temp-dir=../tmp/ $ENV.$SITE@appserver.$ENV.$SITE.drush.in:files/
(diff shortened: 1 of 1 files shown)


source/_docs/manual-import.md
```diff
@@ -123,7 +123,7 @@ sleep 180
 fi
 done
 ```
-This script connects to your Pantheon site's Dev environment and starts uploading your files. If an error occurs during transfer, rather than stopping, it waits 180 seconds and picks up where it left off.  
+This script connects to your Pantheon site's Dev environment and starts downloading your files. If an error occurs during transfer, rather than stopping, it waits 180 seconds and picks up where it left off.
```
rachelwhitton: @twooten downloading/uploading are both accurate, but in the context of imports my preference is uploading - or we could go with something like this: 
> This script connects to your Pantheon site's Dev environment and starts synchronizing your files.

Let me know what you think
cc @nataliejeremy

twooten: @rachelwhitton I agree that in the context of imports, we should go with uploading. The example script we are showing is actually for downloading. Perhaps we should just change the script example so that it uploads?

`rsync -rlvz --size-only --ipv4 --progress -e 'ssh -p 2222' ./files/* --temp-dir=../tmp/ $ENV.$SITE@appserver.$ENV.$SITE.drush.in:files/`

rachelwhitton: Ah, good catch! Updated, thanks @twooten :+1:

Issue #1449, opened 2016-03-16
Composer Manager


rachelwhitton:
Re: <Doc title and/or link>
## Issue Description
Document composer manager 
Consider adding to [Unsupported/Problematic Modules and Plugins](https://pantheon.io/docs/articles/sites/code/unsupported-modules-plugins/)
## Suggested Resolution
- [x] decide if we need to re-work the installation instructions for Amazon S3 CORS, see [related drupal.org issue](https://www.drupal.org/node/2688553)
- [x] link to a doc on the proper use of Composer Manager - if we decide to go forward with it. @dustinleblanc is stumping that out now.

cc @ttrowell @allenfear

dustinleblanc:
Hey @greg-1-anderson, in this case https://pantheon-systems.desk.com/agent/case/56255 you suggested to the customer to disable composer manager and to instead use composer directly. While I agree with that general practice (using composer directly), after reading more docs on composer manager, I wonder if we can deal with platform compatibility by simply advising clients to commit their deps in development and ensure the auto-generation of files is disabled in test/live per the docs here: https://www.drupal.org/node/2405805
excerpt for easy reading:
```
Production Environments

Dependencies should be managed in development environments and not in production.
Therefore it is recommended to disable the checkboxes that automatically build the composer.json file and run Composer commands when enabling or disabling modules on production environments.

Assuming that you can detect whether the site is in production mode via an environment variable, adding the following snippet to settings.php will disable the options where appropriate:

<?php
// Modify the logic according to your environment.
if (getenv('APP_ENV') == 'prod') {
  $conf['composer_manager_autobuild_file'] = 0;
  $conf['composer_manager_autobuild_packages'] = 0;
}
?>
```
We could test this out to make sure it works as expected on the platform. While using composer the way it was intended is my preference, I think this could help out for folks that use a module that depends on CM (as in the issue @ttrowell dealt with). What do you think?

greg-1-anderson:
Yes, it's kind of a problem for us when a module says that it depends on Composer Manager.  If we ensure that `composer update` is never run on production, and users make sure that Composer Manager always builds on the dev machine (and the user then commits the result), then likely the result would work.

dustinleblanc:
@rachelwhitton so I am still getting the error even with the snippet in place. I think we might need to flip the switch back to "completely unsupported, don't use it" Even with it set to not automatically regenerate the files, I am getting this error just cuz:
[...]

PR #1450, opened 2016-03-16

dustinleblanc:
related to #1449 

## Effect
- Give instructions to use composer manager on the platform

## Remaining Work
- [x] Confirm that using this method allows cron and all other regularly scheduled tasks to run
- [x] Confirm that enabling and disabling modules does not cause Composer Manager to fire any composer tasks (composer update, etc)

@rachelwhitton please test

### DIFF
--- a/source/_docs/content-delivery-network.md
+++ b/source/_docs/content-delivery-network.md
@@ -72,7 +72,7 @@ Here's logic that you can use in settings.php to hard-code the URL of the CDN pe
 
 <div class="alert alert-info">
 <h4>Note</h4>
-The latest version of the Amazon S3 CORS module has Composer Manager as a dependency.  Composer Manager is not recommended or supported on the platform at this time. Native Composer installations are supported.  For details, see related drupal.org issue <a href="https://www.drupal.org/node/2688553">Composer Manager dependency</a>.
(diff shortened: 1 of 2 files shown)


rachelwhitton:
@dustinleblanc tested the snippet on a D7 spin up, looks good! Thanks for the PR. Are you taking on the remaining work? 

cc @ttrowell

dustinleblanc:
@rachelwhitton Lets sync up early next week to make sure I understand what is still needed.

dustinleblanc:
@rachelwhitton I think the only follow ups on this were to ensure cron runs fine and that enabling/disabling modules works with no issues. It looks like you already confirmed both, If that is the case, I think this is good to merge.

dustinleblanc:
@rachelwhitton I'm testing the entire amazon s3 issue from tip to tail and one thing I've noticed is that the instructions we give are for SFTP, if a person is using git and is commiting their vendor deps locally and *then* pushing them up to the server, we should probably have the docs tell them to disable the auto-build in all Pantheon environments, and possible set their vendor dir completely outside an env switch like this:
```php
# Set appropriate paths for Composer Manager
$conf['composer_manager_file_dir'] = $_SERVER["HOME"] . '/code';
$conf['composer_manager_vendor_dir'] = $_SERVER["HOME"] . '/code/vendor';
if (isset($_ENV['PANTHEON_ENVIRONMENT'])) {
  # Disable autobuild on Pantheon.
  $conf['composer_manager_autobuild_file'] = 0;
  $conf['composer_manager_autobuild_packages'] = 0;
}
```

PR #1459, opened 2016-03-21
Manual Site Import doc update

### LINKED ISSUE #1440: Migrate to Pantheon: Manual Site Import Doc Update
Re: [Migrate to Pantheon: Manual Site Import](https://pantheon.io/docs/articles/sites/migrate/manual-site-import)
## Issue Description:
 By default, settings.local.php file is not ignored by Git. It needs to be added to .gitignore file.

`5. Drupal only: Move your settings file to settings.local.php so that it will be ignored by Git and included from Pantheon's settings.php. First, make sure that you can modify it, and restore the protections after the move:`

https://pantheon-systems.desk.com/agent/case/57124
## Suggested Resolution 

Update doc to avoid confusion.

nataliejeremy:
Closes #1440 

## Effect
- Updates step 5 to instruct the user to add the settings.local.php file to .gitignore.

@rachelwhitton please review

### DIFF
--- a/source/_docs/manual-import.md
+++ b/source/_docs/manual-import.md
@@ -26,7 +26,7 @@ From your Pantheon Dashboard:
 
 * Choose **Create a new site**.
 * Name your site.
-* Select **Start from scratch**, and choose your starting codebase.
+* Select **Start from Scratch**, and choose your starting codebase.
 
 Starting from scratch allows your site to connect to that upstream so you can later [apply upstream updates](/docs/upstream-updates/) from your Dashboard with one click.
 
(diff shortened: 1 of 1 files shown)


rachelwhitton:
+1 thanks @nataliejeremy and @rubelyn!

source/_docs/manual-import.md
```diff
@@ -56,7 +56,7 @@ As long as you've chosen the same codebase (Drupal 7, Commerce Kickstart, etc.)
  git remote add pantheon <ssh_url>
  ```
 
-5. **Drupal only**: Move your settings file to `settings.local.php` so that it will be ignored by Git and included from Pantheon's `settings.php`. First, make sure that you can modify it, and restore the protections after the move:
+5. **Drupal only**: Move your settings file to `settings.local.php` and add it to `.gitignore` so that it will be ignored by Git and included from Pantheon's `settings.php`. Make sure that you can modify it, and restore the protections after the move:
```
rachelwhitton: Suggest removing this step - or clarifying the context so that the user knows how/when to reintroduce their configurations on Pantheon.

rachelwhitton: If we keep the step, we should add a snippet for D7 sites demonstrating how to include `settings.local.php`. If the step is removed, [line 66](https://github.com/pantheon-systems/documentation/pull/1459/files#diff-c6680de49ab763dc970a1fb757a86135R66) should be preserved in [Configuring settings.php](https://pantheon.io/docs/settings-php)

nataliejeremy:
@newtoid Can you take a look at the code in step 5 and confirm that it is needed? 
chmod u+w sites/default/{.,settings.php}
mv sites/default/{settings.php,settings.local.php}
chmod u-w sites/default/{settings.local.php,.}

nataliejeremy:
Thanks for updating that @newtoid

PR #1466, opened 2016-03-22
Link to patch, explain errors may persist despite config

### LINKED ISSUE #1449: Composer Manager

rachelwhitton:
Closes #1449 
add patch to composer manager documentation
@dustinleblanc can you review?

### DIFF
--- a/source/_docs/unsupported-modules-plugins.md
+++ b/source/_docs/unsupported-modules-plugins.md
@@ -55,7 +55,7 @@ $baseUrl = '/ckfinder/userfiles/';
 ### Composer Manager
 **Issue**: Composer Manager expects write access to the site's codebase via SFTP, which is prevented in Test and Live environments on Pantheon by design.
 
-**Solution**: As suggested within the [module documentation](https://www.drupal.org/node/2405805), dependencies should be managed in Dev exclusively. Place the following configuration within `settings.php` to disable autobuild on Pantheon. This will also set appropriate file paths for composer and packages so that they are stored within the root directory of the site's codebase and version controlled:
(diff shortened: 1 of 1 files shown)


dustinleblanc:
@rachelwhitton yep I think that gives enough information to reproduce my success and get folks running composer manager happy as clams. The only outstanding question I have is about alternate instructions for doing this in git push mode vs sftp mode. Should I add some verbiage around that to this PR or should we do that seperately?

rachelwhitton:
@dustinleblanc via Git doesn't seem to align with the intended use of the module, or maybe I'm misunderstanding that part. I'm in favor of keeping instructions/support as simple as possible, but if you'd like to add some context for connection mode feel free to do it within this PR and we can evaluate

dustinleblanc:
@rachelwhitton if someone were to use composer manager with a site in git mode they would just be essentially using it on their local machine, and then ensuring that it doesn't run any file manipulation tasks on the server. Basically they'd just make sure to switch off auto-generation of files and deps in *all* environments instead of just test/live. Its the way I've used the module in the past. We could keep the instructions simple and say to disable auto-generation on Pantheon all together and instruct them to _always_ generate the files via drush if they chose to do so on dev (with a note that they can only do so in sftp mode). I'm not sure how best to handle it other than I fear a git mode user might read our docs, and then enable a module in dev, and have their site vommit all over them in sad fashion and then this guy shows up 

sad site broken guy

Because composer tried to update the lock file to have that module listed.

I'll do a quick rewrite on this PR to cover this in the safer way

rachelwhitton:
+1 LGTM! thanks @dustinleblanc 

@nataliejeremy can you review for copy edits?

PR #1467, opened 2016-03-22
New doc: Creating Symbolic Links

### LINKED ISSUE #1446: New Doc Proposal: Creating Symbolic Links from code directories to files
## Description
Sometimes plugins expect for the code directory to have write permissions. This works on Dev, but not in Test/Live. 

## Outline

1. change your dev environment code from SFTP to Git mode. If you do not have Git installed locally, you'll need to follow instructions here: https://pantheon.io/docs/articles/local/starting-with-git/

2. In your local terminal, cd to the site code repository, make a copy of the writeable directory, and delete it from the code repository. 

3. Create, commit, and push the symlink to Pantheon.
```
ln -s ./wp-content/uploads/new-directory ./wp-content/path/plugin-expects-to-write-to
git add .
```
At this point, run `git status` and you should see the symlink in green and all of the language files in red.

4. Commit your change and push to Pantheon

```
git commit -m "create symlink to writeable directory"
git push origin master
```

5. Your symlink will then exist on the Dev environment, and when the plugin creates the files, it should continue to work, even with the Dev environment in Git mode. In your previous configuration, the plugin would fail while in git mode. You should not see the newly created files in the dashboard, as "ready to commit."

6. Deploy your code changes to Test and test them there

7. Deploy your code changes to Live and create the files again there. (You can upload the files you preserved in step 2 to the Live environment's @`/wp-content/uploads/new-directory` directory via SFTP prior to deploying, if you do not want the Live environment to have to recreate them)

Here's a nice resource you can use for creating symbolic links on Mac/Linux. http://apple.stackexchange.com/questions/115646/how-can-i-create-a-symbolic-link-in-terminal

nataliejeremy:
Closes #1446 

@bmackinney @aeligature Please review. We could use a little more content in the intro to explain why you'd need to do this. 
@rachelwhitton please test/review


bmackinney:
Also explain that this is for 'nix systems. Windows instructions don't exist. Run a VM.

rachelwhitton:
+1 @nataliejeremy can you do another review? then it's ready to :ship:

PR #1471, opened 2016-03-23
Remove comment that links to now deleted Okta doc

ari-gold:
Updated general SAML doc should give Okta users what they need to get SAML set up


ari-gold:
@rachelwhitton pls reviewmerge

rachelwhitton:
+1 thanks @ari-gold!

PR #1484, opened 2016-03-30
Update Sendgrid integration

### LINKED ISSUE #395: Create articles on SMTP so that email is not sent from mail()
Story:
As a user I want more stuff about SMTP so that email is not sent from mail()

“Class to use for the format() method” as MimeMailSystem and “Class to use for the mail() method” as SmtpMailSystem

There is a problem with the configuration of some Drupal modules that is resulting in PHP throwing an error that results in a crash of the thread and returns a 502 error:

Helpful links:
http://www.phparch.com/2014/07/sending-html-emails-with-drupal-7-webform-and-mandrill/
https://www.drupal.org/node/1200142
http://drupal.stackexchange.com/questions/77608/drupal-7-webform-module-is-not-sending-email-properly
http://drupal.stackexchange.com/questions/39474/how-can-i-use-the-mimemail-with-stmp-authentication-support-modules-together-on
@timani

alexdicianu:
Closes #395 

## Effect
- Document SendGrid Integration module for Drupal as an alternative to the SMTP module, which can be problematic due to dynamic outgoing IPs on Pantheon. 

@rachelwhitton please test


alexdicianu:
Need to add a screenshot as well.

alexdicianu:
@rachelwhitton +1

rachelwhitton:
+1 Thanks @alexdicianu! 

PR #1514, opened 2016-04-11
Update local-development.md

amklose:
Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed

@rachelwhitton please test

Looks like there was a tool removed from the list in the first bullet point under "Before you Begin" so I cleaned up the wording. I also added "--site=<site> --env=<env>" to the terminus clear-cache example because those parameters are reference in the alert below it.


rachelwhitton:
Thanks @amklose!

PR #1515, opened 2016-04-11
WP Redis: Install via Composer

rachelwhitton:
## Effect
- Remove installation instructions that advise users to copy `object-cache.php`, which can complicate update strategies
- Offer composer instructions as an alternative to the supported symlink

@ttrowell please test/review


rachelwhitton:
Tested and deployed up to live on my launched wp site, lgtm 👍  🚢

PR #1541, opened 2016-04-19
Drupal Performance and Varnish Caching Settings

### LINKED ISSUE #1410: Drupal 8 cache tags + Varnish
Re: https://pantheon.io/docs/articles/drupal/drupal-performance-and-caching-settings

## Drupal 8 Cache tags

Drupal 8 can send cache tags to reverse proxies like Varnish through HTTP headers. See "Integration with reverse proxies" in https://www.drupal.org/developing/api/8/cache/tags These cache tags allow the reverse proxies to know all of the cache tags that make up a response (node 123, taxonomy term 456) so that if node 123 is resaved, all cached responses using node 123 can be cleared from the reverse proxy.

Our page on the general subject of Drupal and Varnish does not make it clear if we support this behavior or how to validate that it is working.

## Suggested Resolution

Confirm the behavior works and add details to the docs page.

nataliejeremy:
Closes #1410

## Effect
- adds that d8 clear caching is not supported

@stevector Please review--thanks!

### DIFF
--- a/source/_docs/drupal-cache.md
+++ b/source/_docs/drupal-cache.md
@@ -8,7 +8,8 @@ keywords: drupal, performance, cache, caching, varnish, varnish caching
 To maximize your site's performance on Pantheon and to take advantage of our Varnish caching, you'll need to configure your site's performance settings. For more information, see [Varnish Caching for High Performance](/docs/varnish).
 
 ## Drupal 8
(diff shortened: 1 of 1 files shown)


stevector:
I'm going push a rephrasing and link to https://www.drupal.org/developing/api/8/cache/tags

nataliejeremy:
Thanks @stevector

PR #1549, opened 2016-04-21
Update generate-ssh-key.md

ari-gold:
@rachelwhitton @nataliejeremy

Git is a primary use case for SSH Keys


nataliejeremy:
Thanks @ari-gold

PR #1565, opened 2016-04-26
Remove --site and --env specific notes

nataliejeremy:
## Effect
- removes notes since --site and --env commands were deleted

@bmackinney Please +1


bmackinney:
👍 when Circle's done.

PR #1567, opened 2016-04-26
More Info on Site Freezing

lauren-kelly:
## Effect
- specifies which sites are frozen
- provides info on what to do for b0rked defrosts

## Remaining Work
- [ ] If needed

@rachelwhitton please test

I'm iffy on this part: "Should there be any issues with the un-freezing process, a backup of the site is available to be restored via the site dashboard." 
But we (cse) are typically seeing a few cases of it per day. We can handle it on our end, but it would be good if there were something here too.

### DIFF
--- a/source/_docs/platform-considerations.md
+++ b/source/_docs/platform-considerations.md
@@ -112,6 +112,6 @@ See [Unsupported Modules and Plugins](/docs/unsupported-modules-plugins) for an 
 
 ## Inactive Site Freezing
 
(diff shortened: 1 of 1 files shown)


nataliejeremy:
Thanks for updating this @lauren-kelly What do you mean when you say you're iffy on the backup part?

lauren-kelly:
I wasn't sure if you wanted to mention the 'if it fails' part, that's all :)

If the wording looks good to you, I'd love to see all of that info there for users.

On Tue, Apr 26, 2016 at 6:23 PM, Natalie Jeremy < notifications@github.com [notifications@github.com] > wrote:
Thanks for updating this @lauren-kelly [https://github.com/lauren-kelly] What do you mean when you say you're iffy on the backup part?

PR #1570, opened 2016-04-27
drush make file and build path

levonlee:
Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed

@rachelwhitton please test

The manifest file path has to start with code/sites otherwise it can't be found. Also, build path needs to be code as by default the build path is sites/all which doesn't work in Pantheon. A bit cumbersome but that's what I needed to get drush make work. The original command downloads the library to /srv/bindings/UUID/sites/all/libraries. It really should be /srv/bindings/UUID/code/sites/all/libraries


newtoid:
+1

nataliejeremy:
Thanks for the update @levonlee

PR #1575, opened 2016-04-28
Update read-environment-config.md

ttrowell:
- adds Note about Domain Access and the Status Tab

@rachelwhitton please test

Adds a Note about using Domain Access and the conflict with the Status Tab.
This is probably not an issue with the module, but instead happening because we extract the PRESSFLOW settings.


newtoid:
+1!!

nataliejeremy:
Thanks @ttrowell and @newtoid

PR #1576, opened 2016-04-29
Remove extraneous closing brace from snippet

icylace:
The code snippet this PR affects originally had an extra closing brace which would cause a syntax error.  The paragraph following the snippet makes it clear that the error in the snippet was not intentional.

@rachelwhitton please test


nataliejeremy:
Thanks @icylace

PR #1584, opened 2016-05-02
Add Instashow to unsupported plugins.

dustinleblanc:
## Effect
- Add Instashow to unsupported plugins.


nataliejeremy:
Thanks @dustinleblanc

PR #1586, opened 2016-05-03
Don't redirect platform domains

### LINKED ISSUE #1582: Remove line in example code defining a redirect on Pantheon CNAME
Currently our docs instruct users to set a redirect for the live-yoursite.pantheonsite.io domain. Redirecting this domain becomes a problem if the client needs to debug downtime that is occurring at the DNS level. If we can quickly check that live-yoursite.pantheonsite.io is up, we can be certain it's not a platform issue which means a faster ID'ing of the root cause.

Doc in question: https://pantheon.io/docs/redirects/

dustinleblanc:
Closes #1582 

## Effect
- Remove pantheonsite.io from examples to discourage redirection from our platform domains

## Remaining Work
none.

Our example shouldn't encourage redirecting platform domains as it can make certain issues more difficult to debug.


nataliejeremy:
Thanks for the update @dustinleblanc

PR #1589, opened 2016-05-05
Adding troubleshooting section if_dir() if_link()

albertcausing:
@rachelwhitton please test

This is when the created symlink was created properly as per our documentation but the module/plugin still say no directory found and when it try to create one the module say unable to create directory already exist.


newtoid:
I think this should be is_link or is_dir, please see http://php.net/manual/en/function.is-dir.php and http://php.net/manual/en/function.is-link.php

nataliejeremy:
@albertcausing Please test and update the PR. Thanks!

PR #1590, opened 2016-05-05
Added Apache Solr Multilingual module to unsupported list

twooten:
Included a fix for customers that have already enabled this module and broken access to the Solr server.
@rachelwhitton please test


nataliejeremy:

PR #1591, opened 2016-05-05
BigPipe update

peter-pantheon:
- Added Big Pipe to the Unavailable Features section

@greg-1-anderson maybe BigPipe docs should be expanded and/or included under caching in the docs too. thoughts or suggestions?
@rachelwhitton please test
@ttrowell for review


rachelwhitton:
@stevector thoughts? Curious because we closed a similar issue on cache tags #1410

stevector:
I think this PR is valuable. We should add a link to https://www.drupal.org/documentation/modules/big_pipe

wimleers:
Related: https://www.drupal.org/documentation/modules/big_pipe/environment

stevector:
Thanks @wimleers! I think our edge layer is the last piece: https://github.com/pantheon-systems/drops-8/issues/146

PR #1595, opened 2016-05-08
Enable Secure HTTPS Communication - Remove line break in openssl command

jgerminario:
Closes #

## Effect
- Removes line break in docs
-
-

## Remaining Work
N/A

@rachelwhitton please test

The key generation code sample is confusing/broken—can't be directly copy/pasted. Since the line wraps to two lines I didn't end up setting the -keyout flag twice while trying to enable SSL, which caused a 'Your key is encrypted with a passcode' issue later on.


nataliejeremy:
Thanks for the update @jgerminario

PR #1604, opened 2016-05-11
Rephrase on server development process

bensheldon:
This PR places the emphasis on returning to the Pantheon Site Dashboard rather than the Code tool. A fast reading of the previous language led to some confusion between Code and Connection Mode.


nataliejeremy:
+1 Thanks @bensheldon

PR #1606, opened 2016-05-12
Update domains.md

albertcausing:
Closes #

## Effect
Adding FAQ ###Why does my domain works without "www." in Safari and Firefox but not in Chrome?

@rachelwhitton please test

This is when FF and Safari trying to access a not properly setup bare domain or has not propagated yet but www.example.com is already accessible. FF and Safari will prepend `www.` so it will automatically work using www.example.com instead. While chrome will give the unresolved error or go to google search and use the typed word as the search keyword.


rachelwhitton:
+1 thanks for the contribution @albertcausing! much appreciated

PR #1610, opened 2016-05-18
Update to dig command example

peter-pantheon:
# Effect
Of the two dig command examples, the first should not have the @8.8.8.8 DNS server specified. And the first dig command example had dig twice.

## Remaining Work
- [x] Peter to work with research if these dig examples should be trying to use the codeserver vs appserver? In a similar dig example on this docs page the appserver is used: https://pantheon.io/docs/drush/#unable-to-connect-to-drush.in-hostnames-(dns)

@rachelwhitton please test

### DIFF
--- a/source/_docs/debug-connections.md
+++ b/source/_docs/debug-connections.md
@@ -19,25 +19,17 @@ Services such as MySQL and SFTP are running; however, the local client is unable
 terminus site wake
 ```
 ### Resolve DNS Issues
-Check to see if you get an IP address returned when you run the following command, replacing `<xxx>` with your site’s UUID:  
+Some ISPs have issues resolving a `drush.in` hostname. Troubleshoot connection errors using the following `dig` command,  replacing `<xxx>` with your [site’s UUID](/docs/sites/#site-uuid):  
 ```
 dig codeserver.dev.<xxx>.drush.in +short
 ```
-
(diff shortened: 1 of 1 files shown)


source/_docs/debug-connections.md
```diff
@@ -19,25 +19,17 @@ Services such as MySQL and SFTP are running; however, the local client is unable
 terminus site wake
 ```
 ### Resolve DNS Issues
-Check to see if you get an IP address returned when you run the following command, replacing `<xxx>` with your site’s UUID:  
+Some ISPs have issues resolving a `drush.in` hostname. Troubleshoot connection errors using the following `dig` command,  replacing `<xxx>` with your [site’s UUID](/docs/sites/#site-uuid):  
 ```
 dig codeserver.dev.<xxx>.drush.in +short
```
rachelwhitton: @peter-pantheon in regards to `codeserver` vs `appserver`, I defer to what CSE recommends. My understanding is that connection attempts from the network's default DNS server will produce similar results across `drush.in` hostnames. Do you have a preference or recommendation on which to use here?

rachelwhitton:
👍

source/_docs/debug-connections.md
peter-pantheon: Being as they are examples they can stay as is.

PR #1614, opened 2016-05-18
Update settings-php.md

tinefin:
Correct ev syntax (php code part in quotes)

@rachelwhitton please test


nataliejeremy:
Thanks! @tinefin

PR #1619, opened 2016-05-20
Update enable-https.md

lauren-kelly:
Closes #

## Effect
- updates to add clarity to the CSR and key creation
- provide an alternative option for those who cannot figure out openssl

@rachelwhitton please test


nataliejeremy:
Thanks! @lauren-kelly

PR #1633, opened 2016-05-31
Update BlogVault doc to reflect use of machine tokens

ataylorme:
There is a new version of the BlogVault migration plugin that uses machine tokens instead of SFTP credentials. We need our docs to be updated to reflect the change.

Please do **not** merge yet - we need to coordinate the timing with the deployment of the updated plugin to the WordPress repository.

@nataliejeremy any copy edit suggestions are welcome

cc @ari-gold


ari-gold:
Thanks @ataylorme! Added `hold for release` label

rachelwhitton:
+1 thanks for the PR @ataylorme! looks great

PR #1653, opened 2016-06-08
updating with reconnect information

lauren-kelly:
## Effect
trying to resolve issues such as https://pantheon-systems.desk.com/agent/case/62588 where users are lost and can't find their way to their site root

@rachelwhitton please test


nataliejeremy:
Thanks! @lauren-kelly

PR #1677, opened 2016-06-17
Fix small typo

twooten:
## Effect
PR includes the following change:
- overrride changed to override.


nataliejeremy:
Thanks @twooten

PR #1695, opened 2016-06-23
Update upstream-updates.md

newtoid:
Show users where to find upstream in case it is different than those listed here


nataliejeremy:

PR #1703, opened 2016-06-29
Use Global variable to explicitly disable cache

dustinleblanc:
## Effect
- adds additional means to exclude Drupal 7 page from cache.
-
-

## Remaining Work
- [ ] Test and confirm with wise folks that this is a good idea.

Based on https://www.drupal.org/node/23797#comment-3600288

@rachelwhitton please test :)


nataliejeremy:
@newtoid Can you review and +1 this?
Thanks for the update @dustinleblanc

newtoid:
+1. seems legit! https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/drupal_page_is_cacheable/7.x

nataliejeremy:
Thanks @newtoid and @dustinleblanc

PR #1716, opened 2016-07-04
Update process to Clone a Pantheon site

ari-gold:
- Previous documentation depends on being able to import 3 separate code, db, and files archives, which we no longer support.
- Simplify path from `clone-new-site-backups` to `clone-site` ( @rachelwhitton  needs redirect?)

/cc @newtoid


rachelwhitton:
+1 LGTM redirect required - I'll add it on merge. Thanks for the update @ari-gold! 

PR #1717, opened 2016-07-05
Update script for disabling APM for Google AMP

fusionx1:
PR includes the following change:

- Added `newrelic_ignore_transaction();` right after ` newrelic_disable_autorum (FALSE);` in the script without it is still being monitored by NR see https://pantheon.io/docs/new-relic/#amp-validation-errors


rachelwhitton:
+1 Thanks for the PR @fusionx1!

PR #1720, opened 2016-07-06
Choosing a Start State update

nataliejeremy:
## Effect
- adds footnote to indicate SFTP mode is required

I'll make a PR on the dashboard repo to update the UI text so customers will have this information before selecting the distribution. Thanks @icylace 
@rachelwhitton Please review/merge


bmackinney:
@nataliejeremy There was another platform change for the Drupal 8 framework that detected customers trying to install Drupal 8 in git mode, and replaced the error page with one that said to change into SFTP mode. Maybe that could be done here too.

rachelwhitton:
I agree @bmackinney - the platform should use similar error handling for all frameworks requiring SFTP mode during installation. Thinking we should close this PR and doc issue in favor of a PR similar to [this change](https://github.com/pantheon-cookbooks/endpoint/commit/3d63a128db369cee6118e2d57978abd0645856f7) 
@nataliejeremy thoughts?

nataliejeremy:
@rachelwhitton @bmackinney That's great. I think we should keep this in the doc as an FYI though.

rachelwhitton:
Replaced footnote with tooltip/popover 

nataliejeremy:
So glad we can do this! +1

PR #1740, opened 2016-07-13
Update manual migration step instructions

bensheldon:


rachelwhitton:
+1 Thanks for the PR @bensheldon! 

cc @icylace

PR #1747, opened 2016-07-19
fix logic for multidev

andrewmallis:
## Effect

Logic is backwards. Let's be positive. We should define TEST and LIVE, then catch the rest: DEV, and multidev. Not only is this clearer to read, but as it stood we would be implementing production cache settings on multidev environments. Perhaps it is desirable to set multidevs to mirror production, like we do TEST, but that really depends on workflow.

## Remaining Work
- [ ] determine if we want to enforce dev standards in multidev, or add a comment ot this effect.

### DIFF
--- a/source/_docs/settings-php.md
+++ b/source/_docs/settings-php.md
@@ -183,17 +183,8 @@ The Drupal 8 [configuration override system](https://www.drupal.org/node/1928898
 
     // All Pantheon Environments.
     if (defined('PANTHEON_ENVIRONMENT')) {
-      // Drupal caching in development environments.
-      if (!in_array(PANTHEON_ENVIRONMENT, array('test', 'live'))) {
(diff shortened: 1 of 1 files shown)


dustinleblanc:
@andrewmallis I think you're right, this is easier to read, the logic should match in the D7 section though, we should make sure a change to the approach in one is matched by the approach in the other.

rachelwhitton:
Thanks for the PR @andrewmallis! Agreed, this logic is easier to read. I updated the D7 snippet to align with your changes and added you to the site as a contributor. Feel free to edit your [contributor profile](https://github.com/pantheon-systems/documentation/blob/master/app/config/sculpin_site.yml#L169-L176)

As for multidev standards, we recommend setting configurations to mirror dev but it's not enforced. 

Thanks again!

PR #1748, opened 2016-07-19
Update redirects.md to remove redirect to lowercase snippet

twooten:
This snippets breaks Drupal 7's default image upload function.

Error message:

"An unrecoverable error occurred. The uploaded file likely exceeded the maximum file size (100 MB) that this server supports."

Reproduced with a clean vanilla D7 site with no extra modules or themes installed, only added the snippet to settings.php.

### DIFF
--- a/source/_docs/redirects.md
+++ b/source/_docs/redirects.md
@@ -227,3 +227,8 @@ When transitioning from a system that used a tilde to indicate a home directory,
       // Your custom logic.
     }
 
+## Redirect to Force Lowercase Letters
+WordPress automatically forces lowercase letters within URLs using the [`sanitize_title_with_dashes()`](https://core.trac.wordpress.org/browser/tags/4.6/src/wp-includes/formatting.php#L1744) function in core. Drupal sites can force lowercase letters using the following:
+
+1. Set general automatic alias settings  to **Change to lower case** within the [PathAuto](https://www.drupal.org/project/pathauto) module configuration (`/admin/build/path/pathauto`).
+2. Enable **Case Sensitive URL Checking** within the [Global Redirect](https://www.drupal.org/project/globalredirect) module configuration (`/admin/settings/globalredirect`).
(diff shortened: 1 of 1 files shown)


rachelwhitton:
Thanks for the PR @twooten! Do you know of an alternative method to force lower case URLs / URIs? cc @dustinleblanc

rachelwhitton:
@twooten  Should we recommend using the [Global Redirect](https://www.drupal.org/project/globalredirect) module for sites running Drupal 7 and leave the snippet in place for other frameworks?

twooten:
@rachelwhitton, for Drupal I believe most people do this:

1. Set general automatic alias settings (admin/build/path/pathauto) to 'Change to lower case' in PathAuto module.
2. Enable Case Sensitive URL Checking in the Global Redirect module (admin/settings/globalredirect)

So Pathauto is used to force generation of lowercase aliases when a node is created, and Global Redirect is used to redirect to that alias if the visitor types in an uppercase URL.

This is how it's done on drupal.org as well: https://www.drupal.org/node/958866

If an existing site has uppercase URLs, then one could just use normal 301 redirect code in their settings.php file to handle the redirection, or just use the Redirect module.

https://www.drupal.org/project/redirect

For WordPress, I tested this redirect code. It didn't throw any errors but I was unable to force an uppercase permalink anyway. It seems that WP has a function called sanitize_title_with_dashes() that lowercases the URL by default, so you have to work to undo that.

http://wordpress.stackexchange.com/questions/5029/how-can-i-make-capital-letter-upper-case-permalinks?answertab=votes#tab-top

The same solution works for WordPress: add the redirect code to /wp-config.php, or use one of the various redirection plugins.

https://wordpress.org/plugins/redirection/
https://wordpress.org/plugins/wp-force-lowercase-urls/

I don't think this is needed for WP, I created a page here for testing:

http://dev-tw-wp.pantheonsite.io/content/2016/07/29/testing-the-redirects/

Even if I change the URL to uppercase, it will automatically change it to lowercase and the page loads fine.

rachelwhitton:
+1 Thanks @twooten!

twooten:
+1 @rachelwhitton

PR #1751, opened 2016-07-19
Update sso-organizations.md

newtoid:
## Effect
 * Added note about connection name characters and max length taken from https://auth0.com/docs/api/management/v2#!/Connections/post_connections


rachelwhitton:
+1 Thanks @newtoid!

nataliejeremy:

PR #1761, opened 2016-07-25
Role-Based Permissions & Change Management

### LINKED ISSUE #1752: Role-Based Permissions
Update/improve doc based on feedback from an org (provided by Kim Kiser):

This stems from this organization dealing with an incident last week.  They have a set up where there is a live client site, and a testing client site (two separate sites).  A dev that thought they were on the testing site pushed code to the live env. However turns out they were on the actual client site, and pushed an in-progress feature that wasn't ready. They are still recovering from this.

There are a number of factors that contributed to this happening, one of which is clarifying what user roles have which permissions. I'm thinking about ways to clarify this in the UI directly, but wanted to send you feedback about making the documentation more understandable.  This org has been using pantheon for years, and they didn't understand what all the roles entailed :/  (and I have a hard time wrapping my head around this too!)

So some specific examples to try to clarify:
- “Team member” vs. "Developer" roles seem counter-intuitive to me. Team member actually means full permissions.  Developer can't push to live.  This seems reversed and is a larger problem in my mind :)  But can we think about calling this out better to help tell this?

- a goal this particular org has is to give the lowest level permission to some users ("lock permissions down" is the language they used) - and it's not clear which role would do that.

- the doc itself is a little confusing in that we are trying to describe all the roles at all the levels (enterprise, partner-level, site level). It might help to group these sections together more clearly?

- we might add some screenshots to this page to help them connect our role descriptions with what they are looking at in the actual UI.

- an update is needed for: "Use the Workflow tool on Test and Live" - the workflows tab was renamed to database/files a little while ago - is this referring to that tab? If so it needs the updated name.

nataliejeremy:
Closes #1752 

## Effect
- adds 2 FAQs
- updates "Workflow" tab to "Database / Files"

@kimby77 and @ari-gold Regarding the discussion in issue #1752, I think some of the confusion can be solved within the UI. But please comment on any changes you'd like to see in the doc to  clarify things more for users. This is confusing for me so any input would be greatly appreciated! 

### Todo
- [x] Replace tables with bootstrap responsive styling

### DIFF
--- a/source/_docs/change-management.md
+++ b/source/_docs/change-management.md
@@ -182,43 +182,63 @@ In some Dashboards, you may notice the "User in Charge" label applied to a user.
 
 <style type="text/css">
 .tg  {border-collapse:collapse;border-spacing:0;}
-.tg td{font-size:14px;padding:10px 5px;border-style:solid;border-width:1px;overflow:hidden;word-break:normal;}
-.tg th{font-size:14px;font-weight:normal;padding:10px 5px;border-style:solid;border-width:1px;overflow:hidden;word-break:normal;}
-.tg .tg-e3zv{font-weight:bold}
+.tg td{font-family:Arial, sans-serif;font-size:14px;padding:10px 5px;border-style:solid;border-width:1px;overflow:hidden;word-break:normal;}
+.tg th{font-family:Arial, sans-serif;font-size:14px;font-weight:normal;padding:10px 5px;border-style:solid;border-width:1px;overflow:hidden;word-break:normal;}
+.tg .tg-baqh{text-align:center;vertical-align:top}
+.tg .tg-9hbo{font-weight:bold;vertical-align:top}
+.tg .tg-amwm{font-weight:bold;text-align:center;vertical-align:top}
+.tg .tg-yw4l{vertical-align:top}
 </style>
 <table class="tg">
-    <tr>
-        <th><strong>Role</strong></th>
-        <th><strong>Permissions</strong></th>
-        <th><strong>Enterprise Organization</strong></th>
-        <th><strong>Partner Organization</strong></th>
-    </tr>
-    <tr>
-        <td>Administrator</td>
-        <td>&#8226; Create and delete sites</td>
-        <td><center>✓</center></td>
-        <td><center>✓</center></td>
-    </tr>
-    <tr>
-        <td>Team Member</td>
-        <td>&#8226; Deploy code to Test and Live<br/>&#8226; Access the Database / Files page<br/>&#8226; Change the PHP versions for all site environments<br/>&#8226; Add Team Members and Supporting Organizations</td>
-        <td><center>✓</center></td>
-        <td><center>✓</center></td>
-    </tr>
-    <tr>
-        <td>Developer</td>
-        <td>&#8226; Can only commit code to Dev and Multidev environments<br/>&#8226; Add and remove add-ons to sites.<br>
-        <strong>Note</strong>: You can promote a Developer to Team Member for a single site by adding the user to the site's team.</td>
-        <td><center>✓</center></td>
-        <td><center>✓</center></td>
-    </tr>
-    <tr>
-        <td>Owner</td>
-        <td>&#8226; Functionally is the same as an Organization Administrator<br>
-        &#8226; Only role that can delete a site after it is live</td>
-        <td></td>
-        <td><center>✓</center></td>
-    </tr>
+  <tr>
+    <th class="tg-9hbo">Permissions</th>
+    <th class="tg-amwm">Administrator<br>Enterprise and Partner Orgs</th>
+    <th class="tg-amwm">Team Member<br>Enterprise and Partner Orgs </th>
+    <th class="tg-amwm">Developer <br>Enterprise and Partner Orgs</th>
+    <th class="tg-amwm">Owner<br>Partner Org</th>
+  </tr>
+  <tr>
+    <td class="tg-yw4l">Create sites within an org</td>
+    <td class="tg-baqh">✓</td>
+    <td class="tg-baqh">✓</td>
+    <td class="tg-baqh">✓</td>
+    <td class="tg-baqh">✓</td>
+  </tr>
+  <tr>
+    <td class="tg-yw4l">Access the org Dashboard</td>
+    <td class="tg-baqh">✓</td>
+    <td class="tg-baqh">✓</td>
+    <td class="tg-baqh">✓</td>
+    <td class="tg-baqh">✓</td>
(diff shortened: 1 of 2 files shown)


source/_docs/change-management.md
```diff
@@ -180,3 +180,11 @@ In some Dashboards, you may notice the "User in Charge" label applied to a user.
         <td><center>X</center></td>
     </tr>
 </table>
+
+## Frequently Asked Questions (FAQs)
+
+#### Which role should I assign a user to give them the lowest level of access?
```
ari-gold: Yes, this is correct at site level. At org level, Unprivileged is lowest level of access.

ari-gold:
@nataliejeremy Good additions. What do you think about reworking the tables to be in the format of the screenshot I provide? I think that format is a lot easier to grok.

nataliejeremy:
@ari-gold Sounds good. Do you think the site-level table should stay the way it is? It displays site-level permissions for Partner and Enterprise orgs: https://pantheon.io/docs/change-management/#site-level-roles/permissions

ari-gold:
Nice! @nataliejeremy Yes, if you can take a stab at re-working the sites level perms table I think that would be valuable.

nataliejeremy:
Done! @ari-gold

source/_docs/change-management.md
```diff
@@ -64,119 +64,187 @@ In some Dashboards, you may notice the "User in Charge" label applied to a user.
...
-        &#8226; Only role that can delete a site after it is live</td>
-        <td></td>
-        <td><center>X</center></td>
-    </tr>
+  <tr>
+    <th class="tg-9hbo">Permissions</th>
+    <th class="tg-amwm">Administrator<br>Enterprise and Partner Orgs</th>
+    <th class="tg-amwm">Team Member<br>Enterprise and Partner Orgs </th>
+    <th class="tg-amwm">Developer <br>Enterprise and Partner Orgs</th>
+    <th class="tg-amwm">Owner<br>Partner Org</th>
```
ari-gold: How about making this the first or second column so it's most to least privileges left to right?

kimby77:
@nataliejeremy I have my head in something now but I will try to look at this soon! Thanks!

kimby77:
I think the tables and the FAQs are great additions! Though overall I still find this page convoluted.

My suggestions are:
- the tables and "how to manage" sections are the most helpful here.  The tables enable the user to tell at a glance what each role means and what it can do.  The "how to" parts of it tell the user exactly how to accomplish what they need to do.  So with that in mind, I think those should both come sooner in the article & have the most focus. Everything else should be secondary info.

- for managing people in an org, and managing an enterprise team: under each of those headings are instructions for how to do things like "add a user, change a role, change a site owner, add a supporting org".  I think we can experiment with those tasks being the headings.  So "Add a User" would be a heading, and instructions underneath can be all the ways to add a user for both an org and a site.  Organizing it this way might help people skimming this page more quickly pinpoint info they've come here to find.

- the "features" section is difficult to understand.  I'm not sure it's necessarily? if it is, can it come near the end and/or maybe broken up and integrated with the FAQ so that it can be in more digestible chunks?  ie. Q: Who can assign roles?  A: Partner organizations can assign administrator, team member, and developer roles for their organization

- The "Enterprise" and "Partners" tables: looks like these are the same with the exception of the Unprivileged role for enterprise.  Can you try to combine these two tables, and call out the "unprivileged" role as a feature only for enterprise level orgs?

nataliejeremy:
Thanks for the feedback @kimby77 
I made most of the requested changes--please review:
https://1752roles-static-docs.pantheonsite.io/docs/change-management/
Note that adding/managing users for a site is different than managing people in an Org, so I left the H2s there to clarify that for the users.

kimby77:
This is really shaping up!  Love that those tables were able to be combined.

- for the two tables - I'm not seeing a great difference in them and am wondering if there's a way of combining into one table? I get one is for orgs, and one is for sites, but the permissions in the left column are identical and there's a lot of overlap of roles, so it's redundant.  Is there a way to combine them and differentiate site vs. org in some way?  If not, an alternative could be having the org table under the "manage people in an org" section, and the site table under the "manage a site's team" section.

- on the table headers, I suggest de-emphasizing the text in parenthesis, to keep attention on the titles themselves. Something like this and same with the "partner orgs only" on the table beneath it:

image

kimby77:
cc @informanddelight  - your thoughts welcome here too.

informanddelight:
@kimby77 hesitant to break out the two charts. Unfortunately we have 2 different contexts for users to have to navigate, their Site level permissions and the Org level permissions. It might be nice to distinguish the two charts a bit more ... I'll do some thinking on what might work. 
For now I'd rather just make the two a little more visually distinct from each other.

Maybe instead of parentheticals we can use a *? Just to have a note off to the side.

kimby77:
Ok.  I think it'll help if we can indeed distinguish the tables then.  An asterisk off to the side might work.  I'm not tied to a particular way of doing it, just looking for it to be de-emphasized in some way so that it doesn't muddy up the header title.

Thanks for your work on this!

rachelwhitton:
@kimby77 @nataliejeremy @informanddelight thoughts?
## De-emphasize Parentheticals
b

## Tooltip
Or we could do a tooltip - thoughts?
a

nataliejeremy:
+1 for tooltips. Thanks @rachelwhitton

informanddelight:
@rachelwhitton @kimby77 I might advocate for a different glyphicon to attach a tooltip to. Like the 
image glyphicon. This way it looks like there's a "group" that Owner or Unpriveleged can be a part of and there's a more specific use case attached to this.

Thanks so much for taking a crack at this. Man... our support/roles structure is so overwhelming, working on a similar project and it's really really really convoluted. This helps though!

rachelwhitton:
+1 Updated! Thanks @informanddelight 

screen shot 2016-08-04 at 1 35 50 pm

kimby77:
I think the tooltip will work! It has the information available if they need it, but it doesn't clutter up the table like the original version.  But that person icon is getting all smushed together at the small size, plus I don't think it's obvious to click there for more information about that column.  I think a question mark would do there.

nataliejeremy:
Thanks @kimby77 and @informanddelight For informational tooltips, we're going to stick with the i indicator to let users know there is information behind it and to remain consistent throughout docs. Thanks for showing us that icon though, we'll definitely find a place for it somewhere else.

kimby77:
Thumbs up for consistency!

kimby77:
@rachelwhitton I also suggest some differentiation between a table header and a table cell.  Currently the text is styled the identically for each.  Some differentiation will help users skim easier.

PR #1763, opened 2016-07-25
Mention that uncommitted code will not be in deployments

icylace:
## Effect
- adds some detail about uncommitted code not being in deployments
- minor formatting clean-up


nataliejeremy:
Thanks for the update @icylace

rachelwhitton:
+1 Thanks @icylace

Issue #1646, opened 2016-06-06
New Doc: Migrating Sites from Drupal Gardens to Pantheon


rachelwhitton:
## Title
Migrating Sites from Drupal Gardens to Pantheon

## Description
Re: https://www.drupalgardens.com/
Drupal Gardens will be ending on August 1

## Outline
- Export Site from Drupal Gardens
- Stategize and Create Archives for Import
- Manually Migrate or Import to Pantheon

## Expected Audience
Drupal Garden users

## Path
`source/_docs/migrate-drupal-gardens.md`

cc @hengkit

Stella3:
This is me. I have a pretty extensive site with a lot of webforms, views, etc. It appears that, if I can migrate, I would stay with D7 and add Sweaver to end up with a similar experience to what I was used to with Gardens.

Stella3:
Rachel,

I'm coming to the realization that migrating is not necessarily the best option. The theme I had with Gardens is dated and non-responsive and I would like a new look for the site. I've found[ a theme that I really like](http://themeforest.net/item/aqua-spabeauty-responsive-commerce-drupal-theme/13018418).  

I will need a bit of help getting that theme loaded into Pantheon and then I get a bit lost. Once that is loaded and I put content in, what, exactly is the process to make minor changes to the theme? Gardens had the ThemeBuilder which was similar to Sweaver but much much better and I could inject CSS, modify elements, etc. In addition, any element (images, text, menus, blocks) could be clicked on and modified without going to the code. 

I guess what I'm asking is, I'd like to hire somebody to load the theme up for me and then add my own content manually but am I going to have to modify code directly beyond that to work with content (background images, etc.)? If so, it appears that I could go way back to using Dreamweaver or modify it in Word and paste it back. 

The other concern is that I don't know if that template uses Blocks on [a page like this](http://themeforest.net/item/aqua-spabeauty-responsive-commerce-drupal-theme/full_screen_preview/13018418) where it says "Our Team" so I could add a dozen more of our staff into that or if I have to actually code in another row. 

I really really REALLY appreciate any insight you have as, with a 40 day deadline, I've slppt about 8 hours total in the past 4 days. I've contacted a few Drupal services who say they do very small projects and ongoing maintenance but they are all either 30+ days out. 

Thank you again Rachel!

PR #1764, opened 2016-07-25
Add a workaround for fixing Drupal Gardens site exports prior to import

greg-1-anderson:
Currently no open issues on this; related to #1646.

## Effect
- Instructions on removing settings.php from Drupal Gardens archives prior to importing on Pantheon.

Drupal Gardens actually includes **two** copies of settings.php in each site archive.  This can be confirmed via `tar tzvf drupalgardenssite.tar.gz | settings.php`; you will see two listings with the same path and filename.

Removing just one copy would probably be sufficient (untar and then tar again), but I did not test this, as removing settings.php entirely is more straightforward and safer, not to mention easier to explain.


rachelwhitton:
+1 Thanks @greg-1-anderson !

@nataliejeremy can you review/merge?

nataliejeremy:
Thanks! @greg-1-anderson

PR #1775, opened 2016-08-01
Update getting-support.md

ttrowell:
## Effect
- clarifies support hours for standard cases (non-Elite)

## Remaining Work

@twooten @newtoid  Thoughts?

### DIFF
--- a/source/_docs/getting-support.md
+++ b/source/_docs/getting-support.md
@@ -34,7 +34,7 @@ If you cannot sign in to your Pantheon Dashboard:
 * **Standard support ticket**: Your general support questions are routed through our support queue ahead of other users. Expedited support is accessed through our Dashboard and email.
 
 ###Paid and Free Sites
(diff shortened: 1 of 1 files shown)


ttrowell:
Tim, I didn't add in PH time just yet.

newtoid:
If this is achievable then +1, and we should update this page https://pantheon.io/pricing-comparison

I would say the 2 hour turnaround on Free/sandbox sites might be ambitious during high volume times, so maybe best effort for sandbox/pro with priority on business, and published/agreed SLAs for Elite and PES, just to manage client expectations.

ttrowell:
@newtoid The page currently lists the 2 hour window for non-SLA sites.  I only added the actual current support hours.  I agree that 2 hours could be ambitious at times, though.

ttrowell:
@nataliejeremy Let's update this so we have some clarification, then we can modify it again if needed.  Sound ok?

I don't think the 2 hour window for is guaranteed at all  - excepting Elite sites.

nataliejeremy:
Thanks! @ttrowell

PR #1789, opened 2016-08-05
Update custom-upstream.md

newtoid:
added note about gitlab required permission levels as per http://docs.gitlab.com/ce/user/permissions.html

"Guest" account level cannot pull code and this prevents site creation


nataliejeremy:

PR #1795, opened 2016-08-09
Suggested Updates to Clarify Drupal Solr Install

peter-pantheon:
After a couple of recent solr support tickets, I was able to  recognize some potential confusions in the solr installation steps. I have added some clarifying text that I think can help. 

## Effect
PR includes the following changes related to these confusions: 
[1] One customer thought they had to install both apachesolr and search_api_solr modules. 
[2] Another customer was not clear that multidev needed posting of schema.xml.  
[3] Customer was confused as to whether there was a difference in whether one posted the apachesolr schema.xml or the search_api_solr schema.xml. And then did the settings.php entry need to match that pick.

## Remaining Work
- [x] Discuss and verify changes

@rachelwhitton i am open to discuss my updates

### DIFF
--- a/source/_docs/solr-drupal.md
+++ b/source/_docs/solr-drupal.md
@@ -62,7 +62,7 @@ if (isset($_ENV['PANTHEON_ENVIRONMENT'])) {
 
 <div class="alert alert-info" role="alert">
 <h4>Note</h4>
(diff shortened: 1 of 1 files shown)


rachelwhitton:
+1 LGTM thanks for contributing @peter-pantheon!

PR #1797, opened 2016-08-09

ttrowell:
## Effect
-  Add a note that the list of modules/plugins is not comprehensive
-

@nataliejeremy please review and re-word if preferred.   We have some customers saying that since we didn't list a module/plugin on this page, that it should be supported.


nataliejeremy:
Thanks! @ttrowell

PR #1801, opened 2016-08-11
Describe how to fix the Site URI for Drush 8

eeeschwartz:
I don't believe drush 8 uses the policy file (it wouldn't work for me anyway). But I have had luck editing .drush/drushrc.php as described in this case.

Let me know if there is a gotcha I'm overlooking.


eeeschwartz:
The best way to get around this error is to avoid redirecting drush. I'll close this and open a separate PR to note that.

PR #1814, opened 2016-08-18
Reliable locations for drushrc.php

### LINKED ISSUE #1779: Drupal Drush Command-Line Utility Doc Update
Re: [Drupal Drush Command-Line Utility](https://pantheon.io/docs/drush)
## Issue Description:

Include info on location for drushrc.php to specify drush operations on platform. 
## Suggested Resolution

Recommend drushrc.php be placed in sites/all/drush

Issue results from this Jira bug:
https://getpantheon.atlassian.net/browse/BUGS-388?focusedCommentId=33666

rachelwhitton:
Closes #1779 
## Effect

- Note most reliable locations for drushrc.php

cc @greg-1-anderson @allenfear

### DIFF
--- a/source/_docs/drush.md
+++ b/source/_docs/drush.md
@@ -413,18 +413,16 @@ See the [Drush Migrate documentation](https://drupal.org/node/1561820) for detai
  drush @pantheon.example.live  st
  Drush command terminated abnormally due to an unrecoverable error.       [error]
  ```
- Setting the `--uri` option will not work. To resolve this error, use a [Drush policy file](#drush-alias-strict-control) or drop the following in `drush/drushrc.php`:
+ To resolve this error, conditionally set `$uri` based on the environment in `drushrc.php`, such as:
 
  ```
-   $uri = 'http://usenix.dev';
-   if (isset($_SERVER['PANTHEON_ENVIRONMENT'])) {
-     $uri = 'https://' . $_SERVER['PANTHEON_ENVIRONMENT'] . '.usenix.dev';
-     if ($_SERVER['PANTHEON_ENVIRONMENT'] === 'live') {
-       $uri = 'https://www.usenix.org';
+   if (isset($_SERVER['PANTHEON_ENVIRONMENT']) &&
+     ($_SERVER['PANTHEON_ENVIRONMENT'] === 'live')) {
+       $uri = 'https://www.example.com';
      }
    }
+   $options['uri'] = $uri;
  ```
- `$options['uri'] = $uri;`
 
  The most reliable locations to put `drushrc.php` files are:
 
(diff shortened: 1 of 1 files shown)


source/_docs/drush.md
```diff
@@ -365,10 +365,30 @@ See the [Drush Migrate documentation](https://drupal.org/node/1561820) for detai
 ## Known Limitations
 - Crontab: Currently, there is no way to manage Crontab on Pantheon. If you need a way to set up your own Cron interval, you can use an external cron service such as [Easy Cron](https://www.easycron.com/user/register).
 - The following Drush commands are not supported and will not work on Pantheon sites: `sql-sync-pipe`, `sql-sqlc`, `php-eval`. An alternative to `drush sql-sync` is to use `drush sql-dump` instead.
-- Incorrect `['uri']` in `pantheon.aliases.drushrc.php` file. Drush may fail if the `['uri']` array key has a different domain than what is expected by Drupal, resulting in the following error:
+- Drush may fail if the `['uri']` array key has a different domain than what is expected by Drupal, resulting in the following error:
```
greg-1-anderson: The failure here isn't so much that the `uri` is wrong -- usually, the "Drush command terminated abnormally" happens when folks add redirects that do not disable themselves when called from a cli tool.  The appropriate solution is to wrap any sort of drupal_goto(), or any other function that might call `exit()` in:

`if(php_sapi_name() != 'cli') { ... }`

There are some modules that require a correct URI in order to work correctly, though, so the basic advice in this document is still useful.

source/_docs/drush.md
```diff
@@ -365,10 +365,30 @@ See the [Drush Migrate documentation](https://drupal.org/node/1561820) for detai
...
  drush @pantheon.example.live  st
  Drush command terminated abnormally due to an unrecoverable error.       [error]
  ```
- Setting the `--uri` option will not work. To resolve this error, use a [Drush policy file](#drush-alias-strict-control).
+ Setting the `--uri` option will not work. To resolve this error, use a [Drush policy file](#drush-alias-strict-control) or drop the following in `drush/drushrc.php`:
+
+ ```
+   $uri = 'http://usenix.dev';
+   if (isset($_SERVER['PANTHEON_ENVIRONMENT'])) {
+     $uri = 'https://' . $_SERVER['PANTHEON_ENVIRONMENT'] . '.usenix.dev';
```
greg-1-anderson: This line will cause problems if the user has not set up DNS for all environments.  Also, do we know if users typically use a DNS wildcard for their dev environments, or whether they add one record per environment?  If the later, then they would need to ensure that they always updated DNS every time they created a new multidev environment.

It might be worth a couple of words of warning here.  Usually folks do not set up their dev environments to redirect (I think -- maybe you have more information on this).  Perhaps the best thing would be to comment out this line, and add a comment that this should be customized iff needed for dev environments.

Finally, I'm not sure that ".dev" is the best example TLD to use in the document.  Some devs set up DNS on their dev machines such that "*.dev" == localhost.  Others use "*.local" for this same purpose, so this is not universal.  I'm not sure what folks do for their Pantheon environments, but I would expect that most of the time they would set up 'test.usenix.org' as a CNAME for their pantheonsite.io domain.

PR #1817, opened 2016-08-18
Add notice about Drush version

joshkoenig:
A lot of people with older sites still using Drush 5 are going to run into errors in the Dashboard when they go to PHP 7 (e.g. clear cache, update.php) if they don't also upgrade their Drush version.


nataliejeremy:
Thanks for that update @joshkoenig

PR #1820, opened 2016-08-18
Update power-users.md

hengkit:
twitchy OCD edit....

Changing an `an` to a `a`


nataliejeremy:
Thanks for catching that @hengkit

PR #1837, opened 2016-08-24
Use json_decode instead of json_load

eojthebrave:
## Effect
- update to code example, switches json_load for valid PHP function json_decode()

json_load is not a valid php function. php.net/json_load. The example here should probably use [json_decode()](https://php.net/json_decode) instead.


ttrowell:
Someone had Python on their mind, maybe!  
+1 thanks!!!

PR #1784, opened 2016-08-04
Update nginx-access-log.md

ttrowell:
## Effect
PR includes the following changes:
- update the snippet needed for formatting

update snippet to keep goaccess happy and working

@rachelwhitton Pleas review! :)

PR #1841, opened 2016-08-25
Update goaccess log-format and html generation syntax

albertcausing:
## Effect

Change Sequence of time, date and log format variable in reference to goaccess.conf
Change log-format according to our existing nginx-access.log
Added syntax how to generate the report.html output file.
Added troubleshooting guide, when goaccess.conf not detected.

### DIFF
--- a/source/_docs/nginx-access-log.md
+++ b/source/_docs/nginx-access-log.md
@@ -42,8 +42,10 @@ If you would like to generate an HTML report, execute the following commands:
 goaccess -f nginx-access.log -a -o report.html
 open report.html
 ```
-## Troubleshooting Guide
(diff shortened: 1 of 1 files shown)


rachelwhitton:
Hey @albertcausing can you give some context around this change? 
## Report using new log format 
reporttimeformat2
There are quite a few rows where the value is 0
## Report using existing log format
reportlogformat2

We recently updated the format in PR #1784 so I'd like to confirm the report generated is expected. Thanks for the PR!

ttrowell:
+1!

I tested this out and the new format adds more good detailed columns to the output.  It now works MOAR better!  Thanks @albertcausing !

nataliejeremy:
Thanks for the update! @albertcausing and @ttrowell

PR #1842, opened 2016-08-25
Correct WordPress files path

ataylorme:
In step 2, importing code, the WordPress files path should be `wp-content/uploads` not `wp-uploads`.


ari-gold:
+1 Thanks for the correction @ataylorme !

PR #1854, opened 2016-08-31
Update quicksilver.md

twooten:
Add a dash before "type".


rachelwhitton:
Thanks for the PR @twooten! Going to close this and merge #1855 since that one corrects both occurances - keep the PR's comin though! Much appreciated

PR #1859, opened 2016-09-01
Added two PHP 7 reference links

peter-pantheon:
## Effect
- Added two reference links for PHP 7 that illuminate some good things to know for developers.


nataliejeremy:
Thanks @peter-pantheon I removed the link to the blog as we typically don't link to external blogs (could easily lead to broken links)

PR #1868, opened 2016-09-06
Update multidev.md

newtoid:
## Effect
added more reserved words


nataliejeremy:

PR #1873, opened 2016-09-08
Correcting --destination and --exclude path

albertcausing:
## Effect

Initially @ruby found the bug and we've verified so here's our PR :D

Fixing --destination= path: 
`code/sites/default/files/<RANDOM_HASH>.tgz` instead of `/sites/default/files/<RANDOM_HASH>.tgz`

Another thing is on --exclude= path: 
`code/sites/default/files` instead of `sites/default/files`

POC: 
- https://monosnap.com/file/SCuayzJOmsmyph3XvJbTdewTC56vq9 without `code/`
- https://monosnap.com/file/1j1ufJiZRNev9lZW9u0QLPJNVDsk0n with `code/`


nataliejeremy:
Thanks for the update @albertcausing

rachelwhitton:
Thanks @albertcausing and @rubelyn! Related bug card: BUGS-687

PR #1879, opened 2016-09-09
Add warning about migration duration

slashzero:
Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed

### DIFF
--- a/source/_docs/migrate-wordpress.md
+++ b/source/_docs/migrate-wordpress.md
@@ -39,12 +39,6 @@ From your Pantheon Dashboard:
 
  ![Successful Migration BlogVault](/source/docs/assets/images/successful-site-migration-complete-blogvault.png)
 
-
-<div class="alert alert-info" role="alert">
-<h4>Note</h4>  
-The migration process may take several hours to complete, this is normal. You will be alerted via email once the migration process is complete.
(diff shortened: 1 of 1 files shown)


ari-gold:
@slashzero How about adding this as a FAQ at the bottom instead of a big warning on the top?

slashzero:
@ari-gold Sounds good, there will be other issues that'll come up

ari-gold:
I'll pull data on min/max/avg transfer time. Maybe we can look at that and update to say something like "Time to migrate your site depends on the site size. Most transfers are completed within X time, but if your site has a very large footprint it can take X time."

nataliejeremy:
@ari-gold Were you able to pull the data? We'd like to get this one merged. Thanks!

ari-gold:
Avg transfer time is just over an hour. This is the email sent out to folks when their migration is started:

> Most migrations to Pantheon are completed within 2 hours. Migration times depend fully on the size of your site, so be aware that it may take more or less time than estimated. Expect another email from us once your migration is complete. On the off-chance there are hiccups along the way, we will notify you by email.

nataliejeremy:
Thanks! @slashzero and @ari-gold

PR #1888, opened 2016-09-14
Add detail to branch naming restrictions

icylace:
When a developer pushes a branch to Pantheon and tries to create a Multidev out of it the attempt will fail if the branch name contains any uppercase letters.


rachelwhitton:
+1 Thanks @icylace!

PR #1895, opened 2016-09-21
Missing article before "linux container"

therealpadams:
Text did not read well because of a missing article. I opted to pluralise "linux container" rather than insert an article because the language following this text makes reference to a multitude of containers rather than _the/a_ container.

Also: I just wanted to play with @rachelwhitton 's integration between Pantheon's docs and GitHub :)


nataliejeremy:
Thanks for catching that! @therealpadams

PR #1903, opened 2016-09-28
Add example of site parameter in backup command

kevinhowbrook:
Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed

Site parameters missing from backup command, bit overkill, but running the command without them may cause unexpected behavior. It may also be worth adding the full available parameters here: eg terminus site backups <get|load|create|list|get-schedule|set-schedule|cancel-schedule> [--site=<site>] [--env=<env>] [--element=<code|files|db|all>] [--to=<directory|file>] [--file=<filename>] [--latest] [--keep-for] [--day] [--username] [--password] [--database]

### DIFF
--- a/source/_docs/local-development.md
+++ b/source/_docs/local-development.md
@@ -78,9 +78,10 @@ Replace <code>database.sql.gz</code> with the name of the database archive downl
 Create and export the database by running the following Terminus commands:
 
 ```nohighlight
-terminus site backups create --element=database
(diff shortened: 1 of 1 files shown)


rachelwhitton:
+1 We should do this throughout. Initially, they were removed to encourage use of dotenv files but i think it's better to show required arguments as a minimum. 

Thanks for the PR @kevinhowbrook! We'll audit the rest of the commands in a follow-up issue.

PR #1918, opened 2016-10-05
Fix list of Drush commands that don't work

icylace:
## Effect

- Replaced "sql-sqlc" which doesn't exist with "sql-cli" (a.k.a. "sqlc") which does.
  - https://drushcommands.com/drush-8x/sql/sql-cli/
- Added commanded aliases.

### DIFF
--- a/source/_docs/drush.md
+++ b/source/_docs/drush.md
@@ -364,7 +364,8 @@ See the [Drush Migrate documentation](https://drupal.org/node/1561820) for detai
 
 ## Known Limitations
 - Crontab: Currently, there is no way to manage Crontab on Pantheon. If you need a way to set up your own Cron interval, you can use an external cron service such as [Easy Cron](https://www.easycron.com/user/register).
-- The following Drush commands are not supported and will not work on Pantheon sites: `sql-sync-pipe`, `sql-cli` (`sqlc`), `php-eval` (`eval`, `ev`). An alternative to `drush sql-sync` is to use `drush sql-dump` instead.
+- The `sql-sync-pipe` and `sql-cli` (`sqlc`) Drush commands are not supported and will not work on Pantheon sites. An alternative to `drush sql-sync` is to use `drush sql-dump` instead.
+- The `php-eval` (`eval`, `ev`) command partially works on Pantheon since `pantheonssh` rejects any argument that contains characters such as `;` or `$`. As an example, `drush @pantheon.site.env ev 'return foo()` works because Drush provides a trailing `;` if needed.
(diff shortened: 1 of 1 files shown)


rachelwhitton:
Thanks for the update @icylace! @greg-1-anderson thoughts?

greg-1-anderson:
Looks good to me.

One thing of note is that `php-eval` partially works on Pantheon -- that is, it works just fine, except for the fact that pantheonssh rejects any argument that contains characters such as `;` or `$`.  So `drush @pantheon.site.env ev 'return foo()` works, because Drush will provide a trailing `;` if it is needed.

Just thought I'd mention that, in case you thought it should be documented.  It's probably simpler to just leave it as is, with php-eval being listed as "unsupported", though.

icylace:
I'm all in favor of adding as much detail and clarity about edge cases as possible.

rachelwhitton:
+1 Updated, thanks @icylace and @greg-1-anderson!

greg-1-anderson:
I tried to live-edit this PR, but it didn't seem to work right.  I'd suggest the following adjustment to @rachelwhitton 's text:
```
- The `php-eval` (`eval`, `ev`) command only partially works on Pantheon. Pantheon rejects any incoming command-line request that has an argument or option that contains characters such as `;` or `$`. As an example, `drush @pantheon.site.env ev 'return foo()` works, because Drush provides a trailing `;` if needed.
```

greg-1-anderson:
LGTM!

source/_docs/drush.md
```diff
@@ -364,7 +364,8 @@ See the [Drush Migrate documentation](https://drupal.org/node/1561820) for detai
 
 ## Known Limitations
 - Crontab: Currently, there is no way to manage Crontab on Pantheon. If you need a way to set up your own Cron interval, you can use an external cron service such as [Easy Cron](https://www.easycron.com/user/register).
-- The following Drush commands are not supported and will not work on Pantheon sites: `sql-sync-pipe`, `sql-sqlc`, `php-eval`. An alternative to `drush sql-sync` is to use `drush sql-dump` instead.
+- The `sql-sync-pipe` and `sql-cli` (`sqlc`) Drush commands are not supported and will not work on Pantheon sites. An alternative to `drush sql-sync` is to use `drush sql-dump` instead.
+- The `php-eval` (`eval`, `ev`) command only partially works on Pantheon. Pantheon rejects any incoming command-line request that has an argument or option that contains characters such as `;` or `$`. As an example, `drush @pantheon.site.env ev 'return foo()` works, because Drush provides a trailing `;` if needed.
```
rachelwhitton: @joshkoenig thoughts?

joshkoenig: I'd rather direct users to using `drush php-script` instead. The degree to which `php-eval` is impeded makes it functionally useless, and it will only frustrate people vs having them write a script and execute it.

PR #1929, opened 2016-10-08
Adding common query parameters link

allenfear:
Adding a link to make it easier for folks to configure the understand how to specify and use a query parser that is not our default parser.


rachelwhitton:
Thanks @allenfear! +1

PR #1938, opened 2016-10-17
Add the fact that `api_version` is required

icylace:
## Effect

- Adds to the troubleshooting section about the need to have `api_version` in every `pantheon.yml` file.


rachelwhitton:
+1

PR #1967, opened 2016-11-07
Update change-management.md to show developer site role is Enterprise only

lauren-kelly:
Closes #

## Effect
- Clarification that only Enterprise Orgs can assign Developer Roles on Sites

## Remaining Work
- [ ] Check with @1dwaynemcdaniel to make sure this is accurate


lauren-kelly:
@rachelwhitton I just C&P'd that in from above and edited the message

rachelwhitton:
Nice!! Thanks for the PR @lauren-kelly! Much appreciated 

+1

rachelwhitton:
Actually, I checked with @1dwaynemcdaniel on this and we decided to revert. Seems as if that's not accurate @lauren-kelly

1dwaynemcdaniel:
This seems right to me.  Org level is correct above it and the individual site permissions is not available outside of an Enterprise org I believe

rachelwhitton:
Thanks @1dwaynemcdaniel! I added this directly on master within commit 2715a583cadf7620bd2c27fa8ca949086cf04c8a

PR #1974, opened 2016-11-10
Provide link to D.O issue regarding Global Redirect front page handler.

dustinleblanc:
@rachelwhitton might want to copy check :)


rachelwhitton:
+1 Thanks @dustinleblanc!

PR #1981, opened 2016-11-14
Recommend SSH key as workaround for un/pw failures across an Org

### LINKED ISSUE #1961: Authentication Update: Recommend SSH Keys
Re: [Developing on Pantheon Directly with SFTP Mode: Authenticating](https://pantheon.io/docs/sftp/#authenticating)
## Issue Description:
Any large agency that has multiple developers who login frequently via username/password will trigger failed logins for everyone else who works on the site, even when everyone is using the right password. Even one developer who logs in and out successfully 3 times rapidly, will block out everyone else.
## Suggested Resolution 
Consider noting within other docs as appropriate, but clarify the recommendation to use SSH keys over username/password 

Related: BUGS-683

cc @joshkoenig

rachelwhitton:
Closes #1961 

## Effect
- Add question on access issues for users across an entire Org when authenticating via un/pw, recommends SSH keys as a workaround.
- Stronger recommendation for SSH keys within the SFTP doc

@dustinleblanc can you review for accuracy?

cc @joshkoenig

### DIFF
--- a/source/_docs/organization-faq.md
+++ b/source/_docs/organization-faq.md
@@ -35,6 +35,12 @@ See [Role-Based Permissions & Change Management](/docs/change-management) for de
 When you or the administrators, team members, or developers in your agency [create sites](https://dashboard.pantheon.io/sites/create), you have the option of associating it with your organization. Once there, all members of the organization can access the Site's Dashboard from the Organization Dashboard.
 
 ## Support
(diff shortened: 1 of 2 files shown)


dustinleblanc:
I'm Dustin LeBlanc and I approve this message 🇺🇸  👍

PR #2002, opened 2016-11-29
Advise users to upgrade prior to changing PHP versions

greg-1-anderson:
Closes Jira issue: https://getpantheon.atlassian.net/browse/BUGS-620

## Effect
- Advises upgrading software before trying new PHP version to avoid potential problems with old code.

## Remaining Work
- [x] Confirm @joshkoenig wants this change
- [x] Docs team review wording / grammar

### DIFF
--- a/source/_docs/php-versions.md
+++ b/source/_docs/php-versions.md
@@ -10,6 +10,9 @@ Upgrading your site's PHP version will improve the security, performance, and su
 ## Verify Current PHP Versions
 Verify current PHP settings from the Site Dashboard by clicking **Settings** > **PHP version**.
 
+## Upgrade Your Site
+Before you change your PHP version, ensure that your CMS and all of your themes, modules and plugins are up-to-date. Older software is more likely to contain code that is incompatible with recent PHP versions.
+
 ## Configure PHP Version
 Manage PHP versions by committing a `pantheon.yml` configuration file to the root of your site's code repository. Your `pantheon.yml` file will look like the following:
 


joshkoenig:
This is fine. It's not the doc I want in response to the bug though.

I do not believe adding these lines to the middle of a doc page will cause many more people to update all the things before trying PHP 7. I expect people will continue to see the behavior that is described in the bug report, and have a serious "WTF" moment.

I think we should ideally have a doc page that is a HOWTO for debugging PHP 7 compatibility notices and errors, describing the processes we used to identify the causes of the various notices. This will be useful to the internet writ large, to our customers, and to CSE in answering questions.

Subsequently we can reference this doc page in a more informal blog post (which I am happy to write).

greg-1-anderson:
I'm happy to continue writing the doc page we discussed yesterday. As I was working through it, though, the information just felt very stale. If you think that this situation will in fact continue to be experienced by some folks, then I will document it.

I will submit that page in a separate PR.

dustinleblanc:
We should also be sure to mention that disabling opcache to stop notices is a performance tradeoff. Opcache exists for a good reason.

greg-1-anderson:
The opcache issue is https://github.com/pantheon-systems/documentation/issues/2005

We will definitely advise that disabling opcache should never be done in the live environment. Since Pantheon turns off notices entirely on live, there is little motivation to do this on purpose. To reduce the chances that someone might do it on accident, we could guard all of the examples with `if not live`

PR #2004, opened 2016-11-30
Be more explicit that /files is not web-addressable

greg-1-anderson:
Closes #n/a

## Effect
- Don't call the Valhalla mount-point `/files`, because that's confusing.

## Remaining Work
- [ ] Improve suggested text -- still needs some refining. @hengkit volunteered.  :)


rachelwhitton:
+1 lgtm! Thanks!

PR #2010, opened 2016-12-01
Fixes #2005: Describe cause of intermittent PHP notices

### LINKED ISSUE #2005: New doc on debugging PHP 7 compatibility notices and errors
Re: [BUGS-620](https://getpantheon.atlassian.net/browse/BUGS-620)

greg-1-anderson:
Closes #2005

## Effect
- Adds a new page on deprecated constructor notices.
- Adds a cross-link from the php errors page to the new page.

## Remaining Work
- [x] Review wording

Caused by notices generated during PHP code parsing (deprecated constructor notices) when opcache is enabled.

### DIFF
--- a/source/_docs/deprecated_constructor_notices.md
+++ b/source/_docs/deprecated_constructor_notices.md
@@ -25,7 +25,7 @@ This error is caused by a combination of two factors:
 
 To fix these errors, convert your class constructor to `__construct()`. See [deprecation notice for PHP-4-style constructors](http://php.net/manual/en/migration70.deprecated.php#migration70.deprecated.php4-constructors) for more information.
 
-## Finding PHP Files Containing 
+## Finding PHP Files Containing Parse-Time Notices
 
 To make it easier to find and debug intermittent notices, **temporarily** disable the opcache in your settings.php file:
 ```php
--- a/source/_docs/deprecated_constructor_notices.md
+++ b/source/_docs/deprecated_constructor_notices.md
@@ -1,0 +1,52 @@
+---
+title: Deprecated Constructor Notices in PHP
+description: Debug and fix "Deprecated Constructor" notices in your Pantheon site.
+categories: [developing]
+tags: [code, php]
+keywords: debug, php 7, php error, php errors, php fpm error log, fpm error log, php log, php execution, execute php, caching, cache, php exception, exceptions, errors
+---
+
+PHP notices are usually handled automatically by the Pantheon Platform as described on the page [PHP Errors and Exceptions](/docs/php-errors); however, occasionally a PHP notice might be emitted directly into the web page content.
+
+An example notice might look like this:
+```nohighlight
+Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; views_display has a deprecated constructor in /srv/bindings/46027a30ca4c4980a7188036eb2fcea5/code/sites/all/modules/views/includes/view.inc on line 2553 
+Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; views_many_to_one_helper has a deprecated constructor in /srv/bindings/46027a30ca4c4980a7188036eb2fcea5/code/sites/all/modules/views/includes/handlers.inc on line 753
+```
+
+When an error like this is encountered, it will usually only appear intermitantly; most page loads will come up correctly until, after a period of inactivity on the site, the same message will appear again.
+
+This error is caused by a combination of two factors:
+
+- PHP notices generated during source code parsing
+- PHP opcache is in use
+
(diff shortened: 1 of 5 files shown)


source/_docs/deprecated_constructor_notices.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+When an error like this is encountered, it will usually only appear intermitantly; most page loads will come up correctly until, after a period of inactivity on the site, the same message will appear again.
+
+This error is caused by a combination of two factors:
+
+- PHP notices generated during source code parsing
+- PHP opcache is in use
+
+Starting with PHP 7, a [deprecation notice for PHP-4-style constructors](http://php.net/manual/en/migration70.deprecated.php#migration70.deprecated.php4-constructors) is emitted if a method with the same name as its class is used. This was the recommended way to declare class constructors in PHP 4, but it was not until PHP 7 that this form actually started producing a notice. At this time, the PHP-4-style constructor is the only deprecation notice that is reported during source code parsing. In theory, this structure should only be encountered in very old code; in practice, though, some projects may have continued using the deprecated form, so it may be encountered from time to time.
```
ari-gold: How about removing these three paragraphs? Or putting them after the solution?

source/_docs/deprecated_constructor_notices.md
```diff
@@ -0,0 +1,52 @@
...
+
+Opcache is always enabled on Pantheon. When opcache is in use, PHP will print any notice emitted during source code processing directly to the standard output, ignoring any error handler that may be set. These notices may still be disabled by setting `error_reporting` to ignore E_DEPRECATED; however, Drupal always enables E_DEPRECATED in `error_reporting` early in its bootstrap process, so these notices will be printed if `display_errors` is set to `on`.
+
+Opcache also influences when this error is displayed. Since opcache caches the compiled form (opcodes) of the PHP that is being executed, the deprecation notices will not be printed when the PHP opcodes are fetched from the opcache. This is what leads to the intermittant nature of this problem. To make it easier to find and debug PHP 4 constructors, you may disable the opcache in your settings.php file:
```
ari-gold: and keep/tweak this line: 

To make it easier to find and debug PHP 4 constructors, **temporarily** disable the opcache in your settings.php file:

greg-1-anderson:
The deploy script failed with "you are not logged in" (unusual). The login failed with "The provided machine token is not valid." Acting as if I pushed to an untrusted branch, maybe?

Just re-run the build?

greg-1-anderson:
Despite the failure, it appears that the latest changes were deployed anyway.

source/_docs/deprecated_constructor_notices.md
```diff
@@ -0,0 +1,62 @@
...
+if (!isset($_ENV['PANTHEON_ENVIRONMENT']) || ($_ENV['PANTHEON_ENVIRONMENT'] != 'live'))
+{
+    ini_set("opcache.enable", 0);
+}
+```
+**IMPORTANT:** Disabling opcache has a sever impact on performance, so care should be taken not to do this on a production system. Note that `display_errors` is set to `off` on Pantheon live environments, so there is no motivation to disable opcache in production anyway. When disabling opcache in dev and multi-dev environments, be sure to re-enable it again once you are done debugging.
+
+Note that you may also search for deprecated constructors using the PHP linter on the command line. First, create a local copy of your site as described in [Local Development](/docs/local-development); then, run:
+```nohighlight
+$ php -l sites/all/modules/views/includes/handlers.inc
```
ari-gold: Does this require a local version of PHP 7.0?

source/_docs/deprecated_constructor_notices.md
```diff
@@ -0,0 +1,62 @@
...
+```
+Replace the path to the file you would like to check. To check every .php, .inc and .module file in a Drupal site, run:
+```nohighlight
+find -E . -iregex ".*\.(inc|php|module)$" -exec php -l {} \; | grep -v 'No syntax errors'
+```
+Adjust the regular expression as needed to scan other file extensions that may contain php code.
+
+## Explanation
+
+Deprecation notices for deprecated constructors are emitted if a class contains a method with the same name as the class name. This was the recommended way to declare class constructors in PHP 4, but it was not until PHP 7 that this form actually started producing a notice. At this time, the PHP-4-style constructor is the only deprecation notice that is reported during source code parsing. In theory, this structure should only be encountered in very old code; in practice, though, some projects may have continued using the deprecated form, so it may be encountered from time to time.
```
ari-gold: Add an example of a class that contains a method with the same name as the class name?

ari-gold:
Also link to this new doc from https://pantheon.io/docs/php-versions/ ?

source/_docs/deprecated_constructor_notices.md
greg-1-anderson: Yes.

source/_docs/deprecated_constructor_notices.md
greg-1-anderson: Already linked to http://php.net/manual/en/migration70.deprecated.php#migration70.deprecated.php4-constructors above; that page contains an example.

greg-1-anderson:
Hm, maybe. I already link here from https://pantheon.io/docs/php-errors/, which has a link on the php-versions page. I tend to think that this situation is rare enough that it wouldn't be worthwhile to put it in the TOC-like entry at the bottom of php-versions. If you think it should be added, I'll put it in, though.

rachelwhitton:
+1 Thank you @greg-1-anderson! Thanks for the review @ari-gold

greg-1-anderson:
I'm kind of neutral on @ari-gold 's last couple of suggestions. Happy to put them in if @ari-gold feels strongly, or @rachelwhitton wants them.

PR #2016, opened 2016-12-06
Legacy DNS settings issue

peter-pantheon:
## Effect
- Clarifies legacy DNS settings vs current DNS settings recommendations

## Remaining Work
- [x] Review and polish copy changes

An old site may be using legacy DNS recommended settings that no longer match the current recommended settings displayed in the Dashboard.


newtoid:
+1 here. contemplating including old IPs but that would likely lead to more confusion.

rachelwhitton:
Thanks @peter-pantheon! Yeah, I thought about that too @newtoid. I think it's best to leave them out, so we steer others clear of using outdated values.

PR #2018, opened 2016-12-06
Set Hash Salt

### LINKED ISSUE #1713: Setting drupal_hash_salt value need review
Re: https://pantheon.io/docs/settings-php/
## Description

We have no explanation or information regarding setting the $drupal_hash_salt value in D7 via settings.php.  A customer reported issues trying to override it, so we need to test and provide the process.

When some sites are migrated over, they need to retain a previous hash value so that DB values don't break.  This is not terribly common.
## Suggested Resolution

Document a working process 

Customer used this to override Pressflow:

```
// Force old hash salt settings 
$pf = json_decode($SERVER['PRESSFLOWSETTINGS']); 
$pf→drupal_hash_salt = 'xxxxxxxxRXvuWVkAlyxxxxxxxxxxxxxxxxxxxx_A4';
$SERVER['PRESSFLOW_SETTINGS'] = jsonencode($pf);
```

Case:  64974

peter-pantheon:
Closes #1713 
## Effect
- Indicate how to set hash salt for D7

## Remaining Work
- Review for clarity, language

Set hash salt in settings.php Drupal 7

### DIFF
--- a/source/_docs/settings-php.md
+++ b/source/_docs/settings-php.md
@@ -226,6 +226,21 @@ It depends on your site configuration. Stripping commented-out or non-functional
 
 Pantheon automatically injects database credentials into the site environment; if you hard code database credentials, you will break the Pantheon workflow.
 
+#### Where do I set or modify the drupal_hash_salt value in Drupal 7?
+
+There can be an occassion when you may need to set the hash salt to a specific value. If you install Drupal 7, it will create a drupal_hash_salt value for you, but if you want to use a different one, you can edit settings.php before installation. Pantheon uses Pressflow to automatically read the environmental configuration and the Drupal 7 hash salt is stored as part of the Pressflow settings. 
+```   
+   // All Pantheon Environments.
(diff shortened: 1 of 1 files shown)


source/_docs/settings-php.md
```diff
@@ -226,6 +226,21 @@ It depends on your site configuration. Stripping commented-out or non-functional
 
 Pantheon automatically injects database credentials into the site environment; if you hard code database credentials, you will break the Pantheon workflow.
 
+#### Where do I set or modify the drupal_hash_salt value in Drupal 7?
+
+There can be an occassion when you may need to set the hash salt to a specific value. If you install Drupal 7, it will create a drupal_hash_salt value for you, but if you want to use a different one, you can edit settings.php before installation. Pantheon uses Pressflow to automatically read the environmental configuration and the Drupal 7 hash salt is stored as part of the Pressflow settings.
```
rachelwhitton: What happens if they edit this after installation? or if they’re migrating an existing site to the platform?

PR #2024, opened 2016-12-08
D8 $base_url clarification

peter-pantheon:
## Effect
- Clarify that $base_url remains a global in D8

## Remaining Work
- Review for clarity, language, form. Is this a worthwhile note?


rachelwhitton:
I think since it's still available, it's not worth noting.

PR #2040, opened 2016-12-15
Explicitly call out changes needed to run in full (strict)

dustinleblanc:
## Remaining Work
- [x] @rachelwhitton Copy/markdown edits. I don't think I worded this well and it might be better expressed as a note or warning.

If a customer wants to run in full (strict) mode with Cloudflare, they need to make sure to use A records pointed at their load balancer and not the CNAMES that we typically recommend with Cloudflare.


rachelwhitton:
+1 Thanks @dustinleblanc!

PR #2042, opened 2016-12-15
 Document Browser Support

rachelwhitton:
## Effect
- New section within Platform Considerations: **Browser Support for Pantheon's Dashboard**
- Tooltip for IE11 support: **Minimum Support** - Feature support only, allows for visual regression. 

cc @slivermon @drainpip @bensheldon @JonathanPorta


drainpip:
At this time we have a thumbs up from everyone, but I had a question from a documentation perspective. Does everyone know what a visual regression is, or should it be something like "visual inconsistencies" (for example)? If this is used elsewhere please ignore and push at your leisure 😄

PR #2045, opened 2016-12-16
Correct WordPress information

cyberhobo:
The snippet given here doesn't work in `wp-config.php` as suggested, because the `add_action` function is not yet available when `wp-config.php` is loaded.

It's more likely you would do this in a site-specific plugin. If that plugin lives in `mu-plugins`, the action needs to be added at a priority later than 10 to be sure it runs after the Pantheon caching plugin action.


lauren-kelly:
+1 @rachelwhitton

rachelwhitton:
+1 Thanks @cyberhobo! Much appreciated

PR #2050, opened 2016-12-21
SSL resource that works with GoDaddy

kveldalf:
Lauren noted that [SSL.com](https://www.ssl.com/online-csr-and-key-generator/) seems to be a generator that works with GoDaddy.


lauren-kelly:
@rachelwhitton I incorporated this change in the other PR for this doc - let me know if you want me to go take that back out of there

PR #2051, opened 2016-12-21
Note about not entering password

kveldalf:
Suggested change moves note to not add a password higher in the doc, closer to the command and description. Also includes a note to run process again if a password was added the first.

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed


rachelwhitton:
+1 Thanks @kveldalf!

PR #2059, opened 2017-01-03
Add info on how to contact support via email

rubelyn:
Added email contact (helpdesk@pantheon.io) so Customers can still reach us if they having trouble accessing the Dashboard login page.

### DIFF
--- a/source/_docs/getting-support.md
+++ b/source/_docs/getting-support.md
@@ -137,6 +137,8 @@ Access support history for chats in the User or Site Dashboards by selecting **S
 #### Can I start a chat session via email?
 No. Initiate chats via the Pantheon User or Site Dashboards.
 
+#### I can't access the Pantheon Dashboard login page, can I contact support via email?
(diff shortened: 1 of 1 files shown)


source/_docs/getting-support.md
```diff
@@ -137,6 +137,8 @@ Access support history for chats in the User or Site Dashboards by selecting **S
 #### Can I start a chat session via email?
 No. Initiate chats via the Pantheon User or Site Dashboards.
 
+#### I can't access the Pantheon Dashboard login page, can I contact support via email?
```
newtoid: I don't really think we should encourage this, maybe a form somewhere?

lauren-kelly: A form would be great - make it clear that is isn't to start a support ticket - other than to let us know they cannot access their dashboard.

rachelwhitton:
Closing this, to be handled in FEAT-332

PR #2060, opened 2017-01-03
Add local install to manual migration list

ataylorme:
We get asked about local installations on the migrations email fairly often so it's worth noting in the migration doc.


lauren-kelly:
+1

ataylorme:
@rachelwhitton I think calling it out specifically in the scenario list can't hurt. The FAQ is a ways down on the page. I honestly didn't read down that far and don't expect users to when glancing at the list of when they can/can't use the migration plugin.

rachelwhitton:
+1 Thanks @ataylorme!

PR #2062, opened 2017-01-04
Add Restrict Content Pro to known plugins

dustinleblanc:
Quick addition to known offenders based on customer chat issue.


rachelwhitton:
+1 Thanks @dustinleblanc

PR #2064, opened 2017-01-04
Update Drush examples and workarounds

greg-1-anderson:
Closes #n/a

## Effect
- Correct from `drush cc cache` (incorrect) to `drush cc drush` (correct)
- Use SITENAME.ENV consistently in examples
- Add examples of how to do sql querries and run php code remotely on Pantheon sites
- Update from `terminus sites aliases` (old) to `terminus aliases` (current)
- Remove deprecated example of using policy file to select Drush version (using pantheon.yml is better)
- Add troubleshooting tip about spun-down site environments
- Reformat unsupported commands and add links to workarounds

## Remaining Work
n/a

### DIFF
--- a/source/_docs/drush.md
+++ b/source/_docs/drush.md
@@ -36,7 +36,7 @@ Download your Pantheon site aliases to manually update your local aliases file:
 3. Clear Drush cache:
 
  ```
- drush cc cache
+ drush cc drush
  ```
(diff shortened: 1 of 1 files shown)


rachelwhitton:
+1 thanks @greg-1-anderson!

PR #2067, opened 2017-01-05
More "use New Relic" focus

joshkoenig:
Quick update and light re-org to make the New Relic doc page more centered on _using_ New Relic, in addition to troubleshooting problems that existing users might encounter.

- Added section with basic explanation and links to more in-depth guides
- Moved all troubleshooting sections to go under one header
- Removed some overly general "read more" links


rachelwhitton:
Thanks @joshkoenig!

PR #2074, opened 2017-01-09
Clarify cron run time

ilovelamp:
Clarifies that cron runs at generally within 5-10 minutes of half-past each hour.


rachelwhitton:
+1 Thanks @ilovelamp

PR #2078, opened 2017-01-09
Sidebar Scroll

### LINKED ISSUE #2043: Contents sidebar of docs cutoff
## Issue Description
The "Contents" sidebar of a doc is fixed. If the section has a height greater than the viewport sections of it are not visible. This causes it to be unusable as navigation.

## Example
[Modules and Plugins with Known Issues](https://pantheon.io/docs/unsupported-modules-plugins/)

docs-contents-sidebar-cutoff

## Suggested Resolution
Give the `overflow: scroll` attribute to the `div.pio-docs-sidebar` element. The JavaScript should also be updated so the sidebar scrolls as the user scrolls.

rachelwhitton:
Closes #2043 

## Effect
- Add `overflow: scroll` to improve usability of sidebar nav
- Adjust sidebar js so automatic scrolling matches user scroll

@ataylorme can you review and +1?

### DIFF
--- a/source/docs/assets/css/main.css
+++ b/source/docs/assets/css/main.css
@@ -1653,6 +1653,7 @@ margin-bottom: 40px !important;  }
 
 /* By default it's not affixed in mobile views, so undo that */
 .pio-docs-sidebar.affix {
+  overflow: scroll;
   position: static;
   max-height: 90%;
 }
--- a/source/docs/assets/js/jquery.tocify.js
+++ b/source/docs/assets/js/jquery.tocify.js
@@ -85,7 +85,7 @@
 
             // **scrollTo**: Accepts Number (pixels)
             // The amount of space between the top of page and the selected table of contents item after the page has been scrolled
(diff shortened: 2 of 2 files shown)


ataylorme:
@rachelwhitton on https://2043sideba-static-docs.pantheonsite.io/docs/unsupported-modules-plugins/ I am able to scroll the sidebar (yay!).

You can scroll the sidebar so the highlighted item is out of view. When the main page is scrolled the highlighted item stays out of view. Can the sidebar scroll to the current item when the main content scrolls?

jackfoust:
Just wondering, would this be better addressed with an overflow-y: auto? Short sidebars have X and Y scrollbars even though they are not scrollable. See https://pantheon.io/docs/pantheon_stripped/#issue-pantheon_stripped-displays-in-the-utm_source-url-parameter-in-google-analytics

jackfoust:
Strike that. Does not work as intended at the end of a long sidebar.

PR #2092, opened 2017-01-12
Update Terminus syntax for WP Cron

dustinleblanc:
Terminus no longer works appropriately if you quote the command string passed to `terminus wp`.

## Remaining Work
- [x] check for other references to Terminus commands that use quotes.

### DIFF
--- a/source/_docs/wordpress-cron.md
+++ b/source/_docs/wordpress-cron.md
@@ -45,7 +45,7 @@ You can also schedule your own jobs, execute existing jobs, and manage just abou
 One of the first things you'll want to do is test WP-Cron to make sure everything is working correctly. When you execute the command below, make sure to replace SITE_NAME with your site's name from your Pantheon Dashboard and replace ENV_NAME with the desired environment ("dev", "test", "live", or multidev branch name).
 
 ````nohighlight
-$ terminus wp <site>.<env> cron test
+$ terminus wp <site>.<env> -- cron test
 ````
 
 If everything works correctly, the result looks like this:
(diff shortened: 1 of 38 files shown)


source/_docs/wordpress-cron.md
```diff
@@ -45,7 +45,7 @@ You can also schedule your own jobs, execute existing jobs, and manage just abou
 One of the first things you'll want to do is test WP-Cron to make sure everything is working correctly. When you execute the command below, make sure to replace SITE_NAME with your site's name from your Pantheon Dashboard and replace ENV_NAME with the desired environment ("dev", "test", "live", or multidev branch name).
 
 ````nohighlight
-$ terminus wp <site>.<env> "cron test"
+$ terminus wp <site>.<env> cron test
```
greg-1-anderson: This should be `terminus wp <site>.<env> -- cron test`

The `--` can be anywhere, and is not necessary at all if there are no options. However, I have my eye on using it as a hint to disambiguate <site>.<env> from the next parameter. Never a problem for wp-cli (site.env must always have a `.`, and no wp-cli commands contain a `.`), but it may be important in other instances, so let's start training users early.

greg-1-anderson: The `--` is relevant even today, as it allows the command to keep working if the user copies the example and adds an option.

dustinleblanc:
@greg-1-anderson just corrected, the previous version worked in this case but you are correct that it is better for our docs to show the explicit and 'stricter' example to guard against modifications by the user that could cause failure. I was still a touch hazy on how to do that the best way so my brain defaulted to the simpler answer 😄

rachelwhitton:
+1 Thanks @dustinleblanc!

PR #2101, opened 2017-01-17
Add ewww_image_optimizer

slashzero:
Stores absolute image path in db

### DIFF
--- a/source/_docs/unsupported-modules-plugins.md
+++ b/source/_docs/unsupported-modules-plugins.md
@@ -274,9 +274,7 @@ For more details, see [SERVER_NAME and SERVER_PORT on Pantheon](/docs/server_nam
 <hr>
 
 ### [EWWW Image Optimizer](https://wordpress.org/plugins/ewww-image-optimizer/)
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
I'm curious what the "container ID" is in the file paths stored here. Is this the site uuid, or is this some internally-created ID value?  Maybe something calculated based on the database credentials?

rachelwhitton:
@greg-1-anderson Refers to the application container's binding ID, which is created internally.


@slashzero If multiple application containers are serving the site, would the resolution here work?

greg-1-anderson:
Ah, I see; the code is using something like `realpath` and writing the full absolute path to the file into the database.  It seems like the proposed description of this situation in this PR is only applicable to sites with a single app server. I believe that @rachelwhitton is correct, and this plugin simply cannot be used if the site has multiple app servers.

rachelwhitton:
+1 I removed the recommended solution and clarified the issue for sites served by multiple app servers. Can you review @slashzero?

slashzero:
+1 plugin won't be able to work on multiple app servers.

PR #2108, opened 2017-01-18
adjust vagrant sync type

alexfornuto:
This change adjusts the way vagrant syncs files from the host. By using rsync intead of NFS Sculpin will recognize file changes and live-update.


rachelwhitton:
+1 Thanks @alexfornuto! :shipit:

PR #2109, opened 2017-01-18
Rewrite the performance test documentation

### LINKED ISSUE #1851: Load and Performance Testing Doc Update
Re: [Load and Performance Testing](https://pantheon.io/docs/load-and-performance-testing)
## Issue Description:

Title of doc is Load and Performance Testing, but doc intro is all about Load testing, so let’s revise it!

We should talk about 
1. **Load** vs **performance** testing, highlight value of smooth scaling and consistent infrastructure. Performance characteristics of Live are same as other environments so if it’s slow on Dev, don’t deploy Live. If you uncover performance issues with a single request, no need to conduct a load test, fix the performance issue first!
2. How to do performance testing in Dev/Test. Highlight New Relic
3. Load testing in Live environment **before launch**
4. Use math to calculate load capacity **after launch**. After launch, you can establish a baseline that X response time will let you handle Y traffic. If X degrades in Dev/Test, that will impact how much traffic Live can handle. We do not recommend load testing in Live after launch because you risk overwhelming your Live site and causing downtime. 

Add heading:
How to calculate how much traffic you can handle:

If you know traffic will increase,  you can understand the relationship between your site’s performance and how much load it can handle by doing a little math.

Maybe best way to communicate above is via a table, like this interactive spreadsheet? (https://docs.google.com/a/getpantheon.com/spreadsheets/d/1K9BdmdJDqHMvHMaerZZmCcflSvmEJhgBsGL8zwjpxE8/edit?usp=sharing)

For example, if your application's average response time is Xms and the Live environment gets Y requests per minute (rpm), then a Business plan with 16 php workers will handle the load. If you anticipate a traffic increase and you’d like to upgrade to Elite, then <contact Sales>.   (validate this math based on new pricing comparison doc)

Note that
Dev environments defaults to not cache static assets like CSS and JS to facilitate development. This will affect frontend performance of those environments, and you should be aware of that if you are perf testing there.

Also include link to https://pantheon.io/docs/guides/load-testing-with-blazemeter

bentekwork:
Closes #1851 

## Effect
Rewrite of performance test documenation

### DIFF
--- a/source/_docs/load-and-performance-testing.md
+++ b/source/_docs/load-and-performance-testing.md
@@ -4,197 +4,80 @@ description: Learn how to monitor internal execution performance of your Pantheo
 tags: [performance, cache]
 categories: [platform, cache]
 ---
-We highly recommend load testing a site both prior and post launch to ensure your site is optimally configured.
 
-## Before You Begin
+## Load vs Performance Testing
 
-You should:
+Before getting started testing your site its important to understand the difference between load and performance testing and know when to use each.
 
-- [Enable New Relic Pro](/docs/new-relic) to monitor internal execution performance without needing any additional modules or tools.
-- Have access to a command-line environment, preferably with administrative privileges.
-
-<div class="alert alert-info" role="alert">
-<h3 class="info">Note</h3>
-<p><strong>Load testing should only be performed on the Live environment</strong>. Dev has much lower default caching settings than other environments to facilitate iterative development. Test has the exact same configuration as Live, but Test can only have one appserver, while Live can have as many as your plan allows. If disruptive behavior occurs outside of the Live environment, the site may be temporarily disabled to prevent disruption to other customers.</p></div>
-
-## Performance vs. Scalability
-
-There are two things to test for:
-
-1. **Performance**: the response time for an individual request
-2. **Scalability**: the ability to deliver with optimal response time to a larger number of concurrent requests
-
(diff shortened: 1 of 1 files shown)


source/_docs/load-and-performance-testing.md
```diff
@@ -4,197 +4,80 @@ description: Learn how to monitor internal execution performance of your Pantheo
...
+ * <a target="_blank" href="http://locust.io/">Locust.io</a>
+
+The Pantheon onboarding team uses Locust.io for load testing.  The benefit of locust is that is open source, it is fairly easy to build out test scripts and it allows you to crawl the site instead of using predefined urls for the load test.  Crawling the site has the added benefit of hitting every page that is linked to anywhere on the site.  This means that a lot of edge case performance bottlenecks can be found where other more rigid tests would have not.
+
+Ultimately, is doesn't matter what tool you decide to use as long as it allows you to test your site properly.  Be sure to allow for any authenticated traffic as well as anonymous.  Which brings the next step, how to determine how much load to apply.  
+
+For performance tests a smaller load should suffice as you should be able to see transactional bottlenecks with 10-20 concurrent users.
+
+For load tests, to determine how many concurrent users you must start with looking at your historical analytics of your site.  Try to find the peak hourly sessions of your site and the average session duration.
```
rachelwhitton: @bentekwork how do I determine load to apply in the test after calculating concurrent users?

rachelwhitton:
Moving from fork to a branch for easier peer review and copy edits. See 2140

PR #2133, opened 2017-01-24
How 'bout we sudon't?

joshkoenig:
Provide alternative to `sudo` for installation.

Also add a notice about previous installations getting in the way.


rachelwhitton:
+1 Thanks @joshkoenig!

PR #2167, opened 2017-02-03
Add Terminus Push Code plugin

pbuyle:
Add Terminus Push Code plugin

## Blocked By
- [ ] Quality standards for inclusion set by eng and documented in the Terminus guide (tracked internally

### DIFF
--- a/source/_docs/terminus/plugins/directory.md
+++ b/source/_docs/terminus/plugins/directory.md
@@ -89,6 +89,17 @@ The following plugins are supported for Terminus 1.0:
     </div>
   </div>
 </div>
+ <div style="margin-bottom:30px;" class="col-md-4">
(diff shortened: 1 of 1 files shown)


rachelwhitton:
Thanks for the PR @pbuyle!! We're holding off including more 3rd party plugins until we formalize review procedures and quality standards. We'll let you know ASAP!

greg-1-anderson:
@pbuyle also posed the question as to whether push code should be part of the Terminus Build Tools plugin. I don't have a firm answer on that yet, but it seems like it might be a good fit.

We should circle back here once we have established the review procedures for external plugins and make a decision about whether it will remain separate or be integrated.

alexfornuto:
Closing this in favor of a single issue to track all plugins awaiting entry: #2895

PR #2172, opened 2017-02-07
Update snippet to redirect all traffic to HTTPS

ari-gold:
remove HTTP_X_FORWARDED_PROTO until we can guarantee it will work in all cases

BUGS-1034


rachelwhitton:
+1 Thanks @ari-gold

PR #2173, opened 2017-02-07
rewording spin down notes

ilovelamp:
## Effect
- Reword and mention SFTP and MySQL may fail when site is spun down.


rachelwhitton:
+1

ilovelamp:
Good call on the wake, I thought about adding it.

But, no "copy" button @rachelwhitton ? :)

ilovelamp:
@rachelwhitton - Can you also put a named anchor on that "Note" section? We link to the page a lot and an anchor would help.

PR #2178, opened 2017-02-09
timezone and log timestamps in UTC

ilovelamp:
## Effect
- Server time zone and log file timestamps are in UTC.


rachelwhitton:
+1!

PR #2195, opened 2017-02-17
Custom upstream vanilla site usage example

ilovelamp:
https://pantheon.io/docs/custom-upstream/

Updated to demonstrate how to create vanilla WordPress / Drupal site when setting up custom upstream.


rachelwhitton:
+1 Thanks @ilovelamp!

PR #2202, opened 2017-02-22
Add issue #3 for Wordfence

slashzero:
Writes .user.ini to docroot with absolute server erference.

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed

### DIFF
--- a/source/_docs/unsupported-modules-plugins.md
+++ b/source/_docs/unsupported-modules-plugins.md
@@ -181,7 +181,7 @@ Customers have also reported success by making the export path [configurable](ht
 
 **Solution**: Follow the remedy provided within the [module's documentation of the issue on Drupal.org](https://www.drupal.org/node/1306214), which is to alter the code to prefix the cookie name with `SESS`.
  <hr>
- 
+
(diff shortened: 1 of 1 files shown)


rachelwhitton:
@slashzero thanks for the PR! It may be best to work with the plugin authors to get the two issues on absolute paths addressed rather than having the plugin listed as entirely unsupported. If we create the issues, we can link back here so folks can check progress.

We should submit one, possibly two or even three issues to the [Wordfence support page on WordPress.org](https://wordpress.org/support/plugin/wordfence/) requesting that `auto_prepend_file` is set relative to the plugin's parent dir and that the log file path be set relative to the plugin's parent dir within the standard location (`wp-content/uploads`). We should also report that the Live Tracking feature conflicts with edge, but that problem may be more difficult to resolve so we may have to keep that as an unsupported feature. 

thoughts @ttrowell and @slashzero?

rachelwhitton:
https://wordpress.org/support/topic/write-logs-to-the-standard-file-path/
https://wordpress.org/support/topic/set-auto_prepend_file-path-relatively/

rachelwhitton:
+1

PR #2210, opened 2017-02-27
Update pantheon-workflow.md

fusionx1:
Added troubleshooting Item


rachelwhitton:
Thanks for the PR @fusionx1! Can you add an example of the Redis error discussed? Either via a comment here on the PR or as a follow up commit - whatever you prefer, thanks!

rachelwhitton:
@fusionx1 I made the changes we discussed in slack, thanks again for the PR! 

+1

PR #2213, opened 2017-03-02
Silent Failures From Redirects

rachelwhitton:
Re: AL-847

## Effect
- Add section: Command Line Conditionals (above redirect examples) so users are informed at the time they're adding redirect logic. 
- Add troubleshooting sections to Drush and WP-CLI docs 

@ataylorme @sehamilton


ataylorme:
+1 - @rachelwhitton thanks for the descriptive examples for both WP-CLI and Drush!

PR #2218, opened 2017-03-06
Fix typos about hash salt, reword.

dustinleblanc:
This entry was confusing to some users and the code snippet had some typos.


rachelwhitton:
+1

Issue #2191, opened 2017-02-15
HTTPS Cert Formatting


alexfornuto:
Re: [Enable Secure HTTPS Communication](https://pantheon.io/docs/enable-https/)
## Issue Description
As described in bug 1030, customers may include additional SSL cert/key formatting not accepted by the dashboard. 

## Suggested Resolution
Specify what formatting is (and more importantly isn't) allowed around SSL certs and keys.

garydylina:
Adding BUGS-1109 to the list of Bugs that would benefit from this clarification.

alexfornuto:
Resolved by #2219

PR #2219, opened 2017-03-06
Adding example with cert header and footer

allenfear:
Closes HTTPS Cert Formatting #2191 

## Effect
Adds example of an SSL cert so that users can see what a header and footer looks like.

## Remaining Work
n/a


alexfornuto:
👍

PR #2221, opened 2017-03-06
Serve traffic from your preferred domain (www or non-www)

ari-gold:
* Add "Serve traffic from your preferred domain (www or non-www)" section
  - Method 1: Add redirect from WordPress or Drupal
* Switch order of FAQ and Troubleshooting
* Add links to https://mxtoolbox.com/ and https://www.whatsmydns.net/
* Minor edits

Relates to BUGS-1040

### DIFF
--- a/source/_docs/domains.md
+++ b/source/_docs/domains.md
@@ -60,16 +60,9 @@ From the target environment's Domains tool, click **Show DNS Recommendations** n
 While it’s good for visitors and DNS to resolve both www and non-www domain, it's best practice to choose one or the other and redirect from www to non-www (or vice versa, your call). This optimizes SEO by avoiding duplicate content and prevents session strangeness, where a user can be signed on one domain but logged out of other domains at the same time.
 
 ### Serving sites from www
+Using the provided destinations in the Site Dashboard, create the recommended DNS entries at the domain's DNS provider.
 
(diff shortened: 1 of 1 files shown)


ari-gold:
@lauren-kelly  Can you please review?

rachelwhitton:
@ari-gold @lauren-kelly are there any objections to removing the www-redirection service entirely from this doc and encouraging all users to redirect to a common domain within their sites configuration? I think new users would find the methods confusing (especially for the www section). It's hard to tell which one applies to HTTP

allenfear:
@ari-gold it makes sense to me that we would stop surfacing the www-redirection service as an option at this point if it will not work with the Global Edge. Does that sound right to you?

lauren-kelly:
Yeah - I think everyone I've sent the info to opted to just go ahead and add the redirect snippet. It makes more sense to keep all that logic together, I think - should someone take over a site or whatever.

rachelwhitton:
@lauren-kelly can you review? Updated to remove the www-redirection service.

ari-gold:
Haven't had a chance to review the changes, but **the www-redirection service will work for any site that is not using HTTPS**. It's still being recommended on some site dashboards, so might be good to keep it in. I'm also okay with removing it.

PR #2232, opened 2017-03-09
improve FAQ section  - create subsection 

rubelyn:
make FAQ items clickable so it can easily be shared in desk ticket


rachelwhitton:
+1 Thanks @rubelyn!

PR #2237, opened 2017-03-13
Update sendgrid.md

ttrowell:
From Nick:  Okay, I have isolated the issue. SendGrid breaks specifically when a plugin called "ilightbox" is installed (https://codecanyon.net/item/ilightbox-revolutionary-lightbox-for-wordpress/3939523).

This is a beautiful plugin that really handles photo galleries very well. I've contacted the author about the incompatibility, but have not yet heard back. I'm not sure how many people this would impact, but I suspect that Pantheon would have a vested interest in finding out WHY this incompatibility exists (i.e. what causes it) because it is likely the case that anyone having problems with SendGrid's plugin has another plugin installed that suffers from the same issue.

## Effect
- Adds a Troubleshooting section to the Sendgrid page

## Remaining Work
- [ ] Find out what exactly breaks the Sendgrid config


NicholasRBowers:
@iprodev is the developer of iLightbox. Tagging here for visibility.

rachelwhitton:
+1 Thanks @ttrowell and @NicholasRBowers!

PR #2245, opened 2017-03-15
Custom Upstreams: Fix Service Levels

rachelwhitton:
Re: AL-862

## Effect
- Correct the list of service levels eligible for the custom upstreams feature. 

cc @sehamilton


rachelwhitton:
+1 in AL-862

PR #2238, opened 2017-03-13
Adapt SimpleSAMLphp doc to include WP configuration details

danielbachhuber:
PR includes the following changes:
- Expands scope of document to include WordPress, in addition to Drupal.
- Updates referenced SimpleSAMLphp version to 1.14.11 (latest stable).
- Includes literal steps for downloading SimpleSAMLphp latest stable.
- Includes WP environment settings and basic configuration instructions, including a pointer to more info.
- Variety of small readability edits.

PR #2250, opened 2017-03-16
Update simplesamlphp version

alexfinnarn:
I have reports from myself and at least one other person that the 1.14.x release seems to work fine on Pantheon. Seems reasonable to try and keep as up-to-date as possible.


alexfornuto:
Thanks @alexfinnarn! It looks like we have another PR that covers the version updates as well other additions, #2238. I'm going to close this as a duplicate, but thanks so much for the PR!

PR #2252, opened 2017-03-17
Add documentation on testing Terminus plugins with Bats and Circle CI.

greg-1-anderson:
Closes #n/a

## Effect
- Adds instructions on adding automated functional tests to the "Creating Plugins" page.

## Remaining Work
- [x] Code review by @TeslaDethray 
- [x] Review and wordsmith prose

### DIFF
--- a/source/_docs/terminus/plugins/create.md
+++ b/source/_docs/terminus/plugins/create.md
@@ -184,122 +184,141 @@ The first digit of a plugin version should always the Terminus MAJOR version the
 If your plugin has a minimum required version of Terminus, you can specify that in the `compatible-version` section of `composer.json`. You can use the [standard composer version constraints syntax](https://getcomposer.org/doc/articles/versions.md). If you do change `compatible-version`, please make sure that your constraint expression does not accidentally include the next major version of Terminus. In other words, `>=1.3 <2.0.0` is fine but `>=1.3` is not.
 
 ## Testing Plugins
-Adding automated testing is an important step to take for plugins that are being distributed. Automated tests give prospective new users the assurance that the plugin is still working, and also provide a basis for evaluating changes to the plugin.
+Adding automated testing is an important step to take for distributed plugins. Automated tests give prospective new users the assurance that the plugin works, and provides a basis for evaluating changes to the plugin.
 
 The instructions in this section demonstrate how to set up simple functional tests for Terminus plugins using Bats, the [Bash Automated Testing System](https://github.com/sstephenson/bats). Bats allows tests to be written with simple Bash statements.
 
-To get started, copy the `require-dev` and `scripts` sections from the composer.json file below into the composer.json of your Terminus plugin.
-```
-{
-    "name": "my-username/terminus-hello-world",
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
I have added tests to a number of Terminus plugins:

- https://github.com/pantheon-systems/terminus-composer-plugin
- https://github.com/pantheon-systems/terminus-drupal-console-plugin
- https://github.com/pantheon-systems/terminus-rsync-plugin
- https://github.com/pantheon-systems/terminus-secrets-plugin
- https://github.com/pantheon-systems/terminus-build-tools-plugin

I selected Bats as the test system for these plugins because it is simple, easy to add setup steps, and require a minimum of files to get running. I also considered using Behat, like Terminus itself, which I feel is a great tool for larger projects, but requires more files and more special knowledge (gherkin, context files) to get started. When folks are writing lightweight plugins, they are going to want their tests to be as easy to set up as possible. As can be seen in this document, these tests can be set up with only a few files, and only the test files themselves need to be customized.

source/_docs/terminus/plugins/create.md
```diff
@@ -184,122 +184,141 @@ The first digit of a plugin version should always the Terminus MAJOR version the
 If your plugin has a minimum required version of Terminus, you can specify that in the `compatible-version` section of `composer.json`. You can use the [standard composer version constraints syntax](https://getcomposer.org/doc/articles/versions.md). If you do change `compatible-version`, please make sure that your constraint expression does not accidentally include the next major version of Terminus. In other words, `>=1.3 <2.0.0` is fine but `>=1.3` is not.
 
 ## Testing Plugins
-Adding automated testing is an important step to take for plugins that are being distributed. Automated tests give prospective new users the assurance that the plugin is still working, and also provide a basis for evaluating changes to the plugin.
+Adding automated testing is an important step to take for distributed plugins. Automated tests give prospective new users the assurance that the plugin works, and provides a basis for evaluating changes to the plugin.
```
greg-1-anderson: You improved the English by changing this to `distributed plugins`, but now it might ambiguously imply that the plugins run on multiple machines due to overloading of the word `distributed`. Not sure if this would actually confuse anyone or not.

source/_docs/terminus/plugins/create.md
```diff
@@ -184,122 +184,141 @@ The first digit of a plugin version should always the Terminus MAJOR version the
...
-```
-#!/usr/bin/env bats
+        composer cs
 
-#
-# confirm-install.bats
-#
-# Ensure that Terminus and the Composer plugin have been installed correctly
-#
+4. Fix any PRS-2 compliance errors in your plugin:
```
greg-1-anderson: `cbf` fixes most problems, but there are some that it just tells you about and does not fix. For example, if your method is missing a `public`, `protected` or `private`, it doesn't know which was intended, so it tells you to add  the visibility modifier yourself. Maybe just changing `any` to `most` is sufficient.

alexfornuto:
@TeslaDethray Can you please take a look, make sure that these additions follow the preferred methodology?

PR #2253, opened 2017-03-17
fix up heading typography

kimby77:
Closes #AL-834

## Effect
Fixed up heading typography to include margins
Also aligned topmost heading with the "contents" heading on the right sidebar
Added a new class for the plugin directory to use the bootstrap margin on its titles, because the new margins were interferring.

## Remaining Work
- [ ] On "version updates" page - adjust the changelog states like fixed, added, deprecated, etc, as those are dynamic and I couldn't affect those on my local version


rachelwhitton:
+1 

The version updates can only be modified in the terminus project - we're doing an API request for the contents here: https://github.com/pantheon-systems/terminus/releases
So the source is outside the docs repository.

PR #2257, opened 2017-03-20
Adds information about Google default https indexing

alexfornuto:
## Effect
- Documents in several guides how Google's https-by-default indexing can link to your site using Pantheon's certificate.

## Remaining Work
- [x] CSE Signoff
- [ ] Copy Review

### DIFF
--- a/source/_docs/bots-and-indexing.md
+++ b/source/_docs/bots-and-indexing.md
@@ -55,7 +55,7 @@ You can index your site under your production domain once it's added to the Live
 Google's search engine will [index HTTPS pages by default](https://webmasters.googleblog.com/2015/12/indexing-https-pages-by-default.html). Because Pantheon serves all environments at `pantheonsite.io` over HTTPS by using our own certificate, this can lead to certificate mismatch security warnings in Google search results when you don't use your own HTTPS certificates. There are several possible solutions, including:
 
  - adding an `X-Robots-Tag: noindex` header to responses resulting from `https://` requests,
- - createing a sitemap that contains only `http` links,
+ - creating a sitemap that contains only `http` links,
  - enabling HTTPS for your domain, either [manually](/docs/enable-https) or with a service like [CloudFlare](/docs/guides/cloudflare-enable-https/).
 
 ### Sitemaps Produce a White Screen of Death (WSOD)
--- a/source/_docs/bots-and-indexing.md
+++ b/source/_docs/bots-and-indexing.md
@@ -50,13 +50,13 @@ You can index your site under your production domain once it's added to the Live
 
 ## Troubleshooting
 
-### Google Links to My Site over HTTPS with Pantheon's Certificate
+### Google Results for My Site Link to Pantheon's Platform Domain Using HTTPS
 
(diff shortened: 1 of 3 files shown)


source/_docs/bots-and-indexing.md
```diff
@@ -55,7 +55,7 @@ You can index your site under your production domain once it's added to the Live
 Google's search engine will [index HTTPS pages by default](https://webmasters.googleblog.com/2015/12/indexing-https-pages-by-default.html). Because Pantheon serves all environments at `pantheonsite.io` over HTTPS by using our own certificate, this can lead to certificate mismatch security warnings in Google search results when you don't use your own HTTPS certificates. There are several possible solutions, including:
 
  - adding an `X-Robots-Tag: noindex` header to responses resulting from `https://` requests,
- - createing a sitemap that contains only `http` links,
+ - creating a sitemap that contains only `http` links,
```
alexfornuto: That's what I get for rushing. bitmoji

source/_docs/bots-and-indexing.md
```diff
@@ -48,16 +48,24 @@ The `pantheonsite.io` domains are intended for development use and cannot be use
...
 
-### Troubleshooting
+## Troubleshooting
 
-#### Sitemaps Produce a White Screen of Death (WSOD)
+### Google Links to My Site over HTTPS with Pantheon's Certificate
+
+Google's search engine will [index HTTPS pages by default](https://webmasters.googleblog.com/2015/12/indexing-https-pages-by-default.html). Because Pantheon serves all environments at `pantheonsite.io` over HTTPS by using our own certificate, this can lead to certificate mismatch security warnings in Google search results when you don't use your own HTTPS certificates. There are several possible solutions, including:
+
+ - adding an `X-Robots-Tag: noindex` header to responses resulting from `https://` requests,
```
newtoid: If the links are already https from google, then they'll need a valid cert just to be able to serve the noindex response

alexfornuto: Meaning that this won't fix existing search indexes from Google, but it can prevent it from happening, and be a slow-to-fix solution long-term, right?

newtoid: yeppers, I would want them working asap and recommend free cloudflare, protocol redirects can be forced at the application level once the user gets past the bad cert

newtoid:
+1 :-)

source/_docs/bots-and-indexing.md
```diff
@@ -50,13 +50,13 @@ You can index your site under your production domain once it's added to the Live
 
 ## Troubleshooting
 
-### Google Links to My Site over HTTPS with Pantheon's Certificate
+### Google Results for My Site Link to Pantheon's Platform Domain Using HTTPS
```
alexfornuto: They link to the customer's Domain, but using the `*.pantheonsite.io` certificate.

allenfear:
Has the "noindex" approach been tested? I'm concerned that we're offering a solution that could result in having all pages dropped from the index, which is an even worse outcome than having the https ones there. I think we should have enabling https and using cloudflare be at the top of this list.

alexfornuto:
@allenfear my latest version moves the noindex to the bottom of that list. @newtoid could weigh in better than I on the results of that fix. I added it because CSE included it in their list of potential fixes.

newtoid:
@allenfear I think the noindex method will work, eventually, but it happens at the application level. not sure if google ignores the certificate and then goes on to honour the noindex

alexfornuto:
@allenfear @newtoid sounds to me like something we should include for now, and remove down the road if we can determine that Google isn't honoring it. Logic indicates that they should, but only time will tell.

PR #2278, opened 2017-03-27
improve FAQ layout - editing wp-config.php

rubelyn:
make FAQ items clickable so it can easily be shared in desk ticket


rachelwhitton:
+1 Eventually, we'll clean these up throughout based on a new style guide. Thanks for the PR @rubelyn!

PR #2289, opened 2017-03-29
Okta specific config

bentekwork:
The documentation is incorrect for configuring SSO for okta.  Just added okta specific config.


newtoid:
+1

PR #2296, opened 2017-03-30
FAQ section for renaming sites

rubelyn:
This is a common question. It will be nice to include this.


alexfornuto:
Sorry @rubelyn, This should have shown as "Merged" not "Closed". Next one will be right!

PR #2300, opened 2017-03-30
Recommend --temp-dir=~/tmp/ rather than ../tmp/

greg-1-anderson:
The relative path can cause problems when the target is somewhere other than `files/`. `~` is correctly evaluated as the home directory, e.g. the binding directory, and is therefore the better option.

## Effect
- Recommend an absolute path for the temporary directory
- see: [#power-users](https://pantheon-community.slack.com/archives/C2GJ3JG7Q/p1490889812894828)

## Remaining Work
- [ ] Review and merge

### DIFF
--- a/source/_docs/rsync-and-sftp.md
+++ b/source/_docs/rsync-and-sftp.md
@@ -52,7 +52,7 @@ export SITE=[YOUR SITE UUID]
 # Site UUID from dashboard URL: https://dashboard.pantheon.io/sites/<UUID>
 
 # To Upload/Import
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Actually, this suggestion doesn't work. Investigating further.

alexfornuto:
@greg-1-anderson that comment came in the nick of time!

greg-1-anderson:
OK, this is actually fine!

I found some odd results with `--temp-dir`, but it turns out the only issue was when copying files from Pantheon to the local system with `--temp-dir`. This document is already structured to recommend `--temp-dir` only for copies **to** Pantheon, so it's good to go as-is.

n.b. I updated the [Terminus rsync plugin](https://github.com/pantheon-systems/terminus-rsync-plugin/pull/3) to use this technique (copies only), and the tests are passing -- further proof this is valid.

alexfornuto:
👍  Thanks!

PR #2302, opened 2017-03-30
add note about user in charge permissions

### LINKED ISSUE #2301: User in Charge is unable to select site plan
Re: https://pantheon.io/docs/change-management/

## Issue Description
When an org is the holder/owner of a site, ONLY an org admin can change the service level. A user in charge is not allowed to launch a site (otherwise, for example, students at colleges with Pantheon Orgs could launch on their own site).

## Suggested Resolution
Update docs to reflect this.

alexfornuto:
Closes #2301 

## Effect
- Adds hover note explaining that users in charge cannot change site plan on sites owned by an org.

## Remaining Work
- [x] Review
- [ ] Merge

### DIFF
--- a/source/_docs/change-management.md
+++ b/source/_docs/change-management.md
@@ -123,7 +123,7 @@ In some Dashboards, you may notice the "User in Charge" label applied to a user.
       </tr>
       <tr>
         <td>Manage a site's plan</td>
-        <td>✓</td>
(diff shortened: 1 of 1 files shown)


alexfornuto:
@joshkoenig Do you feel like this accurately solves the issue?

PR #2312, opened 2017-03-30
Sendgrid module no longer requires Composer Manager.

greg-1-anderson:
Closes #n/a

## Effect
- Updates the documentation on the SendGrid module to remove warnings about the Composer Manager module, which is no longer used.
- Makes note that there is a beta release of the SMTP module for Drupal 8 now.

## Remaining Work
- [x] Review and merge


rachelwhitton:
Thanks!

PR #2330, opened 2017-04-06
Private Upstreams: Align Product Info with Admin Tool

rachelwhitton:
Re: AL-892

## Effect
- Update doc to request product information according to the new upstreams admin tool


kimby77:
Let's call out a few more important things directly on the page instead of in the tooltip:

- "URL of Logo (70x80px)"
- Repository Authentication - I think we should give them an example of using username:password and an example of using a token on the page.
- change "Connection Mode" to "Initial Connection Mode" for clarification
- Maybe "Visibility" should be called "Access"?  @ronan  what do you think of the terminology in that label?

ronan:
> Maybe "Visibility" should be called "Access"? @ronan what do you think of the terminology in that label?

I agree with "access" but am 4/10

source/_docs/custom-upstream.md
```diff
@@ -165,31 +165,26 @@ When testing is complete, merge any outstanding pull requests into the master br
...
-- Upstream Branch (usually master)
-- If the upstream repository is public or private
-- Description (< 200 characters, basic HTML allowed)
-- Logo (70x80)
-- Default connection mode (SFTP or Git)
-- Product author name and URL
-- Supporting information (< 200 characters, basic HTML allowed) i.e. drupal.org issue queue, support contacts, website
-- Links to evidence of testing on Pantheon. This should include the testing site, test files within the repository, and test reports.
+- **Upstream Name**
+- **Description**: (Optional) Less than 200 characters, basic HTML allowed
```
ronan: We don't want to encourage HTML anymore. I think we want to allow markdown but am not sure we actually do. Either way let's encourage plain text for new upstreams until we figure it out.

ronan:
1 minor note about allowing html. Otherwise LGTM

source/_docs/custom-upstream.md
kimby77: 👍

PR #2331, opened 2017-04-06
fix inconsistent log downloads

newtoid:
I tested this a few times, and the slash wildcard gave consistently better results. I was looking for a php-error.log file which I knew existed on the application server but was not being downloaded.

## Effect
Changes rsync path to include wildcard so all files in /logs/ are downloaded

## Remaining Work
- [x] peer test to replicate


alexfornuto:
Thanks @newtoid 
bitmoji

PR #2333, opened 2017-04-07
Add info about the CA bundle available

bwood:
https://pantheon-community.slack.com/archives/C2GJ3JG7Q/p1491582190890165

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed


joshkoenig:
Cool! If we decide to endorse this, I'd prefer if we hosted the CA bundle in a more obvious/public way. We'll follow up on this shortly.

joshkoenig:
Discussed internally and we don't want this particular suggestion in docs at the moment. It will work, but we don't want to support it long-term.

Next steps to meet the need:

[ ] We create a high-grade standalone CA cert (vs our legacy SHA-1 bundled)
[ ] Make that cert available directly via the docs site and embed a link in the dashboard
[ ] Update docs on how to use this cert

PR #2338, opened 2017-04-10
Implement Video Feedback

rachelwhitton:
## Effect
- Increase frame width for existing inline doc videos
- Add same frame to existing videos on topic pages

@AndrewKWilson can you review?


AndrewKWilson:
@rachelwhitton we also need to add the frame to video-only pages (e.g., https://videocsst-static-docs.pantheonsite.io/docs/videos/migrate-wordpress/)

rachelwhitton:
@AndrewKWilson great catch! Updated

PR #2342, opened 2017-04-11
WIP - Drupal 8 Console & Composer

### LINKED ISSUE #925: Create Console on Pantheon article
## Description

The Drupal Console is a suite of tools that you run on a command line interface (CLI)
to generate boilerplate code and interact with a Drupal 8 installation.
Build from https://pantheon.io/introduction-drush-and-drupal-console 
## Outline
1. Explain Console and how it works on Pantheon
2. Installation
3. Commands and their usage
## Expected Audience

D8 users
## Reference

 https://www.drupal.org/project/console
## Path

alexfornuto:
Closes #925 

## Effect
- New guide on using Drupal Console
- Expanded information on Composer and Drupal 8 on Pantheon

## Remaining Work
- [ ] Finish writing.
- [ ] SME review
- [ ] Docs Review

### DIFF
--- a/source/_docs/composer-drupal-8.md
+++ b/source/_docs/composer-drupal-8.md
@@ -17,7 +17,13 @@ If you're already familiar with Composer on Pantheon, you can move on to [Instal
 
 ### Workflow and Nested Docroot
 
-Before using Composer to manage your site development you should choose a workflow. By default, Pantheon controls and distributes core updates via git commits, which conflicts with updates processed by Composer. We suggest developers using Composer install their sites with the [Example Drops 8 Composer](https://github.com/pantheon-systems/example-drops-8-composer) repository, which moves Drupal to a "nested docroot". The repository README file covers the two common methods of deployment.
+Before using Composer to manage your site development you should choose a workflow. By default, Pantheon controls and distributes core updates via git commits, which can cause conflicts with files updated by Composer. Two common workflows around this issue are:
+
+ - Use Pantheon as a build environment, and managing development changes with an externally (or locally) hosted repository.
+
+ - Set a custom upstream (available to organizations), and manage core updates manually.
+
+We suggest developers using Composer install their sites with the [Example Drops 8 Composer](https://github.com/pantheon-systems/example-drops-8-composer) repository, which moves Drupal to a "nested docroot". The repository README file covers the two common methods of deployment.
 
 <div class="alert alert-info">
 <h4 class="info">Note</h4><p markdown="1">
--- a/source/_docs/composer-drupal-8.md
+++ b/source/_docs/composer-drupal-8.md
@@ -7,7 +7,7 @@ contributors:
   - stevector
 ---
 
-[Composer](https://getcomposer.org/) is a dependency management tool that allows PHP code to be more easily shared across projects. As Drupal 8 was written, many of its subsystems written specifically for Drupal were replaced by more widely used packages, using Composer. For example, Drupal's notoriously unwieldy function for making external requests, `drupal_http_request()`, [was replaced](https://www.drupal.org/node/1862446) by [Guzzle](http://guzzlephp.org/). Guzzle is included in Drupal 8 via Composer.
+[Composer](https://getcomposer.org/) is a dependency management tool that allows PHP code to be more easily shared across projects. As Drupal 8 was written, many of its subsystems written specifically for Drupal were replaced by more widely used packages, using Composer. For example, Drupal's function for making external requests, `drupal_http_request()`, [was replaced](https://www.drupal.org/node/1862446) by [Guzzle](http://guzzlephp.org/). Guzzle is included in Drupal 8 via Composer.
 
 ## Caveats
 
--- a/source/_docs/composer-drupal-8.md
+++ b/source/_docs/composer-drupal-8.md
@@ -31,7 +31,7 @@ The methods described by the Example Drops 8 Composer repository replace the sta
 </p>
 </div>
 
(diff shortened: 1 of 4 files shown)


source/_docs/composer-drupal-8.md
```diff
@@ -9,6 +9,16 @@ contributors:
 
 [Composer](https://getcomposer.org/) is a dependency management tool that allows PHP code to be more easily shared across projects. As Drupal 8 was written, many of its subsystems written specifically for Drupal were replaced by more widely used packages. For instance, Drupal's notoriously unwieldy function for making external requests, `drupal_http_request()`, [was replaced](https://www.drupal.org/node/1862446) by [Guzzle](http://guzzlephp.org/). Guzzle is included in Drupal 8 via Composer.
 
+Before using Composer to manage your site development you should choose a workflow. By default, Pantheon controls and distributes core updates via git commits, which conflicts with updates processed by Composer. We suggest developers using Composer install their sites with the [Example Drops 8 Composer](https://github.com/pantheon-systems/example-drops-8-composer), which moves Drupal to a "nested docroot". The repository README file covers the two common methods of deployment.
```
alexfornuto: adding the word "repository" after the link.

source/_docs/composer-drupal-8.md
```diff
@@ -9,6 +9,16 @@ contributors:
 
 [Composer](https://getcomposer.org/) is a dependency management tool that allows PHP code to be more easily shared across projects. As Drupal 8 was written, many of its subsystems written specifically for Drupal were replaced by more widely used packages. For instance, Drupal's notoriously unwieldy function for making external requests, `drupal_http_request()`, [was replaced](https://www.drupal.org/node/1862446) by [Guzzle](http://guzzlephp.org/). Guzzle is included in Drupal 8 via Composer.
```
greg-1-anderson: Let's not use quite so colorful adjectives to describe the poor drupal_http_request() function. Perhaps just noting that it was replaced, without being descriptive about why, would be better.

alexfornuto: image

I'll change it.

source/_docs/composer-drupal-8.md
greg-1-anderson: Are we going to describe multiple workflows for the user to choose between? Certainly there is a choice between "Pantheon upstream" vs. "composer update"; perhaps in this document, the choice should be limited to those two options. The GitHub workflow can be introduced later.

`By default, Pantheon controls and distributes core updates via git commits, which conflicts with updates processed by Composer.` is a little unwieldy. Maybe something like "which can cause conflicts with files updated by Composer" would be clearer.

 `The repository README file covers the two common methods` - at this point it's not clear which two methods you are talking about. c.f. comment above.

source/_docs/composer-drupal-8.md
```diff
@@ -9,6 +9,16 @@ contributors:
...
 
+Before using Composer to manage your site development you should choose a workflow. By default, Pantheon controls and distributes core updates via git commits, which conflicts with updates processed by Composer. We suggest developers using Composer install their sites with the [Example Drops 8 Composer](https://github.com/pantheon-systems/example-drops-8-composer), which moves Drupal to a "nested docroot". The repository README file covers the two common methods of deployment.
+
+<div class="alert alert-info">
+<h4 class="info">Note</h4><p markdown="1">
+The methods described by the Example Drops 8 Composer repository replace the standard Drupal 8 upstream. By choosing this method, you become responsible for ensuring that your core remains up to date, per our [scope of support](/docs/getting-support/#scope-of-support)
```
greg-1-anderson: Previously, the user is responsible for clicking the dashboard button to accept updates provided by Pantheon. With this method, the user is responsible for running `composer update` to accept updates resolved by Composer. Analogous to the comment in the "scope-of-support", if the user selects a non-Pantheon site template (e.g. drupal-composer/drupal-project), then their site will not work correctly. However, if the user selects a Pantheon site template, then Composer will provide the core files. The dashboard will still offer to provide updates. If the user applies them, then bad things will happen.

alexfornuto: So... I _think_ I understand what you're saying, but not at a deep enough level that I could properly articulate it. Would you mind taking a stab at that in the doc?

source/_docs/composer-drupal-8.md
```diff
@@ -9,6 +9,16 @@ contributors:
...
+</p>
+</div>
+
+If you'd instead prefer to the default Drupal 8 installation provided by Pantheon, use a tool like the [Composer Merge Plugin](https://github.com/wikimedia/composer-merge-plugin), which parses multiple `composer.json` files into a single version when running updates. This method is the most likely to see merge conflicts, as core updates will also update Drupal's `composer.json`.
```
greg-1-anderson: Drupal provides the Composer Merge Plugin, and uses it automatically; this isn't something that the user has to choose to do. Also, the composer merge plugin isn't really the thing that is causing conflicts in this workflow. Rather, the fact that both  Composer and the Pantheon upstream may change the same files in the vendor directory, and both will update the composer.lock file.

alexfornuto: > Drupal provides the Composer Merge Plugin, and uses it automatically; this isn't something that the user has to choose to do.

Does that mean the plugin is already included in core's `composer.json` file? If not, how/when is it introduced?

greg-1-anderson: Drupal core does not update `composr.json` very often; rather, the conflicts will come in vendor and composer.lock most of the time. These are easier to resolve (just run `composer update` again), but unfortunately this may only be done from the commandline.

alexfornuto:
Assigned to @greg-1-anderson, who's going to add some more information for us to process about the issues stemming from using Composer with Drupal 8 on Pantheon.

rachelwhitton:
Closing, to be tracked internally at DSI-74

PR #2367, opened 2017-04-17
Clarifying Early Access

joshkoenig:
Just clarifying the early access offer: we should only promise Pro+ plans and Live envs at this time.


rachelwhitton:
+1 Thanks @joshkoenig

joshkoenig:
+1 looks awesome.

PR #2369, opened 2017-04-17
Updating the written out number to be consistent

davidneedham:
## Effect
- Changes "1" to "one" so that numbering is consistent on this page.


alexfornuto:
👍  Thanks!1!

PR #2378, opened 2017-04-18
add note on tracking blockers

### LINKED ISSUE #2376: Document known issues with Chat support and ad blockers
Document that some ad-blockers can prevent Chat support from working.  This generally manifests as the 'Launch Chat' button not working.  Specifically, Sybu JS blocker to my Chrome Browser prevents Chat from working (probably others as well).

Would be nice to also have tips for determining if this is the issue.

(Internal reference BUGS-1194)

Some other intercom docs: https://docs.intercom.com/faqs-and-troubleshooting/your-users-and-leads-data-in-intercom/why-am-i-missing-users-that-are-showing-up-on-my-own-system

alexfornuto:
Closes #2376 

## Effect
- Adds note that ad/tracking blockers can interfere with chat support.
- Links to Intercom doc with more details.

## Remaining Work
- [x] Approval from @nstielau that this resolves his concerns.
- [ ] Copy Edit

### DIFF
--- a/source/_docs/getting-support.md
+++ b/source/_docs/getting-support.md
@@ -98,6 +98,11 @@ Start a chat with our Support Team to request assistance on issues within our [s
  - Error messages received, if applicable.
  - Links to screenshots or screencasts of the behavior, if necessary.
 
+<div class="alert alert-info" role="alert">
+<h4 class="info">Warning</h4>
+<p markdown="1">Some ad or tracking blockers can interfere with the **Launch Chat** button. If you don't see it, try whitelisting `pantheon.io` or disabling the extension. Intercom has posted more details on [why this happens](https://docs.intercom.com/faqs-and-troubleshooting/your-users-and-leads-data-in-intercom/why-am-i-missing-users-that-are-showing-up-on-my-own-system) on their site.</p>
+</div>
+
 ## Ticket Support
 The ticket support feature is available to certain plans and account types. For details, refer to the [support feature table above](#support-features-and-response-times).
 
--- a/source/_docs/getting-support.md
+++ b/source/_docs/getting-support.md
@@ -99,8 +99,8 @@ Start a chat with our Support Team to request assistance on issues within our [s
  - Links to screenshots or screencasts of the behavior, if necessary.
 
 <div class="alert alert-info" role="alert">
-<h4 class="info">Warning</h4>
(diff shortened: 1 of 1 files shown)


source/_docs/getting-support.md
```diff
@@ -98,6 +98,11 @@ Start a chat with our Support Team to request assistance on issues within our [s
  - Error messages received, if applicable.
  - Links to screenshots or screencasts of the behavior, if necessary.
 
+<div class="alert alert-info" role="alert">
+<h4 class="info">Warning</h4>
```
rachelwhitton: Let's use "Warning" headers with `alert-danger` only, not `alert-info`

source/_docs/getting-support.md
```diff
@@ -98,6 +98,11 @@ Start a chat with our Support Team to request assistance on issues within our [s
  - Error messages received, if applicable.
  - Links to screenshots or screencasts of the behavior, if necessary.
 
+<div class="alert alert-info" role="alert">
+<h4 class="info">Warning</h4>
+<p markdown="1">Some ad or tracking blockers can interfere with the **Launch Chat** button. If you don't see it, try whitelisting `pantheon.io` or disabling the extension. Intercom has posted more details on [why this happens](https://docs.intercom.com/faqs-and-troubleshooting/your-users-and-leads-data-in-intercom/why-am-i-missing-users-that-are-showing-up-on-my-own-system) on their site.</p>
```
rachelwhitton: Can we include the word browser in here? Something like this: 

>Browser extensions that block tracking or ads can interfere with the.. 


Also, suggest clarifying the second sentence: 
> If you don't see the **Launch Chat** button, try whitelisting `pantheon.io` or disabling the extension.

source/_docs/getting-support.md
alexfornuto: Oversight in my part, bad copy pasta.

source/_docs/getting-support.md
alexfornuto: 👍

alexfornuto: > Also, suggest clarifying the second sentence:

>>If you don't see the Launch Chat button, try whitelisting pantheon.io or disabling the extension.

In this case "it" refers to the Launch Chat button at the end of the second sentence. Repeating it would read:

> Some in-browser ad or tracking blockers can interfere with the **Launch Chat** button. If you don't the **Launch Chat** button, try whitelisting `pantheon.io`...

Might that be redundant?

rachelwhitton: Ah, yeah. Makes sense - thanks!

PR #2390, opened 2017-04-21
fix formatting

NickVolynkin:
## Effect

- fixed formatting:
    - bulleted list entries indented with 4 spaces (were indented with 1 \
  space, thus not recoginzed as second level list items).
    - fenced code block indented with 4 spaces, 
  to make it belong to the list item.
    - added bash marker for syntax highlighting


rachelwhitton:
Awesome, thanks @NickVolynkin!

PR #2392, opened 2017-04-22
Updating the terminus drush example

davidneedham:
Updating the terminus drush example to work with Terminus 1.x.

## Effect
-Updates the second drush example to include syntax for Terminus 1.x.


alexfornuto:
bitmoji

PR #2403, opened 2017-04-26
Learn Private Upstreams

rachelwhitton:
## Effect
- Delete the following pages: 
  - [/docs/managing-upstreams](https://pantheon.io/docs/managing-upstreams/)
  - [/docs/running-custom-upstream](https://pantheon.io/docs/running-custom-upstream/)
  - [/docs/custom-distribution-updates](https://pantheon.io/docs/custom-distribution-updates/)
- Add/rewrite the following pages:
  - [/docs/custom-upstream](https://upstreams-static-docs.pantheonsite.io/docs/custom-upstream/)
  - [/docs/create-custom-upstream](https://upstreams-static-docs.pantheonsite.io/docs/create-custom-upstream/)
  - [/docs/maintain-private-upstreams](https://upstreams-static-docs.pantheonsite.io/docs/maintain-custom-upstream/)
  - [/docs/upstream-updates](https://pantheon.io/docs/upstream-updates)

## Remaining Work
Redirects required for: 
https://pantheon.io/docs/managing-upstreams/ -> /docs/maintain-custom-upstream
https://pantheon.io/docs/running-custom-upstream/  -> /docs/maintain-custom-upstream
https://pantheon.io/docs/custom-distribution-updates/ -> /docs/upstream-updates

cc @ajbarry @kimby77


ronan:
I'm not sure if this goes against the doc style but I'd like to see numbers on larger titles sections that are connected to each other. Once you've read through a big paragraph you kind of forget you're in a list. The two examples are the subsections of Benefits on the intro and the steps on create.

I would remove 'level' from Site Level Repository in the glossary unless that phrase exists outside of this doc. Also we later talk about 'Site Codebase' where I think we could just talk about the 'Site Repository' since the concepts are very close and we've already defined one.

None of these is major

kimby77:
image

"Learn how to use custom upstreams to free up developer time."?  Because learning terminology is kind of an aside & falls within that goal.


image
I am wondering if you have any design pattern in Docs that could separate the last paragraph visually from the preceding two?  That might help it standout more as not just being part of the text, but that a user can learn about a real world use case by reading the case study.  And same comment for the equivalent case study links in the next two sections.

image
"over time"?

image
This is the only place (out of three places) that we call out the author name - seems kind of out of place for documentation?  What about just "Hear from Kalamuna" so it matches how we refer to the other two case studies.

image
The user isn't agreeing on terms, so maybe something more like "Here are definitions for commonly used terms."

image
Would it make sense to add the "copy" buttons here on the right side of the codeblock?

image
Why is the styling different for the top two code blocks vs. the third?

image
Same here.

source/_docs/maintain-custom-upstream.md
```diff
@@ -0,0 +1,263 @@
...
+
+3. Checkout a new branch:
+
+    ```command
+    git checkout -b core-update
+    ```
+
+    It's important to use feature branches when applying and testing updates. Updates applied on the master branch and pushed to the remote repository on GitHub or Bitbucket become available to individual sites downstream. Using a feature branch gives us a chance to reveal issues before distributing updates.
+
+4. Pull down Pantheon's core updates from the appropriate upstream:
```
alexfornuto: The commands in the tabs below rebase against master. Based on the preceding paragraph, wouldn't we want to rebase `pantheon-<CMS>` to the `core-update` branch?

rachelwhitton: The rebase to `pantheon-<CMS>/master` is run against the current branch, which is `core-update`

PR #2409, opened 2017-04-28
Free HTTPS: why action required after DNS update

ari-gold:


rachelwhitton:
+1 Thanks @ari-gold!

source/_docs/free-https.md
```diff
@@ -171,6 +171,13 @@ No, please take a moment to fill out the [HTTPS survey](https://www.getfeedback.
 ### Is the CDN configurable? Do I get access to hit rates or other statistics?
 No, we pre-configured the CDN so you don’t have to hassle with configuration. We’ve optimized configuration for Drupal and WordPress sites. Hit rates or other statistics are not currently available.
 
+### Why does the Domains and HTTPS tool indicate action required even though I've updated DNS records?
+The action required message may be due to one of the following:
+
+  - **DNS changes can take some time to take effect:** Check the current state of DNS propagation from different parts of the world using this [free web tool](https://www.whatsmydns.net/).
```
alexfornuto: You *could* change the header to:  **DNS changes can take some time to spread:**, which is a little shorter, and slightly more accurate. But I don't feel strongly about it.

source/_docs/free-https.md
```diff
@@ -171,6 +171,13 @@ No, please take a moment to fill out the [HTTPS survey](https://www.getfeedback.
...
+  -  **AAAA records not detected:**: Ensure you've added both AAAA records for bare domains (e.g. `example.com`) to route IPv6 traffic to your site. There are two AAAA records for improved uptime and reliability.
```
alexfornuto: > nsure you've added **both** AAAA records for bare domains

I suggest "both AAAA records for _the_ bare domain." As currently written, it sounds like **both** is referring to a second, unmentioned item.

source/_docs/free-https.md
rachelwhitton: I prefer the existing - I think it will help those affected find what they need faster.

source/_docs/free-https.md
```diff
@@ -171,6 +171,13 @@ No, please take a moment to fill out the [HTTPS survey](https://www.getfeedback.
 ### Is the CDN configurable? Do I get access to hit rates or other statistics?
 No, we pre-configured the CDN so you don’t have to hassle with configuration. We’ve optimized configuration for Drupal and WordPress sites. Hit rates or other statistics are not currently available.
 
+### Why does the Domains and HTTPS tool indicate action required even though I've updated DNS records?
+The action required message may be due to one of the following:
```
alexfornuto: Could this be made more succinct as: 
>The **Action Required** message may be caused by:
?

alexfornuto: ...nevermind, that would require changing the tense of the subheadings.

source/_docs/free-https.md
alexfornuto: bitmoji

PR #2413, opened 2017-04-28
D8 HASH_SALT section is not clear

peter-pantheon:
D8 HASH_SALT section not clear. It appears to be contradictory or is discussing two different aspects of HASH_SALT, not sure

Closes #

## Effect
- I flagged the areas the seem contradictory. 

## Remaining Work
- Actual correct edits still needed


alexfornuto:
@peter-pantheon so which one is correct? Do you need the `HASH_SALT` for a local site or not?

alexfornuto:
@peter-pantheon I've updated the phrasing to show the difference. You need *some* hash_salt for a local site, but it need not be the same one used on Pantheon.

PR #2414, opened 2017-04-28
Env Ind Module is released for D8

peter-pantheon:
D8 Environment Indicator module version 8.x-3.0 is out.

Closes #

## Effect
- Updated D8 module version info


alexfornuto:
Awesome, thanks Peter!

PR #2434, opened 2017-05-05
Add Drupal 8 settings.php configuration snippet

sbbutkcin:
The Drupal configuration appears to be just for D7 sites. I had to tinker around/read the D8 module's README.md in order to get the module to recognize the library and become enabled. I am not sure this is the best/recommended method but it appears to work, at least for me.

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed

### DIFF
--- a/source/_docs/shibboleth-sso.md
+++ b/source/_docs/shibboleth-sso.md
@@ -94,13 +94,13 @@ Add the following lines to `settings.php` so that the Drupal module can locate S
 For Drupal 7 sites:
 ```php
 # Provide universal absolute path to the installation.
-$conf['simplesamlphp_auth_installdir'] = $_SERVER['HOME'] .'/code/private/simplesamlphp-1.14.x';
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
@rachelwhitton @alexfornuto Just a note so that I don't forget. It should be possible to use `$_SERVER['HOME']` in both the Drupal 7 and Drupal 8 examples. I don't think it is necessary to decode the pressflow setting environment variable a second time; should work just fine to remove this from both examples.

greg-1-anderson:
Sorry for not remembering this previously, @rachel, but `$_ENV` is preferred to `$_SERVER` on Pantheon. There are a lot of examples that use `$_SERVER`, but we prefer to use `$_ENV` consistently.

So, `$_SERVER['HOME']` should be changed to `$_ENV['HOME']`

rachelwhitton:
Thanks for the PR @sbbutkcin, much appreciated! We made a small change to the code block fyi

PR #2441, opened 2017-05-08
Correct terminus syntax for exporting D8 config files

calmher:
Closes #

## Effect
- corrects syntax found under 'Export Configuration' in https://pantheon.io/docs/solr-drupal-8/#3.-configure-solr


alexfornuto:
👍

PR #2446, opened 2017-05-09
Adds Ticket Tips

### LINKED ISSUE #2039: Improve Getting Support Doc
## Title
How to best work with support

## Description
I would so love a guide on how to best work with support. Somethings like "help us to help you". 

## Outline
Things like:
- Please keep 1 ticket pr issue - no more, no less
- More issues pr ticket confuses the subject and makes it harder for us to assist and provide proper advice. CSEs also get confused!
- Many tickets / chats on the same issue ends with CSEs spending time trying to consolidate the scattered information between us, wasting time that could have been spent solving the issue.
More tickets != more/quicker support, quite the contrary!
- If opening a chat on an issue that already has a ticket, please let us know the ticket is there. Chat is currently a new product and we might miss the connection.
- Please add exact steps to reproduce the issue. We can only debug what we can see. Help us see what you see.
- Adding information on which environment, time, error message, screenshot, urls, etc is super helpful. Context is also necessary!
- If the site has an unusual configuration that means we cannot access login / edit pages etc as normal, add this information as well.
- This speeds things up: Stay responsive, help us by answering the questions we ask.
- Let us know if something doesn't make sense or we have misunderstood the issue. CSEs are highly aware of the importance to take note of every detail - but we're human too, it does happen we miss stuff.
- In general: Help us to help you!

## Expected Audience
Customers

## Path

alexfornuto:
Closes #2039 

## Effect
- Adds suggested tips from #2039 to improve support interactions.

## Remaining Work
- [ ] @pantheon-systems/cse buy-in
- [x] Copy review.

### DIFF
--- a/source/_docs/getting-support.md
+++ b/source/_docs/getting-support.md
@@ -168,7 +168,7 @@ Keep the following tips in mind when opening a new support ticket or chat:
 
  - **Custom Access**: If your site uses nonstandard admin URLs, custom logins, etc, please let us know.
 
- - **Clarification**: We're all of us only human. We always try our best to understand the situation, but we may not always get it right the first time. At the same time, our explanations or solutions might not always make sense. Don't be afraid to correct us, or to ask for clarification if we don't seem to make sense.
+ - **Clarification**: We're all only human. We always try our best to understand the situation, but we may not always get it right the first time. At the same time, our explanations or solutions might not always make sense to you. Don't be afraid to correct us, or to ask for clarification if we don't seem to understand the issue.
 
 ## FAQs
 #### Is there a support number we can call? If so, is this service available for every plan?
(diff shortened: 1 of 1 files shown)


rachelwhitton:
+1

ttrowell:
From the Peanut gallery:

I would edit this paragraph a little:

original

> We're all of us only human. We always try our best to understand the situation, but we may not always get it right the first time. At the same time, our explanations or solutions might not always make sense. Don't be afraid to correct us, or to ask for clarification if we don't seem to make sense.

revised

> We're all human. We try our best to understand the situation, but we may not always get it right the first time. At the same time, our explanations or solutions might not always make sense to you. Don't be afraid to correct us, or to ask for clarification if we don't seem to understand the issue.

LoVE the rest!  +1

PR #2453, opened 2017-05-11
Add database import method according to size

thomas-thackery:
Borrowing language from the database import information at https://pantheon.io/docs/migrate-wordpress-site-networks/#database, this more general manual migration documentation would benefit by allowing users to see which import method is required according to the size of the database.


rachelwhitton:
+1

PR #2458, opened 2017-05-11
Update sendgrid.md

### LINKED ISSUE #2457: Sendgrid SMTP port update
Re: https://pantheon.io/docs/guides/sendgrid/

## Issue Description
We want to ensure all users utilizing sendgrid have the most up to date port information compatible with the platform

## Suggested Resolution
Add new port information to sendgrid documentation

lauren-kelly:
Closes #2457 

## Effect
- update port
- update acceptable encryption options
-

## Remaining Work
- [ ] review for ambiguity on encryption options

### DIFF
--- a/source/_docs/guides/sendgrid.md
+++ b/source/_docs/guides/sendgrid.md
@@ -79,10 +79,12 @@ A stable release for Drupal 8 is not yet available for the [SMTP Authentication 
 3. From within Install Options, select **On**.
 4. Use the following within SMTP Server Settings:
 
- **SMTP server**: smtp.sendgrid.net <br>
- **SMTP port**: 2525<br>
- **Use encrypted protocol**: We strongly reccomend selecting **TLS**<br>
-
+   **SMTP server**: smtp.sendgrid.net 
+ 
+   **SMTP port**: 2525
+ 
+   **Use encrypted protocol**: We strongly reccomend selecting **TLS**
+ 
 5. Provide your site-specific SendGrid credentials and click **Save configuration**.  
 
 Your Drupal application on Pantheon is now set up to send email through SendGrid. Provide an address within the Send Test E-mail configuration field and click **Save configuration** to test.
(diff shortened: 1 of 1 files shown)


lauren-kelly:
@alexfornuto can you take a look and make sure that looks good?

source/_docs/guides/sendgrid.md
```diff
@@ -80,8 +80,8 @@ A stable release for Drupal 8 is not yet available for the [SMTP Authentication
 4. Use the following within SMTP Server Settings:
 
  **SMTP server**: smtp.sendgrid.net <br>
- **SMTP port**: 465<br>
- **Use encrypted protocol**: Use SSL<br>
+ **SMTP port**: 2525<br>
+ **Use encrypted protocol**: Either No Encryption or TLS<br>
```
garydylina: Is there a good reason to suggest No Encryption?

taurusismysign: We (re)documented here what SendGrid offers on 2525 and leave for the user to decide on protocol.

alexfornuto: We could include a parenthetical **(not recommended)**

garydylina: I would feel better about not recommending the option we don't recommend unless we do recommend it and then our direction is clear. ;)

lauren-kelly: @garydylina my thought on it was to head off questions of "can I just leave it as no encryption?" But seeing as we didn't see that with it saying Use SSL before, Use TLS seems fine - and we can update it later if there are problems/questions coming in

alexfornuto: > I would feel better about not recommending the option we don't recommend unless we do recommend it and then our direction is clear. ;)

Stating the truth is not the same as recommending it. Things can go into documentation that we don't recommend, and including them does not automatically mean they are recommended.

lauren-kelly:
@alexfornuto @garydylina  @taurusismysign Are we all set here?

garydylina:
👍

PR #2471, opened 2017-05-14
Update site-owner-faq.md about Annual invoicing for sites

twooten:
This question comes up periodically.


lauren-kelly:
@twooten unfortunately it isn't that easy/straightforward :/
@ccorpus maybe we should meet and see if we can come up with some language that would be helpful to customers and reduce the number of request?

lauren-kelly:
@alexfornuto Take a look and see what you think?

alexfornuto:
LGTM, although I'd like to restructure the entire page so that these questions are in H3s with anchor links instead of H4s. What do you think, @rachelwhitton?

rachelwhitton:
+1

PR #2480, opened 2017-05-16
Adding enterprise

joshkoenig:
Enterprise customers also benefit. Adding that + links.


rachelwhitton:
Thanks @joshkoenig! 
cc @ronan @kimby77

PR #2482, opened 2017-05-16
Clarify next steps after adding "web_docroot: true"

bdimaggio:
When I tried following these with an existing Drupal 8 site, I got through step 5 and immediately started testing my site on the Pantheon dev environment. I could've used a step 6 that told me I would have to follow one of the headers that followed. I've added a suggestion for that step here. Hope it helps!

Closes #

## Effect
- Adds a "step 6" to make clear what one should do after "Add, commit, and push the pantheon.yml file with Git."

## Remaining Work
- [ ] If needed


ilovelamp:
I think a step 6 referring to next sections makes sense.

alexfornuto:
Thanks! 👍

PR #2484, opened 2017-05-16
New Diagram: Custom Upstream Intro

rachelwhitton:
Re: AL-1054

## Effect
- Add svg source files 
- CSS and inline HTML for diagram
 - Inserts an arrow icon after the first two containers
- Up asset ver to force browser reload

@ronan @kimby77


ronan:
This lgtm

kimby77:
You did that fast!  I have some comments:

- I meant to save out the arrow as a svg too - I did that and attached it to the task.  The font awesome icon is squashed & the down direction is kinda lost, so would you be able to switch to the svg?
- I tried out some adjustments in the browser.  Here's what I did (but use whatever makes sense with your patterns that'll get this close):

image

- moved the left column up so it's more in line with the heading (I just took off the `margin-top: 30px` and that did it)
- list item should have less of a line height than the entire list, so that when text wraps to more than 1 line it's easy to tell where that line stops and the others begin.  I used `line-height: 1.35` on the li.
- added more padding around the entire container to give them more breathing room. I used `padding: 25px` on `upstream-overview-outter`

kimby77:
Do you think the updates image looks better centered?

image


image

PR #2487, opened 2017-05-17
Adding info to bring environments into sync

### LINKED ISSUE #2486: How do users fix when Dev is behind Test and Live
Re: docs/undo-commits
## Issue Description
If you reset dev, it will be behind Test and Live and you cannot bring them to the same commit as Dev.

## Suggested Resolution
make a dummy commit

lauren-kelly:
also slight edit on wording about testing...

Closes #2486

## Effect
- Adds info on dummy commits
- update wording about testing commits before deploying...


alexfornuto:
👍

PR #2494, opened 2017-05-19
Terminus sql-query example

ilovelamp:
RE: https://pantheon.io/docs/drush/

Provides an example of how to run a single SQL query via Terminus -> drush sql-query.


alexfornuto:
Looks good, thanks! 👍

PR #2502, opened 2017-05-23
Typo fix in drupal8-commandline.md

stevector:
I was just running through this guide on a Jump Start and noticed that the last command references the wrong environment.


rachelwhitton:
Ah, thanks! +1

PR #2408, opened 2017-04-27
Update upstream-updates.md

ttrowell:
Please format and wordsmith this mess.  :P
## Effect
PR includes the following changes:
-
-Adding steps to manually align a codebase with the Pantheon upstream.

## Remaining Work
- [x] Verify the paths not to be overwritten for each upstream

PR #2531, opened 2017-06-01
Manually Overwrite Core: Last Resort

rachelwhitton:
Replace #2408

## Blocked By
- [ ] Consult with product and/or @pantheon-systems/cse  required, as this method doesn't seem entirely safe.

### DIFF
--- a/source/_docs/upstream-updates.md
+++ b/source/_docs/upstream-updates.md
@@ -168,7 +168,11 @@ There are multiple reasons that 503 errors might occur when updating:
 
 
 ### Process to Resolve Conflicts Locally responds with 'Already-up-to-date'
-If you know your site's core codebase does not match Pantheon's upstream, and attempts to resolve it locally do not work because your site repository is seen as up-to-date, you may need to manually overwrite core files. We only recommend this method as an absolute last resort.
+If you know your site's core codebase does not match Pantheon's upstream, and attempts to resolve it locally do not work because your site repository is seen as up-to-date, you may need to manually overwrite core files. We only recommend this method as an absolute last resort. A common reason to end up in this state is that core updates have been merged, but a revert commit has been applied on top of the core updates, causing Git to believe all updates have been applied when the code does not actually match. This happens because the dashboard is only checking for the inclusions of commits in the history that match the upstream.
+
+<div class="alert alert-danger" role="alert">
+    This process is destructive. Any core patches applied will be removed when overwriting core with our upstreams. This is normally not an issue, but if the site includes core patches, they will need to be re-applied.
+</div>
 
 1. Download an archive of the Pantheon upstream:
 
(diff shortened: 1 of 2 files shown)


source/_docs/upstream-updates.md
```diff
@@ -165,3 +165,35 @@ There are multiple reasons that 503 errors might occur when updating:
 - PHP segfault: These are tricky to troubleshoot because very little debugging information is present. A temporary fix is available. Contact Pantheon Customer Support if you think you have been affected.
 
 - Timeouts are another cause of 503 errors, though they are much less likely to occur if you are using the Pantheon domains. If the operation takes more than 60 seconds, you might see a timeout occur.
+
+### Process to Resolve Conflicts Locally responds with 'Already-up-to-date'
+If you know your site's core codebase does not match Pantheon's upstream, and attempts to resolve it locally do not work because your site repository is seen as up-to-date, you may need to manually overwrite core files. We only recommend this method as an absolute last resort.
+
```
dustinleblanc: We might want to add some common reasons this happens. The most common that I see is that a customer has pulled updates, but subsequently, they've applied a revert commit. We also might want to add a warning to notify them that all patches to core will be overwritten

source/_docs/upstream-updates.md
```diff
@@ -165,3 +165,35 @@ There are multiple reasons that 503 errors might occur when updating:
...
+    * Wordpress: `/wp-content` and `wp-config.php`
+    * Drupal 8: `/sites` and `/modules`
+    * Drupal 7: `/sites`
+
+5. Stage, commit, and push your changes to Pantheon:
+
+        git commit -am "Replace core files with Pantheon upstream"
+        git push origin updates
+
+6. Test changes on a Multidev environment and [merge the update](/docs/multidev/#merge-code) branch into the Dev environment, then deploy up to Test and Live.
```
dustinleblanc: May not be available to all customers

source/_docs/upstream-updates.md
```diff
@@ -165,3 +165,35 @@ There are multiple reasons that 503 errors might occur when updating:
...
+
+4. Use a file manager, such as [Finder for MacOS](https://support.apple.com/en-us/HT201732) or [File Explorer for Windows](https://support.microsoft.com/en-us/help/17217/windows-10-whats-changed-in-file-explorer), to copy the contents of the downloaded archive to your site repository excluding paths where custom code is set.
+
+  Exclude at least the following directories when copying, in addition to any other paths your site uses to store custom code:
+
+    * Wordpress: `/wp-content` and `wp-config.php`
+    * Drupal 8: `/sites` and `/modules`
+    * Drupal 7: `/sites`
+
+5. Stage, commit, and push your changes to Pantheon:
```
dustinleblanc: It may be worth calling out a review of the changes as an additional step here. That might be out of scope though so not sure.

rachelwhitton:
I still feel like this is too risky to recommend in the docs, but I don't want to be a blocker if CSE agrees it's needed. +1

alexfornuto:
Closed due to lack of support.

PR #2536, opened 2017-06-02
ssh authenticity prompt and pro tip

ilovelamp:
https://pantheon.io/docs/git/

Explains that they can "yes" the SSH authenticity prompt.

Also adds a "PRO TIP" about configuring StrictHostKeyChecking=no for  *.drush.in entries your SSH config file.


rachelwhitton:
+1  Great addition to the docs, thanks @ilovelamp 

I updated our global usage of accordion styles and made some style tweaks while I was at it, sorry for the noise

PR #2539, opened 2017-06-02
correcting pantheon settings include file name

dustinleblanc:
see https://github.com/pantheon-systems/drops-8/tree/master/sites/default


dustinleblanc:
Wow this diff is rediculous, I only changed 3 characters... from inc to php

alexfornuto:
Yea, that is ridiculous. I rebased to no avail, so instead I reproduced your fix in #2540

PR #2543, opened 2017-06-05
Do not set SFTP remote path

ilovelamp:
RE: https://pantheon.io/docs/sftp/

Instructs users to keep the Remote Path setting in their SFTP client blank.

Occasionally users will ask how to find their binding path, or they hardcode it after logging in once and it is then incorrect after the container changes due to system maintenance.


alexfornuto:
👍

PR #2547, opened 2017-06-07
toggling connection mode discards uncommitted files

ilovelamp:
Occasionally chat requests come in asking how to reset the 'state' of their Dev site.

This PR provides instructions regarding toggling the Connection Mode setting to blow away any uncommitted files.

Added as its own section and also a bullet-point under the SFTP warnings box (because its important!)


rachelwhitton:
+1

PR #2551, opened 2017-06-08
Remove outdated information.

### LINKED ISSUE #2550: Starting With Git Doc Update
Re: [Starting With Git](https://pantheon.io/docs/git/)

## Issue Description:
There's a section which suggests using diffie-hellman-group1-sha1 as a preferred method for improving the connection performance. Being a simpler encryption method it does provide a speed benefit at the cost of a weaker security. 

SHA1 has already been broken (a collision was found) in an experiment run by Google recently and Diffie-Hellman doesn't provide forward secrecy. I'm not really comfortable with the suggestion here.

https://www.theverge.com/2017/2/23/14712118/google-sha1-collision-broken-web-encryption-shattered

Also, this algorithm is only proposed as a legacy option in openssh - http://www.openssh.com/legacy.html

## Suggested Resolution 
I haven't seen any complaints related to connection problems due to a cipher mismatch (as a trend I mean). I propose we eliminate this section.

alexfornuto:
Closes #2550 

## Effect
- Removes outdated information.

## Remaining Work
- [x] Review

### DIFF
--- a/source/_docs/git.md
+++ b/source/_docs/git.md
@@ -141,16 +141,6 @@ Now, requests to any `*.drush.in` server address should automatically accept the
 </div>
 </div>
 
-
-
-
-### Git Connection is Slow
-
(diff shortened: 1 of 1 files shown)


rachelwhitton:
+1

PR #2558, opened 2017-06-08
Updated Email Recommendations

taurusismysign:
Updated Email Recommendations for higher deliverability.

Closes #IO-2365

## Effect
- 3 different recommendations
- Emphasis on using port 2525, if SMTP
- Emphasis on using REST API module/ plugins over SMTP mode

### DIFF
--- a/source/_docs/email.md
+++ b/source/_docs/email.md
@@ -10,7 +10,7 @@ Pantheon does not host inboxes for incoming mail. We recommend using an external
 
 ## Outgoing Email
 
-For outgoing emails, we recommend integrating a third-party service provider that supports a REST API configuration. You can use a SMTP configuration, but because SMTP requests are associated with dynamic outgoing IPs there can be negative impacts to deliverability. For a detailed comparison between API configurations and SMTP, see [this related blog post from SendGrid](https://sendgrid.com/blog/web-api-or-smtp-relay-how-should-you-send-your-mail/).
+For outgoing emails, we recommend integrating a third-party service provider that supports a REST API configuration. You can use an SMTP configuration, but because SMTP requests are associated with dynamic outgoing IPs there can be negative impacts to deliverability. For a detailed comparison between API configurations and SMTP, see [this related blog post from SendGrid](https://sendgrid.com/blog/web-api-or-smtp-relay-how-should-you-send-your-mail/).
 
 
 ### REST API Providers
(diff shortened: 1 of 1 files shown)


source/_docs/email.md
```diff
@@ -35,14 +70,60 @@ While many providers have standalone modules/plugins, you may want or need to in
...
+      <tr>
+        <th scope="row" class="thead-inverse">Amazon SES</th>
+        <td><a href="http://docs.aws.amazon.com/ses/latest/DeveloperGuide/smtp-connect.html" target="_blank">2587 (STARTTS), 2465 (TLSWRAPPER)</a></td>
+      </tr>
+    </tbody>
+</table>
+
+If you do not find your service provider in the table above, kindly check with their support/ documentation.
+
+#### Approach 3: Pantheon via Local MTA (may result in emails not being delivered)
```
rachelwhitton: @taurusismysign since this is not recommended, can we remove it a suggestion? Seems best to direct users towards more reliable configurations.

taurusismysign: @rachelwhitton I wonder if we remove it, whether CSE will have an overhead of answering a question like "Do you not at all support local MTA?"/ "What if, I don't want to use external provider - does that mean I can send no emails?" .. stuff like that. And then they will have to maintain another set of response somewhere.

I think I can work on pulling out from that list and putting it into a separate bucket to make it appear as unsupported feature. Thoughts?

rachelwhitton:
+1!

source/_docs/email.md
```diff
@@ -35,17 +61,54 @@ While many providers have standalone modules/plugins, you may want or need to in
 
     <div class="alert alert-info" role="alert">
       <h4 class="info">Note</h4>
-        <p>Support for Drupal 8 is not yet available for the <a href="https://www.drupal.org/project/smtp"> SMTP Authentication Support</a> module. However, <a href="https://groups.google.com/a/pantheon.io/forum/#!topic/power-users/HxvK7T0MPEM"> some users</a> have reported success with the pre-release version. For details, see <a href="/docs/guides/sendgrid/#sendgrid-smtp-integration"> this article</a>.</p>
+        <p>Support for Drupal 8 is in Beta. <a href="https://www.drupal.org/project/smtp"> SMTP Authentication Support</a> module.</p>
```
alexfornuto: This line ends with a fragment. Not sure if it should be restructured or removed, as the same link lives above.

source/_docs/email.md
```diff
@@ -62,13 +125,18 @@ This is a common error with the SMTP Authentication Support module. It can be fi
...
+
+### What ports are blocked by Pantheon?
+In order to fight spam, Pantheon blocks traffic on popular email ports 25, 465, and 587. Your emails will not be processed if you use any of these ports. Make sure that your service provider allows traffic on a port other than 25, 465 and 587 and that you have correctly configured your site to use that port.
 
-#### Can Pantheon provide, publish, or support SPF records?
+### Can Pantheon provide, publish, or support SPF records?
 
 As consumers of cloud infrastructure, we don’t have control over our IP ranges and they are subject to change without our notice. Publishing an SPF record would imply assurance on our end that it can work, which would be very difficult to guarantee given these circumstances. We take the decision of what we support and what we don’t very seriously, so at this time we’re not in a position to do that due to sustainably.
 
 At this time we do not support email off the platform, and recommend using a third-party solution like those listed above instead.
```
alexfornuto: ...Should this be "on" the platform?

PR #2559, opened 2017-06-09
User Testing Results: Move Workflow Diagram Up

rachelwhitton:
Re: AL-1107

## Effect
- Moves the workflow diagram above the fold so users can get to it sooner

cc @kimby77 @ronan


kimby77:
@rachelwhitton what are your thoughts on having this workflow section come after the "Benefits" section?  So that users can learn why they would use this first, and then they learn how to use it.

kimby77:
👍

PR #2571, opened 2017-06-14
clarify steps for getting IP for hosts file

ilovelamp:
Clarify how to get IP address via "Show DNS Recommendations" when testing site before go live by using the hosts file method.


alexfornuto:
Thanks!

PR #2579, opened 2017-06-16

ttrowell:
Add Bookly to the list

-Adds Bookly to the list
-
-


source/_docs/unsupported-modules-plugins.md
```diff
@@ -269,6 +269,13 @@ define('AUTOPTIMIZE_CACHE_CHILD_DIR','/uploads/autoptimize/');
 ```
 For additional details, see the [Autoptimize FAQ](https://wordpress.org/plugins/autoptimize/faq). An alternative solution is to [create a symbolic link](/docs/assuming-write-access/#create-a-symbolic-link).
 
+<hr>
+
+### [Bookly](https://wordpress.org/plugins/bookly-responsive-appointment-booking-tool/)
```
rachelwhitton: @ttrowell @alexfornuto can we open an issue on the plugin's support page and crosslink? Are there any known workarounds?

alexfornuto: [done](https://wordpress.org/support/topic/incompatibility-with-wp-native-sessions/)

PR #2587, opened 2017-06-20
Adding quotes might help readability

obicke:
... especially around "action required".

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed


alexfornuto:
Oooh, @obicke I love improvements to readability! Before I "merge" this, let me _ask_ you: Are these notices something we should "quote", or reference **In Bold**?

obicke:
To be honest, I don't know what the general approach is. My *guess* is that they should be quoted since we aren't really wanting to draw additional attention to them on the page, we just want to clarify how they should be interpreted within the text.

alexfornuto:
That's why I gotta get off my butt and write a style guide! Until then, your points jive with me. Thanks for contributing!

PR #2610, opened 2017-06-28
Updated email recommendations

taurusismysign:
Discourage use of popular ports - 25, 465 and 587 instead of Pantheon blocking them.

Closes IO-2469

## Effect
- We do not block traffic on 3 ports anymore
- We discourage use of those 3 ports and suggest to use a different port for emails


rachelwhitton:
+1 thanks @taurusismysign!

PR #2633, opened 2017-07-04
Update quicksilver.md

twooten:
## Remaining Work
- [x] needs some wordsmithing here. Basically the quicksilver scripts (private/scripts) need to be located inside of /code/web if the site is using a relocated docroot.


rachelwhitton:
+1! Thanks @twooten

PR #2635, opened 2017-07-05
sftp with password for Google SSO users

ilovelamp:
Tips for SFTP via password for Google SSO users who don't have a password.

Walks them through resetting password for dashboard, while still recommending SSH keys as preferred solution.


rachelwhitton:
+1 Thanks @ilovelamp!

alexfornuto:
👍

PR #2663, opened 2017-07-13
PHP syntax typo: extra } in drushrc.php

ccharlton:
## Effect
- PHP syntax fix


alexfornuto:
👍  Nice catch, thanks!

PR #2668, opened 2017-07-14
Fix Cloudflare capitalization. Clarify Cache Lookups and Clearing.

davidstrauss:


davidstrauss:
Granular cache clearing *is* available in Legacy Edge (and is actually the only implementation now). It's just much slower than Global CDN.

rachelwhitton:
+1

PR #2669, opened 2017-07-14
Multiple improvements to HTTPS redirect instructions

davidstrauss:
This updated guide:

* Dynamically defines `$canonical_domain` as either a specific value (for Live) or the current hostname (non-Live).
* Consolidates www and non-www cases by using a `$canonical_domain` variable.
* Redirects to `$canonical_domain` from every other subdomain, not just www.
* Redirects web requests to HTTPS on every Pantheon environment.
* Tells Drupal 7 to use `https://$canonical_domain` as the base for all absolute URLs.
* Tells Drupal 8 that the incoming domain is trusted (which is always true on Pantheon, anyway).
* Tells WordPress to use `https://$canonical_domain` as the base for all absolute URLs.

I haven't tested this yet, but feedback is welcome.

### DIFF
--- a/source/_docs/guides/launch/05-redirects.md
+++ b/source/_docs/guides/launch/05-redirects.md
@@ -38,7 +38,7 @@ In this lesson, we'll redirect incoming requests to a primary domain name (e.g.,
     <!-- Tab panes -->
     <div class="tab-content">
     <div role="tabpanel" class="tab-pane active" id="wp">
-    <p markdown="1">Comment out existing code that sets `WP_SITEURL`, such as [Line 79 in Pantheon's WordPress upstream](https://github.com/pantheon-systems/WordPress/blob/master/wp-config.php#L79), otherwise you will get a PHP notice. Then add the following to the end of your `wp-config.php` file (replace `www.example.com`):</p>
+    <p markdown="1">Add the following to `wp-config.php` (replace `www.example.com`):</p>
     <pre id="git-pull-wp"><code class="php hljs" data-lang="hljs">if (isset($_SERVER['PANTHEON_ENVIRONMENT']) && php_sapi_name() != 'cli') {
       // Redirect to https://$primary_domain/ in the Live environment
       if ($_ENV['PANTHEON_ENVIRONMENT'] === 'live'):
--- a/source/_docs/guides/launch/05-redirects.md
+++ b/source/_docs/guides/launch/05-redirects.md
@@ -47,7 +47,9 @@ In this lesson, we'll redirect incoming requests to a primary domain name (e.g.,
         header('Location: '. $base_url . $_SERVER['REQUEST_URI']);
         exit();
       }
-      $settings['trusted_host_patterns'] = array('^'. preg_quote($canonical_domain) .'$');
+      if (is_array($settings)) {
+        $settings['trusted_host_patterns'] = array('^'. preg_quote($canonical_domain) .'$');
+      }
     }
     </code></pre>
 
--- a/source/_docs/guides/launch/05-redirects.md
+++ b/source/_docs/guides/launch/05-redirects.md
@@ -26,52 +26,30 @@ In this lesson, we'll redirect incoming requests to a primary domain name (e.g.,
   If you run into issues, please refer to [this documentation](/docs/sftp/#sftp-connection-information).
 
 4. Now open the `code` folder in your SFTP client, and download your site's `settings.php` (Drupal) or `wp-config.php` (WordPress) file.
-5. Edit your configuration file by adding the following snippet for the desired redirect (replace `example.com`):
+5. Edit your configuration file by adding the following snippet for the desired redirect (replace `www.example.com`):
 
-    <ul class="nav nav-tabs" role="tablist">
-      <li role="presentation" class="active"><a href="#www" aria-controls="www" role="tab" data-toggle="tab">Redirect to HTTPS with www</a></li>
-      <li role="presentation"><a href="#non-www" aria-controls="non-www" role="tab" data-toggle="tab">Redirect to HTTPS without WWW</a></li>
-    </ul>
-
-    <!-- Tab panes -->
-    <div class="tab-content">
-      <div role="tabpanel" class="tab-pane active" id="www">
-      <br>
-      <pre><code class="php hljs">
-    // Require https with www
-    if (isset($_SERVER['PANTHEON_ENVIRONMENT']) &&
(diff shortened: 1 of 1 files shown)


source/_docs/guides/launch/05-redirects.md
```diff
@@ -26,52 +26,33 @@ In this lesson, we'll redirect incoming requests to a primary domain name (e.g.,
...
-        header('HTTP/1.0 301 Moved Permanently');
-        header('Location: https://www.example.com'. $_SERVER['REQUEST_URI']);
-        exit();
+    <pre><code class="php hljs">
+    // Redirect to https://$canonical_domain/
+    if (isset($_SERVER['PANTHEON_ENVIRONMENT']) && php_sapi_name() != 'cli') {
+      if ($_SERVER['PANTHEON_ENVIRONMENT'] === 'live') {
+        // Edit www.example.com to be the desired, canonical domain.
+        // Remove www to always redirect to non-www.
+        $canonical_domain = 'www.example.com';
```
ari-gold: How about moving this line to the top of the snippet? I'd also prefer we change variable name to the more colloquial #primary_domain

source/_docs/guides/launch/05-redirects.md
```diff
@@ -26,52 +26,33 @@ In this lesson, we'll redirect incoming requests to a primary domain name (e.g.,
...
-      // Check if Drupal or WordPress is running via command line
-      (php_sapi_name() != "cli")) {
-      if ($_SERVER['HTTP_HOST'] != 'www.example.com' ||
-          !isset($_SERVER['HTTP_X_SSL']) ||
-          $_SERVER['HTTP_X_SSL'] != 'ON' ) {
-        header('HTTP/1.0 301 Moved Permanently');
-        header('Location: https://www.example.com'. $_SERVER['REQUEST_URI']);
-        exit();
+    <pre><code class="php hljs">
+    // Redirect to https://$canonical_domain/
```
ari-gold: // Redirect to https://$primary_domain/ in the Live environment

source/_docs/guides/launch/05-redirects.md
```diff
@@ -26,52 +26,33 @@ In this lesson, we'll redirect incoming requests to a primary domain name (e.g.,
...
+    if (isset($_SERVER['PANTHEON_ENVIRONMENT']) && php_sapi_name() != 'cli') {
+      if ($_SERVER['PANTHEON_ENVIRONMENT'] === 'live') {
+        // Edit www.example.com to be the desired, canonical domain.
+        // Remove www to always redirect to non-www.
+        $canonical_domain = 'www.example.com';
+      } else {
+        $canonical_domain = $_SERVER['HTTP_HOST'];
```
ari-gold: // Redirects web requests to HTTPS on every Pantheon environment.

source/_docs/guides/launch/05-redirects.md
```diff
@@ -26,52 +26,33 @@ In this lesson, we'll redirect incoming requests to a primary domain name (e.g.,
...
+      $base_url = 'https://'. $canonical_domain;  // For the redirect and Drupal 7
+      if ($_SERVER['HTTP_HOST'] != $canonical_domain
+          || !isset($_SERVER['HTTP_X_SSL'])
+          || $_SERVER['HTTP_X_SSL'] != 'ON' ) {
         header('HTTP/1.0 301 Moved Permanently');
-        header('Location: https://example.com'. $_SERVER['REQUEST_URI']);
+        header('Location: '. $base_url . $_SERVER['REQUEST_URI']);
         exit();
       }
+      define('WP_SITEURL', $base_url);  // For WordPress
```
rachelwhitton: We'll need to instruct users to [comment out line 79](https://github.com/pantheon-systems/WordPress/blob/master/wp-config.php#L79), otherwise they'll see a PHP warning: `Notice: Constant WP_SITEURL already defined in /srv/bindings/29b987e6445747e6834324e3729f004d/code/wp-config.php on line 196`

rachelwhitton:
+1 

I pulled the framework specifics into their own tabs

source/_docs/guides/launch/05-redirects.md
```diff
@@ -25,53 +25,90 @@ In this lesson, we'll redirect incoming requests to a primary domain name (e.g.,
...
-    // Require https with www
-    if (isset($_SERVER['PANTHEON_ENVIRONMENT']) &&
-      ($_SERVER['PANTHEON_ENVIRONMENT'] === 'live') &&
-      // Check if Drupal or WordPress is running via command line
-      (php_sapi_name() != "cli")) {
-      if ($_SERVER['HTTP_HOST'] != 'www.example.com' ||
-          !isset($_SERVER['HTTP_X_SSL']) ||
-          $_SERVER['HTTP_X_SSL'] != 'ON' ) {
+    <div role="tabpanel" class="tab-pane active" id="wp">
+    <p markdown="1">Comment out existing code that sets `WP_SITEURL`, such as [Line 79 in Pantheon's WordPress upstream](https://github.com/pantheon-systems/WordPress/blob/master/wp-config.php#L79), otherwise you will get a PHP notice. Then add the following to the end of your `wp-config.php` file (replace `www.example.com`):</p>
```
rachelwhitton: @davidstrauss mentioned that we should put this configuration below the define() for that and simply consume the value from it, but I'm not entirely sure how to accomplish that in this snippet. Any suggestions/tips @ari-gold @lauren-kelly?

davidstrauss: As long as they set it on line 79 and the code snippit is added below it (maybe right below it would be easiest?), it should be possible to use `WP_SITEURL` instead of `$primary_domain`.

davidstrauss: This one is tricky because we need the domain name as well. That's part of the `WP_SITEURL`, but it would require parsing to get it out. So, it's probably best to have our code snippit added right below the `define()` for `WP_SITEURL` and alter our comparisons to expect the data it provides. For the configuration to work properly, it's really important that value defined for `WP_SITEURL` begins with "https" and not just "http".

So, what I would do in the instructions is something like (1) tell users to configure `WP_SITEURL` (on line 79) with their site's HTTPS base URL, (2) change the comparison from `$_SERVER['HTTP_HOST'] != $primary_domain` to `'https://'. $_SERVER['HTTP_HOST'] != WP_SITEURL`, and (3) convert all use of `$base_url` to `WP_SITEURL`.

PR #2677, opened 2017-07-19
Cron Clarity

### LINKED ISSUE #2233: Cron for Drupal Doc is confusing
Re: [Cron for Drupal](https://pantheon.io/docs/drupal-cron)

## Issue Description:

This page is confusing. The following statements are made.

> Pantheon internally invokes drush pantheon_cron 3600.
> This bootstraps your site and invokes drupal_cron_run.

> Pantheon's system cron cannot be disabled and will run Drupal cron every hour, even if cron is disabled in Drupal.
> In Drupal, setting the value to "Never" will be ignored; system cron will always run at least hourly.

> Can I prevent Drupal Cron from running?

> Yes, from within Drupal at admin/system/config/cron, select Never from the "Run cron every" drop-down menu, then save the configuration. However, Pantheon system cron will run every hour and there is no way to disable it. 

## Suggested Resolution 

Explain the difference between Pantheon's internal cron and exactly what is does. How is it different from regular Drupal Cron?

alexfornuto:
Closes #2233 

## Effect
- More info on cron
-
-

## Remaining Work
- [ ] Copy Review

### DIFF
--- a/source/_docs/drupal-cron.md
+++ b/source/_docs/drupal-cron.md
@@ -38,6 +38,8 @@ Alternatively, you can run all scheduled cron tasks with the following [Terminus
 terminus drush <site>.<env> -- cron
 ```
 
+Here's some change, wherein I've described a solution to an issue brought up with this doc.
+
 To ensure that cron tasks have been run, check the reports via the Drupal Admin interface at Reports > Recent Log Messages. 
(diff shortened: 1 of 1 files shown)


alexfornuto:
This was an example PR made for training.

PR #2628, opened 2017-07-03
Early Access: Self Serve Custom Upstreams

rachelwhitton:
Re: [AL-1073](https://getpantheon.atlassian.net/browse/AL-1073) & [AL-1132](https://getpantheon.atlassian.net/browse/AL-1132)

## Effect
PR includes the following changes:
- Repurpose create-custom-upstream.md for early access doc
- Add example usage to Terminus manual: Switch Upstreams

## Remaining Work
- [x] Verify eligibility
- [x] Peer review with Alchemy

PR #2682, opened 2017-07-20
Clean Up Self Serve Custom Upstreams

rachelwhitton:
Implements feedback from #2628 

## Effect
- Add instructions to resolve merge conflicts to existing doc on applying upstream updates
- Crosslink terminus usage to resolving merge conflicts in the section on applying updates
- Add org screenshot 

## Todo
@kimby77 re: [your original comment](https://github.com/pantheon-systems/documentation/pull/2628#issuecomment-314883550), I'm not sure how to resolve the last bullet (safeguards). Might need some help from you and/or @ronan for finishing that up. 

## Comments
We do have a note in the switching upstreams section regarding composer sites. It's not 💯  but I think it's enough for our first iteration. Once more product work is done on composer we'll need a standalone doc on composer + custom upstreams (maybe). Defer to you and @ronan


kimby77:
image
We modified these framework options, so can you change this to: "Drupal 7, Drupal 8, or WordPress"

image

Can you add "initial site connection mode" here as something available to modify?  We don't have this during the creation process as it's mostly unnecessary for the user to think about at that time, but if they need to modify this due to some set up requirement with their own code, we should add it here in some way so that it's findable.

Re: safeguards, how about adding a line in between this command and the blue box, to the extent of:
image
"Note that Terminus will prevent a framework switch (ie. a Drupal site can only switch to another Drupal upstream)."

This framework question was something users did ask about, so we should cover it here to proactively answer it.  If we add more safeguards in the future we can think about how to expand this or make it its own section, but we don't need to solve that yet.

Everything else LGTM!

kimby77:
👍

PR #2689, opened 2017-07-24
trello guide

sukottokun:
Hiya @rachelwhitton this uses a pending QS integration script [here](https://github.com/pantheon-systems/quicksilver-examples/pull/133)


rachelwhitton:
+1! Thanks @sukottokun - once pantheon-systems/quicksilver-examples/pull/133 is merged we can ship this.

PR #2690, opened 2017-07-24
Github Pull Requests doc - missing a crucial step in 'before you begin'

shaal:
Adding "Authorize CircleCI on Github" to the "Before you begin" section.
Without this step, CircleCI tests will always fail.


rachelwhitton:
Nice! Thanks @shaal!! +1

PR #2692, opened 2017-07-25
Revise switch DNS doc

ari-gold:
TODO: Remove HTTPS caveat once we come up with process that involves no interruption in HTTPS


rachelwhitton:
+1

PR #2698, opened 2017-07-25
Delete Legacy HTTPS Docs

rachelwhitton:
CC @ari-gold 
Retire legacy docs and update links throughout

## Required Redirects Post Launch
- [ ] /docs/build-https => /docs/https
- [ ] /docs/domains => /docs/guides/launch/domains
- [ ] /docs/enable-https-vanity-domain => /docs/https
- [ ] /docs/enable-https => /docs/https
- [ ] /docs/ssl-tls => /docs/https
- [ ] /docs/videos/domains => /docs/guides/launch/domains


source/_docs/build-https.md
```diff
@@ -1,17 +0,0 @@
----
-title: Developing with HTTPS
```
ari-gold: I think we need to preserve this content somewhere, but maybe we can deal with that in http-to-https

source/docs.html
```diff
@@ -161,7 +161,7 @@ <h2 class="subtitle">Popular Documentation</h2>
         <li><a href="/docs/terminus/">Terminus Manual</a></li>
         <li><a href="/docs/guides/getting-started/">Getting Started Guide</a></li>
         <li><a href="/docs/platform-considerations/">Platform Considerations</a></li>
-        <li><a href="/docs/domains/">Domains and DNS</a></li>
+        <li><a href="/docs/guides/launch/">Luanch Essentials</a></li>
```
ari-gold: typo

source/_docs/build-https.md
rachelwhitton: What specifically do we need here that isn't already in http-to-https?

ari-gold:
+1 :shipit: 🚢  🚀

PR #2701, opened 2017-07-25
Update Cloudflare domain configuration

ari-gold:
No need to redirect in PHP if doing via CF.

### DIFF
--- a/source/_docs/dns-providers/cloudflare.md
+++ b/source/_docs/dns-providers/cloudflare.md
@@ -52,7 +52,7 @@ You can configure Cloudflare's CDN as an additional layer on Pantheon's Global C
 2. Scroll down and enable **Always use HTTPS**
   ![Cloudflare Always HTTPS](/source/docs/assets/images/cloudflare-always-https.png)
 3. Scroll down and enable **Automatic HTTPS Rewrites**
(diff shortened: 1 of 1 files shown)


davidstrauss:
I'd prefer that we encourage people to enable the Cloudflare "Action" to redirect to HTTPS, but we definitely should not be advising people to redirect in PHP when Cloudflare is in "orange cloud" mode.

rachelwhitton:
@davidstrauss can you clarify how to enable the Action? I can't find seem to find it

Also, I haven't been able to stack Cloudflare so that multiple redirects happen in one response. Then again, single response may not be a blocker but defer to you and @ari-gold

davidstrauss:
Here's the action to redirect to HTTPS:

screenshot from 2017-07-26 13-47-48

> Also, I haven't been able to stack Cloudflare so that multiple redirects happen in one response. Then again, single response may not be a blocker but defer to you and @ari-gold

There isn't really such a thing. It's more of a "can we get everything done in one redirect" question. For example, if I go to `http://example.com`, do I first get a redirect to `https://example.com` and then to `https://www.example.com`, or do I directly get the final URL as the redirect?

rachelwhitton:
+1

PR #2723, opened 2017-07-31
Update private-paths.md

### LINKED ISSUE #2619: Private Paths for Files and Code Doc Update
Re: [Private Paths for Files and Code](https://pantheon.io/docs/private-paths/#resolving-warning-file_put_contentsprivate:.htaccess)

## Issue Description:
When .htaccess file is present and you get the error noted, you have to first delete the .htaccess file THEN you have to resubmit in Drupal Admin

## Suggested Resolution 
Test and add step to delete .htaccess if valid.

ilovelamp:
The problem isn't the .htaccess file, its that the private directory doesn't exist or is unaccessible (i.e. trying to enable the option on Test or Live when using the Private Path for Code and it not being able to write to the *private* folder.

I clarified the page by:

* Removing leading slash from /private and /files as it is misleading: We're not referring to the real root of the file system.

* Added instructions on creating path, as its not an .htaccess issue specifically but a directory found/exists issue AFAIK.

Closes #2619

### DIFF
--- a/source/_docs/private-paths.md
+++ b/source/_docs/private-paths.md
@@ -13,7 +13,7 @@ Drupal: `sites/default/files/private`
 WordPress: `wp-content/uploads/private`
 
 **Private Path for Code (Version Controlled)**  
-Drupal and WordPress: `/private`   
+Drupal and WordPress: `private`   
 
 <div class="alert alert-info" role="alert">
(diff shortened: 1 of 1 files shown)


rachelwhitton:
Nice! +1 

Thanks for the PR @ilovelamp!

PR #2733, opened 2017-08-02
New Sites and Public Distributions

### LINKED ISSUE #2724: Drupal Distributions Documentation
Re: /docs/start-state
## Issue Description
Need clarity what to really do when using Drupal distributions.
We started to tell users to just install vanilla Drupal, then commit the installation profile and continue

## Suggested Resolution
Provide different methods on actual installation of distribution, define profiles and actual steps for installation (commit profile first, before install) etc.,
Also, provide the reason why the previous upstream list is no longer available.

rachelwhitton:
Closes #2724 

## Effect
- Document public distros with direct links to create new sites on Pantheon

cc @eabquina @ronan @kim please review


source/_docs/distros.md
```diff
@@ -0,0 +1,94 @@
...
+  4fa4f7d1-e341-48f7-9594-4e7c21b9bb68   Drupal8 PHP                                   custom   drupal8
+  5b70ea4b-8ba3-4c78-9cf2-8086a5524885   Open Academy                                  custom   drupal
+  bd5bac3b-2ed4-4b0b-b085-b9cfb574940f   Open Enterprise (Deprecated)                  custom   drupal
+  10d6937e-1dd2-4490-9950-11867ba43597   RedHen Raiser                                 custom   drupal
+  21e1fada-199c-492b-97bd-0b36b53a9da0   Drupal 7                                      custom   drupal
+  2adff196-4672-44c9-af2a-4590963b90d8   OpenAid                                       custom   drupal
+  f575e77a-ff49-4bab-a53c-dbf8f59247bc   Empty Drupal 7                                custom   drupal
+  0fce9338-bd6f-4b87-a489-e9928a027696   Elephant Ventures Drupal Patterns             custom   drupal
+  974b75c2-4ba7-49f8-8a54-3a45c07dfe02   Drupal 6                                      custom   drupal
+  c28628b2-6b8f-4962-9838-69295c3fff68   Fairfax County Public Schools                 custom   drupal8
```
kimby77: Can we verify this list is accurate?  For example, I'm not sure if Fairfax County Public Schools should be there.  It's one agency's custom upstream.

source/_docs/distros.md
```diff
@@ -0,0 +1,94 @@
+---
+title: New Sites and Public Distributions
+description: Learn how to create sites using a public distribution start state.
+tags: [create]
+---
+The Pantheon Dashboard provides a quick "click to install" method of creating new sites that utilize our default, "vanilla" frameworks (e.g., WordPress, Drupal 7, and Drupal 8). However, it's possible to create sites that use an available public distribution (e.g., Panopoly) with Terminus, the Pantheon CLI.
```
kimby77: They can also go to a direct link for each distro. Can we try adding them here?

rachelwhitton: +1 I didn't realize that direct links are available. Do we have a list of them, or a format to follow so I can construct them?

source/_docs/distros.md
rachelwhitton: Yeah.. I saw that. I authenticated terminus as our test user (for when we gather screenshots). That user _shouldn't_ be a member of any org utilizing custom upstreams, so I think this is the bare minimum list.

source/_docs/distros.md
kimby77: @ronan do you know where to find them?

source/_docs/distros.md
kimby77: Yes - it's come up on my Terminus too.  I'm not a member of that agency's org though.  So I'm not sure how it's available. Perhaps there was an error made when it was added as a custom upstream....

kimby77:
+1

PR #2759, opened 2017-08-07
Chrome's mixed-content warnings vary by the type of mixed content

davidstrauss:
Some other tweaks as well. For example, "Secure HTTPS" is a bit redundant.


rachelwhitton:
+1

PR #2796, opened 2017-08-15
Revise pantheon-yml

ari-gold:
## Effect
- Emphasize pantheon.yml, de-emphasize pantheon.upstream.yml
- General revision and clean-up

## Remaining Work
- [ ] If needed


ari-gold:
@rachelwhitton @greg-1-anderson  Please review

source/_docs/pantheon-yml.md
```diff
@@ -111,6 +72,25 @@ Add `drush_version` to the top level of the `pantheon.yml` file to configure the
 drush_version: 8
 ```
 For more information, see [Managing Drush Versions on Pantheon](/docs/drush-versions).
+
+
+### Custom Upstreams and pantheon.upstream.yml
+Add a `pantheon.upstream.yml` file to your organization's custom upstream to set default settings for all downstream sites. Downstream sites can override the defaults at the site-level with pantheon.yml. Quicksilver workflows are not supported in pantheon.upstream.yml.
```
rachelwhitton: Suggest preserving original text that defines the upstream config file - seems like it's a bit friendlier than this iteration. For example, downstream is used without much context which can be confusing 

I also like how the original text clearly described the difference between the two


> The `pantheon.upstream.yml` file is used to set default configurations for sites created from a Custom Upstream, while the `pantheon.yml` file is used to manage configurations for one particular site. Both configuration files utilize the same set of properties that define how your Pantheon environments will behave.
>
> When the same configuration value is defined in both files, the value from `pantheon.yml` will override the value from `pantheon.upstream.yml`.

source/_docs/pantheon-yml.md
```diff
@@ -111,6 +72,25 @@ Add `drush_version` to the top level of the `pantheon.yml` file to configure the
 drush_version: 8
 ```
 For more information, see [Managing Drush Versions on Pantheon](/docs/drush-versions).
+
+
+### Custom Upstreams and pantheon.upstream.yml
```
rachelwhitton: Suggest h2 instead of h3 there, so this content isn't nested within the advanced site configuration section.

source/_docs/pantheon-yml.md
```diff
@@ -111,6 +72,25 @@ Add `drush_version` to the top level of the `pantheon.yml` file to configure the
...
+Add a `pantheon.upstream.yml` file to your organization's custom upstream to set default settings for all downstream sites. Downstream sites can override the defaults at the site-level with pantheon.yml. Quicksilver workflows are not supported in pantheon.upstream.yml.
+
+Create the `pantheon.upstream.yml` in the root of the custom upstream's codebase. Help avoid merge conflicts by adding pantheon.upstream.yml to the upstream's .gitignore. Example: [Adding pantheon.upstream.yml to Drops-7](https://github.com/pantheon-systems/WordPress/pull/123/files)
```
rachelwhitton: I can clean this bit up in a copy review, but I think Drops-7 should be WordPress here. Also, looks like we're already ignoring this by default in our upstream. So if folks follow the recommended path when creating custom upstreams, it should already ignore this file. Might not need to include instructions to ignore it now

source/_docs/pantheon-yml.md
```diff
@@ -134,7 +114,7 @@ remote: Valid versions are: 1
 While our parser will reject a `pantheon.yml` that is invalid, it won't necessarily give you the exact reason the file is invalid. Please refer to the examples above for exact syntax.
 
 ### Deploying Configuration Changes to Multidev
-Changes made to the `pantheon.yml` file on a branch **are not** detected when creating the Multidev environment for that branch. It is necessary to create the multidev first, and then modify pantheon.yml. As a workaround, make some modification the `pantheon.yml` file and re-commit to the Multidev environment. You will then receive a notice indicating configuration changes have been detected and applied to the Multidev environment:
+Changes made to pantheon.yml file on a branch **are not** detected when creating the Multidev environment for that branch. As a workaround, make some modification to `pantheon.yml` file and re-commit to the Multidev environment. You will then receive a notice indicating configuration changes have been detected and applied to the Multidev environment:
```
rachelwhitton: For consistency, I think we need to keep wrapping file names in code tags. If it looks weird or wrong, maybe we can make a style adjustment. Defer to @alexfornuto since he's heading up efforts on a style guide soon.

source/_docs/pantheon-yml.md
```diff
@@ -111,6 +72,25 @@ Add `drush_version` to the top level of the `pantheon.yml` file to configure the
...
+
+Because pantheon.upstream.yml is in the .gitignore use git force-add to commit:
+
+```bash
+$ git add -f pantheon.upstream.yml
+$ git commit -n "Change the default php version for this upstream to 7.0.x"
+```
+
+#### Nested docroot
+Set the `web_docroot` property in  upstream's `pantheon.upstream.yml` file.
```
rachelwhitton: Feels weird that this is the only advanced configuration we call out in context of custom upstreams. Can we add context? I think we need to preserve our original warning describing the implications of overriding this setting downstream.

source/_docs/pantheon-yml.md
```diff
@@ -111,6 +72,25 @@ Add `drush_version` to the top level of the `pantheon.yml` file to configure the
...
 ```
 For more information, see [Managing Drush Versions on Pantheon](/docs/drush-versions).
+
+
+### Custom Upstreams and pantheon.upstream.yml
+Add a `pantheon.upstream.yml` file to your organization's custom upstream to set default settings for all downstream sites. Downstream sites can override the defaults at the site-level with pantheon.yml. Quicksilver workflows are not supported in pantheon.upstream.yml.
+
+Create the `pantheon.upstream.yml` in the root of the custom upstream's codebase. Help avoid merge conflicts by adding pantheon.upstream.yml to the upstream's .gitignore. Example: [Adding pantheon.upstream.yml to Drops-7](https://github.com/pantheon-systems/WordPress/pull/123/files)
+
+Because pantheon.upstream.yml is in the .gitignore use git force-add to commit:
```
rachelwhitton: Suggest adding a new h3 header here so this can be a section on how to edit and distribute changes to `pantheon.upstream.yml` -- I also think we need to keep something similar to the original text because it specified that changes to this file are distributed downstream

rachelwhitton:
+1 thanks for the PR @ari-gold!

PR #2800, opened 2017-08-16
Adding FAQ about mail logs

twooten:
The mail logs are not separate for each site, so if customers had access to these logs, they would have access to mail logs from other customers as well.

### DIFF
--- a/source/_docs/email.md
+++ b/source/_docs/email.md
@@ -126,11 +126,12 @@ This is a common error with the SMTP Authentication Support module. It can be fi
 See [available patch](https://drupal.org/node/1369736#comment-5644064).
 
 ## Frequently Asked Questions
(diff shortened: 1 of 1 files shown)


alexfornuto:
+1

source/_docs/email.md
```diff
@@ -126,6 +126,9 @@ This is a common error with the SMTP Authentication Support module. It can be fi
 See [available patch](https://drupal.org/node/1369736#comment-5644064).
 
 ## Frequently Asked Questions
+### Can I access the mail logs for my site?
+No. The mail logs are not accessible.
+
```
rachelwhitton: I'd like to add context for why they are not accessible or move this down the list a bit

rachelwhitton:
+1

PR #2807, opened 2017-08-16
Update wordpress-cron.md

### LINKED ISSUE #2765: Highlight difference between Cron, WP-Cron
Re: [Cron for WordPress](https://pantheon.io/docs/wordpress-cron/)

Priority (Low):

## Issue Description:
User claims that WP is not running and caused by Pantheon. Our docs explanation says it might be because of site visit which causes a stir. 

## Suggested Resolution 
A link that shows WP-Cron in WP Codex on how it works. 
Also, highlight difference between "Drush" cron ran by Pantheon and that crontab is not in play here.

ttrowell:
Highlighting how WP-con is triggered

Closes #2765 

## Effect
-see above

## Remaining Work
- [ ] If needed


rachelwhitton:
+1

PR #2822, opened 2017-08-21
Disable Caching for Specific Pages example

ilovelamp:
Update example to define $regex_path_match, as we're seeing multiple support cases recently where the example is simply cut-and-pasted, resulting in an error because `$regex_path_match` is not defined.


alexfornuto:
@ilovelamp Please take a look at the latest version and tell me if you agree with the current formatting and methodology.

alexfornuto:
👍  achieved from @ilovelamp over another medium. Trust me...

PR #2823, opened 2017-08-21
Update https.md

lauren-kelly:
Closes #

## Effect
- adding # of domains included to FAQ

## Remaining Work
I hate duplication of info, but we really need this here and on the domains/ doc (which is where I C&P'd from)


ari-gold:
+1 for including the FAQ, but let's just link to the answer in the other doc so we don't need to duplicate everything here.

rachelwhitton:
+1 

I think it's fine to repeat info, rather than linking to another page. I put the table in a template file so we can easily reuse it on other pages while managing it from a single source.

alexfornuto:
bitmoji

PR #2826, opened 2017-08-22
Remove the value from the doing_wp_cron query var

mboynes:
Setting `doing_wp_cron=1` will prevent the cron job from running. References:

1. [Where the lock gets set](https://github.com/WordPress/WordPress/blob/9db227ac071c73ca70f5c7ca9492e3fc63131fab/wp-cron.php#L82)
2. [Where the lock gets checked](https://github.com/WordPress/WordPress/blob/9db227ac071c73ca70f5c7ca9492e3fc63131fab/wp-cron.php#L97)

### DIFF
--- a/source/_docs/wordpress-cron.md
+++ b/source/_docs/wordpress-cron.md
@@ -100,7 +100,7 @@ Once you have disabled WP-Cron, you will need a service that calls a URL at regu
 Any of the above services will get the job done. By disabling WP-Cron, you have turned off the automatic checking and calling of the `wp-cron.php` script. You will now have to call that URL yourself using one of the services above. With most of them, it is as easy as this:
 
 1. Set up an account
-2. Set up a job that calls `https://yourdomain.tld/wp-cron.php?doing_wp_cron=1`
+2. Set up a job that calls `https://yourdomain.tld/wp-cron.php?doing_wp_cron`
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
@ataylorme Do you have any input?

@dustinleblanc can you confirm this during your Doc Block?

ataylorme:
Thanks for reporting @mboynes. `cron.php` does look like it returns with a value set for `doing_wp_cron` ([line 256](https://core.trac.wordpress.org/browser/tags/4.8.1/src/wp-includes/cron.php#L256)).

Is there a reason you're including with without a value rather than just removing the parameter entirely?

mboynes:
@ataylorme you're looking in the wrong place, the link you sent is the `spawn_cron()` function, which is responsible for deploying the cron (making the remote request), not running it. See the links I included in my initial description, which show where the problem comes up.

alexfornuto:
Reached out to @ataylorme for further input.

ataylorme:
I think this is alright but would like a second opinion from @danielbachhuber

danielbachhuber:
👍

alexfornuto:
Thanks @danielbachhuber, @ataylorme for reviewing.  Thanks @mboynes for the PR!

PR #2829, opened 2017-08-23
Add turn-off New Relic for specific env

### LINKED ISSUE #2771: How to Disable New Relic Pro on specific environment
Re: [New Relic APM Pro](https://pantheon.io/docs/new-relic/)

Priority (Low):

## Issue Description:
User needs a way to disable New Relic (like the site page code)

## Suggested Resolution 
Need to highlight that this is not recommended as it will be difficult for Pantheon and customers to check issues for performance to their site

As a workaround, provide a sample snippet or code switch, to disable a site environment for New Relic (for edge case)

eabquina:
As an option and only for edge cases

Closes #2771 

## Effect
- Sample code snippet for disabling New Relic in a specific environment
-
-

## Remaining Work
- [ ] If needed

### DIFF
--- a/source/_docs/new-relic.md
+++ b/source/_docs/new-relic.md
@@ -201,6 +201,19 @@ New Relic Pro is automatically provisioned for your site. Unfortunately, you can
 
 Because Pantheon's runtime matrix runs your application across many containers simultaneously, it's common to see old containers with no reporting data as your application shifts around. This is not a cause for concern.
 
+#### Can I disable New Relic for a specific environment? ####
+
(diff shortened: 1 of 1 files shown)


alexfornuto:
@eabquina how many customers have you seen fit into the edge case where disabling NR for one environment made sense?

eabquina:
@alexfornuto I've handled only one so far and gave this sample snippet for them to use. 

This was the reason for that one ticket earlier:
"New relic is adding its javascript right under the <head> tag, which is creating issues in SEO because it is pushing meta tags too down"

PR #2830, opened 2017-08-23
Update getting-support.md

### LINKED ISSUE #2812: Get Support Doc Update
Re: [Get Support](https://pantheon.io/docs/getting-support/)

Priority (Medium):

## Issue Description:
User asks why Tickets are tied per user, which are not. They shold be per "Site" and can be seen per "Organizations"

## Suggested Resolution 
Highlight ticket view hierarchy, and roles that can see the tickets that were opened. 
Need to confirm if this is really per user, per site, and per organization.
We may need to modify content and say that if a ticket is created from which dashboard area, that is where it was tied first.

ttrowell:
Fix a typo and add clarification about how tickets are associated with sites.

Closes #2812

## Effect
-see above

## Remaining Work
- [ ] If needed

### DIFF
--- a/source/_docs/getting-support.md
+++ b/source/_docs/getting-support.md
@@ -92,7 +92,7 @@ The majority of Pantheon's support is provided in the context of a particular we
   </div>
 </div>
 
-All users of the platform can recieve customer service and support for non-site-related questions via any avalable medium regardless of account type.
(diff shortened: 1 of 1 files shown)


alexfornuto:
👍

PR #2835, opened 2017-08-24
Composer Doc Audit

### LINKED ISSUE #2665: Update pull request guide for WordPress
Re: https://pantheon.io/docs/guides/github-pull-requests/
## Issue Description

Our GitHub pull request guide is currently specific to Drupal. That is largely because Drupal 8 has been the focus of our Composer examples thus far. Soon we'll be adding an `example-wordpress-composer` that will be similar to `example-drops-8-composer` and compatible with the Terminus Build Tools Plugin

## Suggested Resolution

Update https://pantheon.io/docs/guides/github-pull-requests/ to make it more generic or just separately explain WordPress differences.

rachelwhitton:
Closes #2665 

## Effect
- Delete stale composer content throughout docs 
- Draft new composer overview page describing basic concepts and suggested workflows
- Move GitHub guide to paginated layout and rebrand as the Build Tools Guide

## Remaining Work
- [x] Peer review on the new [composer overview page](https://composer-static-docs.pantheonsite.io/docs/composer/) from @greg-1-anderson @allenfear and others cc @dustinleblanc @ari-gold @ataylorme @stevector @joshkoenig ? I'll leave some comments with areas I'd like specific help with (Ready for early review)
- [x] Refactor Build Tools Guide so that all frameworks are covered - I'd like to take a shot at a CMS agnostic guide first. If it's too polluted with all three I may change directions and try three single page CMS specific guides. (In progress)
- [x] Peer review on the Build Tools Guide once it has been refactored
- [ ] Redirects:
  - docs/guides/github-pull-requests => docs/guides/build-tools
  - docs/composer-drupal-8.md => docs/composer


source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+      </div>
+      <p class="topic-info__description"><a href="https://circleci.com">CircleCI</a> provides hosted services to run automated tests for a project, and GitHub provides an integration to run these tests to whenever a change is submitted. The process of testing each set of changed files prior to merging them into the main branch is called continuous integration.</p>
+    </div>
+  </div>
+</div>
 
-We recommend managing custom code as plugins or modules within individual repositories and hosting each one remotely on a service such as [GitHub](https://github.com/), [Bitbucket](https://bitbucket.org), or [GitLab](https://about.gitlab.com/). You can then use Composer to manage and organize the software needed for your project to run, including custom plugins or modules.
+Only files unique to the project are tracked as part of the project's main "source" repository on GitHub, which requires an abstraction layer to compile dependencies and deploy an entire "artifact" to the site repository on Pantheon. The abstraction layer is facilitated by CirlceCI in the Pantheon maintained examples, but the principles are the same for other continuous integration service providers.
```
rachelwhitton: Is this too technical? I tried to wordsmith it but got stuck

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+- Drupal 7: [https://packages.drupal.org/7](https://packages.drupal.org/7)
+
+Site's created from Pantheon's example repositories already include the appropriate package repository within the `composer.json` file.
+
+## Nested Docroot
+The docroot is the directory from which your site is served. Without Composer, Pantheon defaults the docroot to the root directory of the site's codebase. In order to require core as a project dependency, a recommended best practice for Composer managed sites, it must be installed in a subdirectory such as `web`.
+
+This is possible on Pantheon by specifying `web_docroot: true` in `pantheon.yml` file. For details, see [Serving Sites from the Web Subdirectory](/docs/nested-docroot/).
+
+## Source and Artifact Workflow
```
rachelwhitton: Thoughts on what I call these two workflows? (Source and Artifact Workflow and Custom Upstream Workflow)

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+
+![Multidev PR workflow](/source/docs/assets/images/pr-workflow/github-pr-diagram.png)
+
+### Automated Tests
+
+It is also common to set up automated tests to confirm that the project is working as expected; when tests are available, GitHub will run them and display the results of the tests with the pull request. Working on projects with comprehensive tests increases the development team's confidence that submitted pull requests will work correctly when they are integrated into the main build.
+
+### Scaling Considerations
+We recommend the "source" and "artifact" workflow for single site use cases, and for most use cases involving larger site portfolios such as EDUs. You can create a scaffold repository based off our example repositories and customize it to your liking, then use the scaffold to create new sites similar to how you would create the codebase for a [Custom Upstream](/docs/custom-upstream/).
```
rachelwhitton: I'm not 100% on the term "scaffold" or what a good alternative is.. 

Also, would love a spot check on technical content / advice. Does this work? Is it scalable as described? Do we need to provide a separate doc with explicit steps on how to do this?

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+
+## Custom Upstream Workflow
+It is possible to preserve the functionality of Pantheon's one-click updates in the Site Dashboard for Composer managed sites, however it's use case is quite narrow. Your scaffold repository based off our example repositories would need to commit all dependencies to the Custom Upstream repository. Updates via Composer would only happen at the Custom Upstream repository level, which would then trickle down to sites created from the Custom Upstream as one-click updates.
+
+This workflow has one very serious shortcoming, that is site-specific dependencies are likely to cause a lot of conflicts. The practical use case for this workflow is for a large group of sites that require a single set of dependencies. You should only use this method if you don’t intend to use site specific themes, modules, or plugins downstream.
```
rachelwhitton: Is this shortcoming effectively laid out? It seems vague, but when I expanded it felt way too technical.

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+
+## Next Steps
```
rachelwhitton: Should we add a CTA for the enablement page? Seems like composer based workflows at scale is something we'd want to offer training for, especially since use cases vary

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+
+### Scaling Considerations
+We recommend the "source" and "artifact" workflow for single site use cases, and for most use cases involving larger site portfolios such as EDUs. You can create a scaffold repository based off our example repositories and customize it to your liking, then use the scaffold to create new sites similar to how you would create the codebase for a [Custom Upstream](/docs/custom-upstream/).
+
+However, unlike Custom Upstreams this method does not support one-click updates in the Site Dashboard. Adopting this workflow means forgoing all other update techniques in favor of Composer. If your use case requires an simpler update strategy for non-technical site admins, this workflow could present problems scaling or at the very least require additional training for your development team.
```
rachelwhitton: Woops, typo. "If your use case requires an simpler update.."

alexfornuto: On L79 You're referencing Custom Upstreams before defining/explaining the term, which you do in the next section. I would suggest moving this down; going over custom upstreams before comparing the two methods.

source/_docs/composer.md
alexfornuto: I don't have any good input on that. Maybe a comparative naming convention that involves packages as being internal versus independent?

source/_docs/composer.md
alexfornuto: I could't possibly think of a simpler way to phrase that 👍

source/_docs/composer.md
alexfornuto: Other options:
Base
Core

source/_docs/composer.md
alexfornuto: Or alternately, start by saying what you can't do, and allude that the next section will cover a method which will handle it.

source/_docs/guides/build-tools/01-introduction.md
```diff
@@ -0,0 +1,112 @@
...
+        </div>
+
+    - The [Terminus Build Tools Plugin](https://github.com/pantheon-systems/terminus-build-tools-plugin):
+
+        <div class="copy-snippet">
+          <button class="btn btn-default btn-clippy" data-clipboard-target="#build-tools-plugin">Copy</button>
+          <figure><pre id="build-tools-plugin"><code class="command bash" data-lang="bash">curl https://github.com/pantheon-systems/terminus-build-tools-plugin/archive/2.0.0-alpha2.tar.gz -L | tar -C ~/.terminus/plugins -xvz</code></pre></figure>
+        </div>
+
+3. [Authorize CircleCI on Github](https://github.com/login/oauth/authorize?client_id=78a2ba87f071c28e65bb).
```
rachelwhitton: This link doesn't work for me

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+
+- A list of dependencies
+- A place to put the dependencies
+
+Understanding how Composer can be used independent of Drupal or WordPress is a good place to learn more about the general concepts. For a summary of basic usage, see [Composer's own documentation](https://getcomposer.org/doc/01-basic-usage.md).
+
+## Dependencies and Packages
+Composer encourages a mental model where code not written specifically for a given project is considered a "dependency". Dependencies for WordPress and Drupal include core, plugins, contrib modules, themes, and libraries. A single dependency, such as a theme, is referred to as a "package".
+
+By default, Composer can only see packages listed on [The PHP Package Repository](https://packagist.org/) which do not include Drupal or WordPress packages. Additional repositories must be configured for Composer to use packages not found in the default repository. Each framework provides it's own respective package repository so dependencies can be managed with Composer:
```
dustinleblanc: This isn't entirely accurate. Composer can pull packages from a huge swath of sources including VCS, local filesystem, etc. We may want to replace 'can only see' with something more akin to Composer's default is x.

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+
+- WordPress: [https://wpackagist.org](https://wpackagist.org)
+- Drupal 8: [https://packages.drupal.org/8](https://packages.drupal.org/8)
+- Drupal 7: [https://packages.drupal.org/7](https://packages.drupal.org/7)
+
+Site's created from Pantheon's example repositories already include the appropriate package repository within the `composer.json` file.
+
```
dustinleblanc: We might want to include an example right here that illustrates how to add a package repository to a composer.json.

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+## Nested Docroot
```
dustinleblanc: This section feels a little wierd to me. Composer can be used with a site with no nested docroot, and some of the arguments for nested docroots have nothing to do with composer (keeping project config out of the site root, including developer docs and build files outside of the web-accessible directories, etc). The push for both concepts kind of end up coming from the same people and started around the same timeframe, but they are actually separate. Its not a big deal, but I feel we might be over simplifying this and suggesting the two ideas as interrelated when they sort of aren't. I don't feel super strongly about this, what do you think @greg-1-anderson  and @stevector ?

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+The docroot is the directory from which your site is served. Without Composer, Pantheon defaults the docroot to the root directory of the site's codebase. In order to require core as a project dependency, a recommended best practice for Composer managed sites, it must be installed in a subdirectory such as `web`.
+
+This is possible on Pantheon by specifying `web_docroot: true` in `pantheon.yml` file. For details, see [Serving Sites from the Web Subdirectory](/docs/nested-docroot/).
+
```
dustinleblanc: An introduction to workflows and their divergence might be helpful here rather than jumping straight into CircleCI/Custom upstreams. I think the custom upstream workflow ends up having enough issues that we might want to avoid documenting it too. The two processes I actually see are using CI worflow, and using Local composer management workflow

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
+## Source and Artifact Workflow
+<div class="flex-panel-group">
+  <div class="flex-panel-item">
+    <div class="flex-panel-body">
+      <div class="flex-panel-title">
+        <h4 class="info" style="margin-top:10px;font-size:larger">GitHub</h3>
```
dustinleblanc: Maybe we should abstract this away from specific vendors? Github -> Hosted Source Code Repository? Not critical, but folks might skip this conceptual section because they use Bitbucket or Gitlab, or xyz awesome cool Git thing.

source/_docs/composer.md
dustinleblanc: The two workflows I'd document are 'Local compilation' and 'CI/Automated'. Those are really the two workflows I see people successfully use

source/_docs/composer.md
dustinleblanc: This sounds great to me! @alexfornuto I think I accidently deleted your comment while trying to move mine, sorry about that!

source/_docs/composer.md
```diff
@@ -1,175 +1,87 @@
...
-      ]
-    }
-  }
- ```
+Composer is used to fetch dependencies declared by the project as part of a CircleCI build step. This ensures that the final composed build results are installed on Pantheon:
+
+![Artifact Deployment](/source/docs/assets/images/artifact-deployment.png)
+
+### GitHub Pull Requests
+One advantage of managing code this way is that it keeps the change sets (differences) for pull requests as small as possible. If a pull request upgrades several dependencies, only the dependency metadata file will change; the actual code changes in the upgraded dependencies themselves are not shown.
```
dustinleblanc: Sweet! 👍

source/_docs/composer.md
alexfornuto: That's OK @dustinleblanc, I pulled it out of #docs:

> I couldn't possibly think of a simpler way to phrase that :+1:

source/_docs/composer.md
greg-1-anderson: I usually call these "template" repositories.

source/_docs/composer.md
greg-1-anderson: Right now, Pantheon maintains no upstreams with nested docroots. So, for most users, I would say that if you use a nested docroot, you should manage your site updates with Composer. If we think folks want nested docroots without Composer, then we should start maintaining some upstreams for them to use.

On the flip side, while it is technically possible to use Composer without using a nested docroot, doing this will often get the user into a situation where there are a lot of conflicts that need to be resolved. It seems much safer to me to just get everyone who wants a Composer-managed site to use a nested docroot.

We could say "SHOULD be installed in a subdirectory" instead of MUST, if you prefer, though.

source/_docs/composer.md
greg-1-anderson: We're working on support for other git repositories in the Build Tools plugin.

source/_docs/guides/build-tools/01-introduction.md
```diff
@@ -0,0 +1,120 @@
...
+      <div class="copy-snippet">
+        <button class="btn btn-default btn-clippy" data-clipboard-target="#mac-mt-auth">Copy</button>
+        <figure><pre id="mac-mt-auth"><code class="command bash" data-lang="bash">terminus auth:login --machine-token=&lsaquo;machine-token&rsaquo;</code></pre></figure>
+      </div>
+
+4. Install the [Terminus Composer Plugin](https://github.com/pantheon-systems/terminus-composer-plugin){.external}:
+
+    <div class="copy-snippet">
+      <button class="btn btn-default btn-clippy" data-clipboard-target="#composer-plugin">Copy</button>
+      <figure><pre id="composer-plugin"><code class="command bash" data-lang="bash">curl https://github.com/pantheon-systems/terminus-composer-plugin/archive/1.0.0.tar.gz -L | tar -C ~/.terminus/plugins -xvz</code></pre></figure>
```
greg-1-anderson: Installing with `curl | tar` is not a good idea. This method contains the version number of the tool to download and install in the URL. Doing it this way would require us to update the documentation every time a new release is made; otherwise, folks would get the old version. There is still a risk that the download instructions could be copied e.g. into user notes, and external blog post, etc., and there is no way to clean up these stale references. We should instead document installing with `composer create-project`, which will always install the most recent version.

source/_docs/guides/build-tools/02-create-project.md
```diff
@@ -0,0 +1,99 @@
...
+3. Create a new project (replace `pantheon-d8-composer-project` with the name of your new site):
+
+    <div class="copy-snippet">
+      <button class="btn btn-default btn-clippy" data-clipboard-target="#d8-create">Copy</button>
+      <figure><pre id="d8-create"><code class="command bash" data-lang="bash">terminus build:project:create d8 pantheon-d8-composer-project</code></pre></figure>
+    </div>
+
+    <div class="alert alert-info">
+      <h4 class="info">Note</h4>
+      <p markdown="1">Pantheon also maintains Composer based examples for [WordPress](https://github.com/pantheon-systems/example-wordpress-composer){.external} and [Drupal 7](https://github.com/pantheon-systems/example-drops-7-composer){.external} that are currently in alpha, requiring `--stability=alpha` in the command line options. While this guide demonstrates Drupal 8, the same workflow can be achieved on all frameworks.</p>
```
greg-1-anderson: We should really tag 1.0.0 stable for WordPress and Drupal 7 before you publish, so that this note is not necessary.

source/_docs/guides/build-tools/02-create-project.md
```diff
@@ -0,0 +1,99 @@
...
+      [Composer\Downloader\TransportException]
+      Content-Length mismatch
+
+
+      create-project [-s|--stability STABILITY] [--prefer-source] [--prefer-dist] [--repository REPOSITORY] [--repository-url REPOSITORY-URL] [--dev] [--no-dev] [--no-custom-installers] [--no-scripts] [--no-progress] [--no-secure-http] [--keep-vcs] [--no-install] [--ignore-platform-reqs] [--] [<package>] [<directory>] [<version>]
+
+      [error]  Command `composer create-project --working-dir=/private/var/folders/lp/7_1gh83s5mn9lwfjvqqlf1lm0000gn/T/local-sitevPumRP pantheon-systems/example-wordpress-composer pantheon-wp-composer-project -n --stability dev` failed with exit code 1
+      ```
+
+     This indicates a network level issue related to the security settings of your modem. We recommend contacting your Internet Service Provider (ISP) and asking them to change the modem's security mode to TKIP/AES. Additionally, you can help mitigate network issues by using a non-standard channel with less activity/noise.
```
greg-1-anderson: Lots of network issues, including simply a slow connection can cause this problem. I think that the solution presented above would only help a few people.

source/_docs/guides/build-tools/02-create-project.md
```diff
@@ -0,0 +1,99 @@
...
+    </div>
+
+    <div class="alert alert-info">
+      <h4 class="info">Note</h4>
+      <p markdown="1">Pantheon also maintains Composer based examples for [WordPress](https://github.com/pantheon-systems/example-wordpress-composer){.external} and [Drupal 7](https://github.com/pantheon-systems/example-drops-7-composer){.external} that are currently in alpha, requiring `--stability=alpha` in the command line options. While this guide demonstrates Drupal 8, the same workflow can be achieved on all frameworks.</p>
+    </div>
+
+    Provide additional information as prompted, such as:
+
+    ![Create Project Prompts](/source/docs/assets/images/pr-workflow/build-env-create-project-prompts.png)
```
greg-1-anderson: Not sure if you do or do not want to document this, but the admin password and team can also be placed in `~/.terminus/config.yml`, in which case those values will be used thereafter for every site created, w/out prompting.
```
command:
  build:
    project:
      create:
        options:
          admin-password: secret-password
          team: Anita Agency
```

source/_docs/guides/build-tools/03-create-pr.md
```diff
@@ -0,0 +1,88 @@
...
+
+    ![Create slogan branch](/source/docs/assets/images/pr-workflow/create-slogan-branch.png)
+
+    <div class="panel panel-drop panel-guide" id="accordion">
+      <div class="panel-heading panel-drop-heading">
+         <a class="accordion-toggle panel-drop-title collapsed" data-toggle="collapse" data-parent="#accordion" data-proofer-ignore data-target="#understand-branch-names"><h3 class="panel-title panel-drop-title" style="cursor:pointer;"><span style="line-height:.9" class="glyphicons glyphicons-lightbulb"></span> Branch Naming Conventions</h3></a>
+       </div>
+       <div id="understand-branch-names" class="collapse">
+         <div class="panel-inner" markdown="1">
+         Be sure to use unique branch names for each feature branch; the Multidev environment created on Pantheon is named after your branch. Since there is a limit to the number of characters that may be used in a Pantheon Multidev name, your environments may conflict if you always use the branch name that GitHub suggests. For details, see [Multidev](/docs/multidev/#what-are-the-naming-conventions-for-branches).
```
greg-1-anderson: This has caused problems for some folks; we are moving towards implementing a new scheme where the multidev environment is named after the pull request number.

source/_docs/guides/build-tools/03-create-pr.md
```diff
@@ -0,0 +1,88 @@
...
+
+    ![Slogan Pull Request](/source/docs/assets/images/pr-workflow/slogan-pull-request.png)
+
+    <div class="panel panel-drop panel-guide" id="accordion">
+      <div class="panel-heading panel-drop-heading">
+         <a class="accordion-toggle panel-drop-title collapsed" data-toggle="collapse" data-parent="#accordion" data-proofer-ignore data-target="#understand-builds"><h3 class="panel-title panel-drop-title" style="cursor:pointer;"><span style="line-height:.9" class="glyphicons glyphicons-watch"></span> Builds</h3></a>
+       </div>
+       <div id="understand-builds" class="collapse">
+         <div class="panel-inner" markdown="1">
+         As soon as you commit changes to a feature branch, CircleCI builds a new Multidev environment on Pantheon to preview the change. Once the Multidev environment has been created, the build script will add a comment to the commit with links to the Multidev environment of the Pantheon Site Dashboard and a button to visit the Mutidev site URL (e.g., `pr-slogan-my-pantheon-project.pantheonsite.io`). The Pull Request page conveniently shows the messages from each commit on the branch:
```
greg-1-anderson: If we make the aforementioned change, then CircleCI will not build the new multidev environment until the pull request is created.

source/_docs/guides/build-tools/07-custom-theme.md
```diff
@@ -0,0 +1,154 @@
...
+          * Type `yes` when asked to confirm generation of the theme
+
+        ![Drupal console generate theme](/source/docs/assets/images/pr-workflow/drupal-console-generate-theme.png)
+
+        Once you do this, the files for your new theme will be written to the directory `code/web/themes/custom/amazing_theme` on the Pantheon Multidev environment. If you gave your theme a different name, replace `amazing_theme` with the appropriate name for your theme.
+        </div>
+        </div>
+      </div>
+
+4. Open a second terminal window and run the following command to open an SFTP session from the command line:
```
greg-1-anderson: Maybe it would be better to have them commit their generated theme from the dashboard? We could advise adding "[ci skip]" in the commit message to [avoid running the tests on Circle](https://circleci.com/docs/1.0/skip-a-build/).

greg-1-anderson:
Great improvements, looking really good.

source/_docs/guides/build-tools/07-custom-theme.md
rachelwhitton: Ooh, yeah I think doing a commit from the dashboard in this part would be smoother. I'll do a pass now to give it a shot. I'm confused on why we'd want to skip ci though, wouldn't we want the dashboard commit pushed to the GitHub repo @greg-1-anderson ?

pandocs: Just realizing that the push back to GitHub is a Quicksilver hook, not a build step on CirlceCI. 🌠 💥

rachelwhitton: Just realizing that the push back to GitHub is a Quicksilver hook, not a build step on CirlceCI. 🌠 💥

greg-1-anderson: Correct, the pushback is done by Quicksilver. If you let the build step run, then it will overwrite the current multidev with the build results, presumably with the same code. This might also reset any configuration changes the user may have made without exporting.

source/_docs/guides/build-tools/02-create-project.md
rachelwhitton: I didn't get this worked in on this PR but I believe there's a related issue open to add it to the terminus manual

alexfornuto:
👍  :shipit:

alexfornuto:
NO WAIT

PR #2861, opened 2017-09-06
Additional info for HSTS configuration

### LINKED ISSUE #2856: Update HSTS Doc
Re: <https://pantheon.io/docs/hsts/>
## Issue Description
(From the SME)
The suggested expiry times vary as if it depends on the module/plugin, and seems to imply that is how it is, but that is a configuration the user should understand/set. also there is no mention of `includeSubDomains` and `preload` which are important parts of the configuration too.

## Suggested Resolution
Address the above issues, and define one year and 6-month expiry numbers:

One year expiry:
`Strict-Transport-Security: max-age=31536000;`

Six month expiry with include subdomains and add to browser preload list:
`Strict-Transport-Security: max-age=15984000; includeSubDomains; preload`

peter-pantheon:
Closes #2856

## Effect
- Additional resource(s) link(s)
- Additional info on using HSTS headers

## Remaining Work
- [x] Proof for style, shorten

### DIFF
--- a/source/_docs/hsts.md
+++ b/source/_docs/hsts.md
@@ -5,10 +5,33 @@ tags: [security]
 ---
 After you have required HTTPS for all pages by adding the [necessary redirect](/docs/domains/#redirect-to-https-and-the-primary-domain), set the HTTP Strict Transport Security (HSTS) header to standardize all client connections on HTTPS and prevent use of HTTP.
 
-## Configure HSTS Headers
+Not only does this header help you get an A+ SSL rating from [SSL Labs](https://www.ssllabs.com/ssltest/){.external}, it will help protect your website against protocol downgrade attacks and cookie hijacking.
+
+Note: Before adding HSTS to your site, you should review and understand the configuration options that are available. A HSTS header that is configured incompletely or not securely enough diminishes the security protection that HSTS provides.
+
+## HSTS Header Configuration Attributes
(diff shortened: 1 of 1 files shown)


source/_docs/hsts.md
```diff
@@ -51,3 +71,31 @@ Once installed and configured, the following header will be sent in responses:
 ```http
 strict-transport-security: max-age=15552000
 ```
+</div>
+
+</div>
+
+
+## Configure Attributes
```
rachelwhitton: Suggest adding instructions to this section for how to configure attributes. If step by step can't be provided, maybe some general advice for how to do this?

alexfornuto: Good idea, I'll see if @peter-pantheon can look into this during his next doc block.

peter-pantheon: @rachelwhitton @alexfornuto 

User's can inject the header via a (configurable) WordPress plugin, (configurable)  Drupal module, code or by hard coding the header into a template file. So I think giving general advisement on configuration options is acceptable here.  

I see there is a header called "Configure Atrributes" and info below it and that looks good to me. What more specifics would be desired here? 

There are some details here that may require further offline discussion too,

rachelwhitton:
+1 Looks good, thanks @peter-pantheon and @alexfornuto :shipit:

alexfornuto:
Thanks @peter-pantheon!

PR #2875, opened 2017-09-11
Terminus Commands: Note help option

### LINKED ISSUE #2844: Terminus Manual Doc Update
Re: [Terminus Manual](https://pantheon.io/docs/terminus/commands/)

Priority (Low):

## Issue Description:
User asks where to find in the docs possible "--fields" for every command

## Suggested Resolution 
Create a table of Fields possible for all commands like workflow, org, etc
Also need to be clear if the default, that not putting "--fields" means we are showing ALL fields already rather than restricting the list

rachelwhitton:
Closes #2844 

## Effect
- Add a note to the top of the Command Reference page within the Terminus Manual directing users to run commands with the `--help` option for additional information like available fields. 

cc @eabquina @kimby77

### DIFF
--- a/source/_docs/terminus/commands.md
+++ b/source/_docs/terminus/commands.md
@@ -9,6 +9,11 @@ previousurl: terminus/examples/
 permalink: docs/terminus/:basename/
 image: terminus-thumbLarge
 ---
+<div class="alert alert-info" markdown="1">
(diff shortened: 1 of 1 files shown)


kimby77:
👍  great solution.

PR #2876, opened 2017-09-11
Cloudflare Updates: Ignore A/AAAA and provide CAA instructions

### LINKED ISSUE #2834: Cloudflare Domain Configuration Doc Update
Re: [Cloudflare Domain Configuration](https://pantheon.io/docs/cloudflare/)
Re: Case 88378

## Issue Description:
This page does not discuss what to do with the recommended AAAA and A values provided in the Dashboard for bare custom domains, it simply says that CNAME flattening is better since it's not tied to a specific IP

## Suggested Resolution 
Verify optimal configuration from @ari-gold -- is a CNAME best? Should AAAA and A required values be ignored? The doc needs explicit instructions for these records since they're shown in the Dashboard.

Maybe the Dashboard should recommend CNAME flattening instead of AAAA/A for Cloudflare?

rachelwhitton:
Closes #2834 

## Effect
- Document CAA records as optional in Cloudflare doc
- Add content to suggest users ignore A/AAAA DNS recommendations when using CNAME flattening 

@ari-gold


alexfornuto:
👍

source/_docs/dns-providers/cloudflare.md
```diff
@@ -55,6 +54,31 @@ You can configure Cloudflare's CDN as an additional layer on Pantheon's Global C
...
+
+A **CAA Record** specifies which certificate authority (**CA**) can issue HTTPS certificates for a domain.
+
+1. Select **DNS** from the Cloudflare menu bar.
+2. Select **CAA** from the dropdown menu.
+3. Enter the bare domain (`example.com`) in the **Name** field then click to configure the record value:
+
+  ![caa click to configure](/source/docs/assets/images/cf-caa.png)
+
+4. Select **Allow wildcards and specific hostnames** for the record's tag and enter  `letsencrypt.org` in the value:
```
ari-gold: (optional) If you'd like to authorize other Certificate Authorities to issue certificates for your domain you can use SSLMate’s CAA Record Generator  (https://sslmate.com/caa/) to help generate a set of CAA records.

alexfornuto: @ari-gold What would be the purpose, if customers can only use the certificates we provide on our platform?

ari-gold: They could, for example, have a subdomain serving traffic from somewhere else other than Pantheon.

rachelwhitton: Does our recommendation to "Allow wildcards and specific hostnames" conflict with someone's ability to create records for subdomains serving traffic from somewhere else other than Pantheon?  Unless it does, I vote we leave it out to keep things simpler. 

If we add it in, folks could think they need to do something there when they don't.

ari-gold: We can leave it out, but then let's find another home to generally explain CAA records

PR #2896, opened 2017-09-15
New Guide: Front End Performance

rachelwhitton:
## Effect
- Add new Front End Performance guide for how to ace an online speed test with Pantheon's Global CDN. 
  - Promote guide on the performance topic page and guide landing page
- Add Terminus Manual to the guides landing page

## Remaining Work
- [x] Finish copy review 
   - [x] Work in suggestions from joshkoenig 
- [x] Final round of peer reviews 
- [x] Replace screenshot of a site acing it (deploy autoptimize on scalewp.io)


source/_docs/guides/front-end-performance.md
```diff
@@ -0,0 +1,271 @@
...
+A redirect will add at least one extra HTTP request-response cycle. As a result, eliminating extraneous redirects can make your website more snappy. Despite your best efforts it still may be necessary to include the occasional [redirect to a primary domain](/docs/guides/launch/redirects/) using HTTPS with or without the WWW.
+
+Other considerations:
+
+- Avoid mobile-specific subdomains and use responsive web design techniques.
+- A DNS service provider such as [Cloudflare](https://support.cloudflare.com/hc/en-us/articles/200170536-How-do-I-redirect-all-visitors-to-HTTPS-SSL-){.external} may allow speedier redirects in some circumstances, but it’s still faster not to redirect at all.
+- Avoid several chained redirects that make small changes such as redirecting to HTTPS, adding or removing WWW, or adding a trailing slash. Instead, [redirect to a primary domain](/docs/guides/launch/redirects/) that has all of these standardized.
+- Pantheon doesn’t read changes to the .htaccess file or support NGINX customization, so redirections via those methods will not work.
+
+## Minify Resources (HTML, CSS and JS)
```
stevector: @rachelwhitton @tessak22 I think `Remove Render-Blocking JavaScript and CSS` and `Minify Resources (HTML, CSS and JS)` should be next each other since they cover very similar material.

And I think `Minify` should come before `Remove Render-Blocking` because it is an easier to understand and more achievable change.

stevector: @rachelwhitton, do you have time today to discuss this idea? I've started drafting a combined version at the bottom of https://docs.google.com/document/d/18N4hxwuopninlUHXtuTzxnLPE1qdUdfTQXpK2QQZWT0/edit?ts=59c129e9#

rachelwhitton: absolutely! @stevector let me know when and we can sync up in slack or hop on a call. either way works for me. 

+1 on the proposed content organization changes, makes sense to group related stuff together. should help us avoid overlap as well.

stevector: Cool, I sent an invite for 15 minutes from now.

rachelwhitton: Sure thing! On it now

joshkoenig:
Sorry to miss the review meeting. This is off to a _great_ start. Some initial suggestions:

 1. We should link to the test we're referencing rather than the thinkwithgoogle homepage: https://testmysite.thinkwithgoogle.com/ 
 2. Also a screenshot of a site Acing it would be great!
 3. We should contextualize and introduce "reduce server response time" better. There's content that can be pulled from David and I's webinar for this. 
 4. Do we want to introduce TTFB as a key concept here?
 5. For the Global CDN, I'd emphasize "in your local network neighborhood" more than  "don't go across the ocean" as the benefit. It's not great if you're going across the continent either, or even halfway. Sub-second-pageloads require a quick local response.
  6. "Use as few requests as possible" - this may not be the right advice in the era of HTTP/2. I think we need to make a recommendation here, not add a disclaimer.
 7. Should we be talking about `script async/defer` somewhere in here? (http://www.growingwiththeweb.com/2014/02/async-vs-defer-attributes.html)

I like how this is shaping up! I can try to make some more substantive contributions next week.

rachelwhitton:
Thanks for the review @joshkoenig! I think I got all your suggestions in.. except for: 
> 6. "Use as few requests as possible" - this may not be the right advice in the era of HTTP/2. I think we need to make a recommendation here, not add a disclaimer.

Planning to work on that part with @stevector later this afternoon

stevector:
Thanks for the feedback @joshkoenig! Regarding 

> "Use as few requests as possible" - this may not be the right advice in the era of HTTP/2. I think we need to make a recommendation here, not add a disclaimer.

@davidstrauss and I talked a lot about this on Tuesday. We tested across a few different sites. David's summary in Slack:

> Yep, in general, h2 is not sufficient to make extra requests "free."
> They're much cheaper than before, but there's still a price.

It's really clear on a clean install of D8 where leaving aggregation off means that literally 100 separate CSS files will be loaded.

So the recommendation is still for turning aggregation off. I'll work on the language here. I do want to acknowledge that h2 has the potential to change the default recommendation in the future but does not do so yet.

stevector:
@rachelwhitton I just pushed a rewrite of the HTTP/2 note along with some other copy edits. Most significantly https://github.com/pantheon-systems/documentation/pull/2896/commits/c5d384aee88f8b0ea110b4ed9db6991ee7aedc84

source/_docs/guides/front-end-performance.md
```diff
@@ -0,0 +1,457 @@
...
+- Reduce Server Response Time
+- Compress Images
+- Deliver Efficient CSS and JavaScript
+- Avoid Redirects
+- Leverage Browser Caching
+
+![Acing Google's speed test](/source/docs/assets/images/guides/front-end-performance/ace-it.png)
+
+## Quick Response Time
+<dt>Waiting</dt>
```
joshkoenig: This didn't read for me. Why is a headline that says "Quick Response Time" followed by a subhead that says "waiting"? Feels like there's a missing intro paragraph perhaps?

source/_docs/guides/front-end-performance.md
```diff
@@ -0,0 +1,457 @@
...
+
+  ```php
+  /**
+  * This is an example of accidentally bypassing cache when rendering a response.
+  * Don't use this without proper conditionals wrapped around the header function.
+  **/
+  add_action( 'send_headers', 'add_header_maxage' );
+  function add_header_maxage() {
+    // some logic that accidentally invalidates full-page cache
+      header('Cache Control: max-age=3600');
```
joshkoenig: Shouldn't this read `max-age=0`?

rachelwhitton: Whoops, yes!

source/_docs/guides/front-end-performance.md
```diff
@@ -0,0 +1,457 @@
...
+  3. Open the **Network** tab, then refresh the page. This allows you to review all of the HTTP requests necessary to display your page.
+  4. Select the first row, which is a request for the initial HTML file.
+  5. Look for the **Age** header. Any number greater than zero indicates that this response was served to the browser from cache.
+  </div>
+</div>
+
+### Debug Common Cache Busters
+#### Cookies
+Cookies are included in the response headers we examined previously. They can include sessions for authenticated traffic to logged in users, which can invalidate cache. For WordPress, it's common for plugins to add their own cookies in such a way that breaks full-page caching.
+
```
joshkoenig: We should probably list cookies that can break caching here, plus how to see if your site is sending them unnecessarily?

source/_docs/guides/front-end-performance.md
```diff
@@ -0,0 +1,457 @@
...
+If you want to make the number of files as low as possible, try [Advanced Aggregation](https://www.drupal.org/project/advagg){.external} module. [It can help take fine grain control over how files are combined](http://www.mediacurrent.com/blog/better-css-js-aggregation-advanced-aggregation){.external}. In the past there have been bugs when using Advanced Aggregation on Pantheon. To our knowledge, those are all resolved.
+  </div>
+</div>
+
+<div markdown="1" class="alert alert-info">
+  <h4 class="info">Note</h4>
+The Pantheon Global CDN includes HTTP/2. One of the benefits of HTTP/2 is that it allows multiple files to be downloaded simultaneously.<br/><br/>
+The Drupal community has long speculated whether HTTP/2 will make CSS and JavaScript aggregation irrelevant. (In theory, lots of little files, each individually cacheable, and downloaded en masse over one HTTP/2 connection would be more performant than aggregated files.)<br/><br/>
```
joshkoenig: This is a bigger speculation than the Drupal community. I'd say "the web community"

joshkoenig: Also, the key tradeoff is whether you have one aggregated file (which can only be beaten by one big inline CSS statement) vs different pages serving different aggregates. But I digress. We can revisit this if/when we get into HTTP/2 Push support. ;)

source/_docs/guides/front-end-performance.md
rachelwhitton: Added this to explore opening sections with a definitions list, and I wasn't sure how I felt about it. Main goal is to define TTFB and TTFP for the next paragraph. 

I'll rewrite into paragraph form, it does feel a bit weird.

source/_docs/guides/front-end-performance.md
```diff
@@ -0,0 +1,457 @@
...
+#### Upgrade to PHP7
+If you haven't done so already, [make the switch to PHP7](/docs/php-versions/). Upgrading your site's PHP version will improve the security, performance, and supportability of your site.
+
+See our blog post for an example of [62% performance gains after upgrading](https://pantheon.io/blog/php-7-now-available-all-sites-pantheon){.external}.
+
+#### Enable Redis Object Caching
+Sites loading a lot of content can benefit from an object cache like Redis. For details, see [Installing Redis on Drupal](/docs/drupal-redis/) and [Installing Redis on WordPress](/docs/wordpress-redis/).
+
+#### Monitor Performance with New Relic
+Use [New Relic](/docs/new-relic/) to monitoring performance if your site isn't sending uncached pages fast enough. This service is provided for free on every Pantheon site, but you must activate it.
```
rachelwhitton: Typo "Use New Relic to *monitoring* performance.."

source/_docs/guides/front-end-performance.md
rachelwhitton: @stevector @joshkoenig I added the cookie patters, but I need some help filling in the content on how to see if your site is sending them unnecessarily.

stevector: I'm adding instructions

stevector: https://github.com/pantheon-systems/documentation/pull/2896/commits/9c02dc26e26e770fb26cffb4fb75a4620cd84658

stevector:
> deploy autoptimize on scalewp.io

@rachelwhitton who is doing that task?

rachelwhitton:
@stevector I have it staged on a multidev, but wanted someone else to take a look at it before deploying to live. Do you have bandwidth @stevector?

source/_docs/guides/front-end-performance.md
```diff
@@ -0,0 +1,481 @@
...
+  1. From the WordPress dashboard, click **Settings** > **Pantheon Page Cache**.
+  2. Review the Time to Live, which translates to `max-age`.
+  3. We recommend setting **Default Time to Live (TTL)** to a higher value, like 86400 seconds (one day):
+
+  ![Pantheon Cache Plugin](/source/docs/assets/images/guides/front-end-performance/pantheon-page-cache.png)
+  </div>
+
+  <!-- 2nd pane content -->
+  <div role="tabpanel" class="tab-pane" id="drupal-config-get" markdown="1">
+In Drupal it is very easy to turn off page caching and forget to turn it back on. Use [Terminus](/docs/terminus/) to check and edit caching configurations:
```
rachelwhitton: FYI 

This section describes how to check cache configurations with Terminus because I found that method cool, and I wanted to avoid repeating content from the *Use as Few Requests as Possible* section. Is this okay? Should we change it to the admin ui instead?

thoughts @davidneedham @stevector ?

stevector: I think the readers are more likely to understand what is happening if we refer to the admin UI instead of a terminus/drush combo.

rachelwhitton: Agreed. I'll get it fixed up. Just to confirm, we're walking through the configs at `/admin/config/development/performance` yeah? Specifically the cache part: 
screen shot 2017-09-22 at 11 28 25 am

What should we recommend they set it to @stevector?

stevector: Yes, that's it. It's hard to make one standard recommendation here. The Drupal 8 default is 10 minutes. But if Pantheon Advanced Page Cache is being used well, then you could go much, much higher (hours, days, weeks). But I don't think we have enough guidance written yet on validating Pantheon Advanced Page Cache is being used well on a site to make a blanket recommendation of some high number.

How about something like

> The Drupal 8 default setting is 10 minutes. You can set much higher cache max ages when using the Pantheon Advanced Page Cache Module to clear specific pages when your underlying data is updated.

source/_docs/guides/front-end-performance.md
```diff
@@ -0,0 +1,475 @@
...
+
+<div class="enablement">
+ <a href="https://pantheon.io/agencies/learn-pantheon?docs"><h4 class="info">Agency DevOps Training</h4></a>
+ <p>Learn industry best practices for caching, how to take advantage of them on the platform, and troubleshooting common issues with help from the experts at Pantheon.</p>
+</div>
+
+
+There are many different ways to measure page speed and performance. This guide will be tuning for the metrics set forth by [Google's mobile speed test](https://testmysite.thinkwithgoogle.com/){.external}:
+
+- Reduce Server Response Time
```
stevector: @rachelwhitton, did we want this list to line up exactly with the major headings in the sidebar? They no longer do.

source/_docs/guides/front-end-performance.md
```diff
@@ -0,0 +1,475 @@
...
+
+  <!-- 2nd Tab Nav -->
+  <li id="tab-2-id" role="presentation"><a href="#drupal-config-get" aria-controls="drupal-config-get" role="tab" data-toggle="tab">Drupal</a></li>
+</ul>
+
+<!-- Tab panes -->
+<div class="tab-content">
+  <!-- Active pane content -->
+  <div role="tabpanel" class="tab-pane active" id="wp-config-get" markdown="1">
+  The Pantheon Page Cache plugin is already included by our upstream as a Must-Use plugin. Check the plugin settings to make sure you're setting the desired TTL and max-age:
```
stevector: > the desired TTL and max-age

These are not different things, just different words for the same thing. Let's change it to "the desired TTL".

source/_docs/guides/front-end-performance.md
rachelwhitton: Hmm, I think this list should use the categories from Google's results. Otherwise I prefer we ditch the list entirely because matching it to the major headings is redundant to the fixed TOC we have on the right.

stevector: k, let's leave it as is.

stevector:
@rachelwhitton, shall we merge once we get one more green pass from Circle?

rachelwhitton:
Yas!! 🚢  :shipit: 🚀 

+1

Thanks for all your help @stevector!

PR #2897, opened 2017-09-17
upgrade prompt

allenfear:
Closes #
N/A

## Effect
Adds prompt for folks to upgrade and a warning that legacy version is unsupported

## Remaining Work
N/A


rachelwhitton:
+1

PR #2903, opened 2017-09-19
remove extra argument to git remote

ilovelamp:
Re: https://pantheon.io/docs/guides/collaborative-development/#add-the-pantheon-site-as-a-git-remote

the `git remote add` command has a site name/label of `pantheon-new-site-import` tacked on the end, which is invalid.  This PR removes the trailing parameter.

> git remote add pantheon ssh://codeserver.dev.{site-id}@codeserver.dev.{site-id}.drush.in:2222/~/repository.git **pantheon-new-site-import**

I also clarified the use of a remote name "pantheon" in the instructions.  It was called an alias, but git docs simply call it a "remote name."

### DIFF
--- a/source/_docs/guides/collaborative-development.md
+++ b/source/_docs/guides/collaborative-development.md
@@ -146,7 +146,7 @@ Automatic merge went well; stopped before committing as requested
 
 ### Add the Pantheon Site as a Git Remote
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
Noice!

PR #2924, opened 2017-09-26
bash on ubuntu on windows is too problematic

mcdwayne:
We should not be recommending windows users use Terminus directly

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed


alexfornuto:
@mcdwayne can you provide some more context here?

alexfornuto:
replaced by #2954

PR #2925, opened 2017-09-26
Changing the Windows reccomendation

mcdwayne:
Only if they already have found success should they try this, been working with too many people who tried it due to Terminus and are failing and we don't know how to help them.

Closes #

## Effect
PR includes the following changes:
-
-
-

## Remaining Work
- [ ] List any outstanding todos
- [ ] If needed

PR #2954, opened 2017-10-03
Terminus windows

alexfornuto:
Replaces #2924 #2925 

## Effect
- Adjusts Terminus documentation to reference proper Windows implementation
-
-

## Remaining Work
- [x] Replace Ubuntu Bash with Git for Windows as suggested environment.
- [x] Copy Review


rachelwhitton:
+1 but would like confirmation that switching from 7 and 10 to 7-10 is safe

PR #2955, opened 2017-10-03
Update unsupported-modules-plugins, Contact Form 7

albertcausing:
Both SERVER_NAME and SERVER_PORT are needed to fix the issue

## Effect

Fix contact form issue on confirmation redirect page having an internal port instead of 80 or 443.

### DIFF
--- a/source/_docs/unsupported-modules-plugins.md
+++ b/source/_docs/unsupported-modules-plugins.md
@@ -291,7 +291,7 @@ For additional details, see the [Autoptimize FAQ](https://wordpress.org/plugins/
 <hr>
 
 ### [Contact Form 7](https://wordpress.org/plugins/contact-form-7/)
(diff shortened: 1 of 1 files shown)


alexfornuto:
I'm testing this solution.

albertcausing:
Reference @chat: 12121911591

alexfornuto:
I tested this plugin (without any tie-ins to other mail applications) and found it to work fine, even after several appserver migrations. The only caveat was setting the "From" email to match my live environment's domain.

Can we add some clarification regarding what sort of configurations require these fixes?

rachelwhitton:
In general, any plugin known to rely on static values (such as appserver binding id) should recommend configuration to correct. 

IMO it's less important to know each and every instance in which this problem surfaces for this particular plugin, and more important to have the recommended config in the docs for when the issue is present during support interactions. 

+1 to ship as is

PR #2961, opened 2017-10-04
Pre-Provisioning HTTPS

### LINKED ISSUE #2945: Pre Provisioning of HTTPS
Re: [Launch Essentials](https://pantheon.io/docs/guides/launch/configure-dns/)
[HTTPS](/docs/https)

Priority (Medium):

## Issue Description:
Multiple requests have been mentioned about how to "pre-provision" for cases both on existing HTTPS, NO HTTPS, or first-timers. 

Need a clear concept, illustration on how pre-provisioning works and what's only possible.

## Suggested Resolution 

Provide section that highlight when should pre-provisioning work, and what are the scenarios that these will actually work on its own.

Also, provide clarity if it is actually possible to pre-provision (even for non HTTPS sites)
Provide troubleshooting points why a domain is not pre-provisioned.

alexfornuto:
Closes #2945 

## Effect
- adds note with pre-provision no-no's

## Remaining Work
- [ ] @eabquina and @newtoid to add other information they feel needs to be included.
- [ ] Copy review.

### DIFF
--- a/source/_docs/https.md
+++ b/source/_docs/https.md
@@ -103,6 +103,17 @@ Yes! As long as your are following the Dashboard DNS recommendations before star
 
 Existing sites that are live over HTTPS which are not already hosted on Pantheon can [pre-provision HTTPS](/docs/guides/launch/configure-dns/) to avoid interruption. If you are unable to prove ownership as described, we recommend a maintenance window.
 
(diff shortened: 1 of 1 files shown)


eabquina:
@alexfornuto thanks! this is good and ok for me

PR #2974, opened 2017-10-09
Cloudflare for DNS only: redirects

thomas-thackery:
"NOTE: Traffic must be passing through Cloudflare for Page Rules to work, if you are using Cloudflare for DNS only then Page Rules will not be active" (https://support.cloudflare.com/hc/en-us/articles/200172286-How-do-I-perform-URL-forwarding-or-redirects-with-Cloudflare-).

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed


alexfornuto:
@rachelwhitton tech-reviewed it on the spot!

PR #2994, opened 2017-10-13
Update modules.md

ttrowell:
move comment on D8 to avoid confusion that the rest of the article pertains to D8

Closes #

## Effect
relocates a note


alexfornuto:
👍

PR #3003, opened 2017-10-17
Reset Dev Env to Live only affects code

thomas-thackery:
Clarification that the Reset Dev Env to Live does not clone down Live's db or files.

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed


rachelwhitton:
+1

PR #3004, opened 2017-10-18
Update http-to-https.md

lauren-kelly:
Closes n/a

## Effect
- add the option to update to HTTPS via dashboard, for non-terminus users

## Remaining Work
not sure if you want to re-word or stylize this better.


source/_docs/http-to-https.md
```diff
@@ -60,9 +60,15 @@ Use [Terminus](/docs/terminus) to run `wp search-replace` to converts URLs from
...
 terminus remote:wp <site>.<env> -- search-replace 'http://www.example.com' 'https://www.example.com' --all-tables --verbose
 ```
+##### Via Dashboard
+Another option that will work for most WordPress sites is to clone the Live database down to Test - and then clone it back to Live, ensuring you select the desired URL conversion and to Convert URLs' Protocol to HTTPS.
+
 #### Drupal
 Drupal 7 sites can use [Drush Search and Replace (sar)](https://www.drupal.org/project/sar) by [adding custom Drush command](/docs/drush/#add-custom-drush-commands). Drupal 8 sites can use the [Entity API](https://www.drupal.org/docs/8/api/entity-api/introduction-to-entity-api-in-drupal-8) to fetch data from entities in the database that may include insecure references.
 
+#### Via Dashboard
+Another option that will work for most sites is to clone the Live database down to Test - and then clone it back to Live, ensuring you select the desired
```
rachelwhitton: @lauren-kelly looks like this bit wasn't quite finished. would this be accurate? 
> Another option that will work for most sites is to clone the Live database down to Test - and then clone it back to Live, ensuring you select the desired values for URL conversion.

rachelwhitton: Ah, on second glance. Maybe this section was an accident? Seems duplicate to the first on lines 63-65

rachelwhitton:
One question, otherwise LGTM +1

source/_docs/http-to-https.md
```diff
@@ -60,8 +60,33 @@ Use [Terminus](/docs/terminus) to run `wp search-replace` to converts URLs from
...
+    ![Cloning the Live Database to Test](/source/docs/assets/images/dashboard/clone-live-to-test.png)
+
+    <div class="alert alert-danger" role="alert">
+      <h4 class="info">Warning</h4>
+      <p markdown="1">Be *sure* that you are cloning in the right direction. If you accidentally replace your Live environment's database, you can lose data.</p>
+    </div>
+
+2. Now, from the **Live** environment, clone your database back from Test, making sure to select "https" under **Convert URLs' Protocol to:**:
+
+    ![Cloning the Test database to Live, while converting URLs](/source/docs/assets/images/dashboard/clone-test-to-live.png)
```
rachelwhitton: I think the UI is different based on framework. I don't think Drupal sites see the same options when importing db

source/_docs/http-to-https.md
```diff
@@ -60,8 +60,33 @@ Use [Terminus](/docs/terminus) to run `wp search-replace` to converts URLs from
...
 terminus remote:wp <site>.<env> -- search-replace 'http://www.example.com' 'https://www.example.com' --all-tables --verbose
 ```
+
 #### Drupal
-Drupal 7 sites can use [Drush Search and Replace (sar)](https://www.drupal.org/project/sar) by [adding custom Drush command](/docs/drush/#add-custom-drush-commands). Drupal 8 sites can use the [Entity API](https://www.drupal.org/docs/8/api/entity-api/introduction-to-entity-api-in-drupal-8) to fetch data from entities in the database that may include insecure references.
+Drupal 7 sites can use [Drush Search and Replace (sar)](https://www.drupal.org/project/sar) by [adding custom Drush command](/docs/drush/#add-custom-drush-commands).
+
+Drupal 8 sites can use the [Entity API](https://www.drupal.org/docs/8/api/entity-api/introduction-to-entity-api-in-drupal-8) to fetch data from entities in the database that may include insecure references.
+
+**Via Dashboard**
```
rachelwhitton: I think this entire section should go above the Drupal header, as an alternative for WP users to search and replace via Pantheon Site Dashboard GUI

rachelwhitton: Confirmed. Drupal sites won't have the ability to clean up DB in the Site Dashboard on Pantheon. That's WP specific 
screen shot 2017-10-20 at 1 57 50 pm

alexfornuto:
@rachelwhitton That dovetails nicely with your comment on the original. It looked to me like @lauren-kelly had it in both locations because you could do this for WP _or_ Drupal. I moved the section so it would be distro-agnostic.

rachelwhitton:
+1

PR #3005, opened 2017-10-18
Fix formatting so the list properly renders

danielbachhuber:


danielbachhuber:
> Tested locally to confirm, and I'm pretty sure there's nothing here Circle would fail it for.

Great, thanks. I have write access now, so I'll submit future pull requests from this repo.

PR #3023, opened 2017-10-24
Adding WP Weather Station, requires php-intl

albertcausing:
A screenshot that says it requires internationalization extension from PHP.
https://monosnap.com/file/iZng5eecE9wdgutMNz0ckJ4TzovBXV


rachelwhitton:
+1

PR #3027, opened 2017-10-24
rename Helper Tools Drupal tab to Drupal 8

jackfoust:
webprofiler module is only available for D8.

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed


alexfornuto:
@jackfoust thanks for the contribution! Are you saying that although Devel itself has versions for both D7 and D8, only D8 includes the webprofiler module?

If so, can you suggest an alternative for D7?

jackfoust:
@alexfornuto WebProfiler was a D8 module that was merged into Devel and is only available in the D8 version of Devel. I'm not aware of any D7 alternatives at the moment, but will take a look.

rachelwhitton:
Awesome, thanks for the PR @jackfoust! Much appreciated 

+1

PR #3042, opened 2017-10-27
Use $_ENV for Drush related code.

dustinleblanc:
Correct recommendations on Drush policy file usage. Also correct folders referenced for storing Drush related code (sites/all/drush has been reported to be ignored by multiple customers).

## Effect
- Advises use of `$_ENV` rather than `$_SERVER` for Drush code
- Corrects folders for `drushrc.php`, sites/default seems to behave much better for customers
-

## Remaining Work
- [ ] Spotcheck voice/style for additional text

### DIFF
--- a/source/_docs/drush.md
+++ b/source/_docs/drush.md
@@ -132,6 +132,10 @@ Replace `SITENAME` with your Pantheon site name, and `example.com` with the corr
 
 ## Troubleshooting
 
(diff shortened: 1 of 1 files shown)


rachelwhitton:
Nice, thanks @dustinleblanc 

+1

alexfornuto:
👍

PR #3102, opened 2017-11-22
rate limits

### LINKED ISSUE #3095: Let's Encrypt Rate Limit When Using Let's Encrypt outside of Pantheon
Re: Let's Encrypt Rate Limit When Using Let's Encrypt outside of Pantheon
## Issue Description

Requesting a Let's Encrypt certificate for a domain not hosted on Pantheon fails with a rate limit error. Rate limiting could especially impact large organizations with many domains, such as universities.

## Suggested Resolution

Pantheon requests new certificates frequently in order to add domains to existing certificates. Let's Encrypt has increased our rate limit so we aren't running into the limit you could be reaching. It looks like there are two potential paths forward (you might try both) 
1) Ask Let's Encrypt to increase your rate limit 
- https://docs.google.com/forms/d/e/1FAIpQLSetFLqcyPrnnrom2Kw802ZjukDVex67dOM2g4O8jEbfWFs3dA/viewform
2) Add example.edu to the public suffix list, which will cause Let's Encrypt to treat every subdomain of the main domain as independent for limit purposes. Also, browsers and malware scanners will treat the subdomains as independent.
- It's just a Pull request to add to the public suffix list: https://github.com/publicsuffix/list/wiki/Guidelines

allenfear:
Closes #3095 

## Effect
Includes instructions for addressing LE rate limits

## Remaining Work
n/a

### DIFF
--- a/source/_docs/guides/launch/04-configure-dns.md
+++ b/source/_docs/guides/launch/04-configure-dns.md
@@ -83,4 +83,15 @@ Both the bare domain and the `www` domain will be accessible over HTTPS once the
 <h3 class="info">HTTPS</h3>
 <span class="glyphicons glyphicons-ok text-success"></span> Let’s Encrypt certificate deployed to Pantheon’s Global CDN. Certificate renews automatically with no additional cost.</blockquote>
 
+## Addressing Let's Encrypt Rate Limits
+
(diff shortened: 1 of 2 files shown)


allenfear:
@ari-gold brought up a good point that the launch essentials guide may not be the best spot for this. This issue doesn't really impact sites being launched on the platform, so this may be better to put into a separate doc on LE rate limits for sites hosted off the platform.

alexfornuto:
@allenfear @ari-gold I'm unclear as to why this should be in our documentation at all. Both have you have said that this issue doesn't affect sites getting certificates from our platform.

ari-gold:
+1  
Thanks @allenfear and @alexfornuto !

PR #3103, opened 2017-11-23
Update See Also section -Migrate to Pantheon: WordPress

rubelyn:
Re: [See Also Section ](https://pantheon.io/docs/migrate-wordpress/#see-also)
Add link to installing plugins via WP Admin interface on migrated sites (commonly asked question)


rubelyn:
invalid URL link

PR #3127, opened 2017-12-05
Update external-libraries.md

twooten:
This line should be removed.

"There is a project on GitHub that claims you can use it; however, we are still testing it to make sure it works."

The only GitHub repo I've found for WordPress + Tika is this - https://github.com/nanodust/masala.
Hasn't been updated in 4 years. I tested this but was unable to make it work.

### DIFF
--- a/source/_docs/external-libraries.md
+++ b/source/_docs/external-libraries.md
@@ -96,7 +96,7 @@ Download and install the Search API Attachments module ([search_api_attachments]
 If everything is working correctly, you will see the success message "Extracted data: Congratulations! The extraction seems working! Yay!".
 
 ### WordPress Tika Configuration
-There are no known plugins in the WordPress.org repository that will enable the use of Tika. There is a project on GitHub that claims you can use it; however, we are still testing it to make sure it works.
+There are no known plugins in the WordPress.org repository that will enable the use of Tika.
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
Nice catch @twooten! Before I merge, does this information change things?

https://www.wpsolr.com/knowledgebase/can-wpsolr-index-pdfs-documents-not-located-wordpress-media-library/

twooten:
@alexfornuto WPSolr might work. It's a paid plugin which I don't have access to for testing.

eostis:
Hi,

(I'm the developer of WPSOLR)

> WPSolr might work. It's a paid plugin which I don't have access to for testing.

WPSOLR comes with a free version:
https://wordpress.org/plugins/wpsolr-search-engine/

The free version can indeed index files content with Apache Solr Tika.

twooten:
Thanks for that @eostis. It looks like WPSOLR requires Solr 5.5 and above, is that correct?

eostis:
Correct: WPSOLR requires Solr 5.5.

alexfornuto:
@twooten should we revisit this issue with this new information?

twooten:
@alexfornuto we'll need to upgrade to a version of Solr that will work with this plugin.

See: FEAT-213

PR #3131, opened 2017-12-07
WIP: Freshen "assume HTTPS" documentation

davidstrauss:
The current documentation doesn't quite capture how sites in transition should assume HTTPS. Specifically, the current directions don't fully stop new content from polluting the database with HTTP URLs. This could cause an issue for transitions where the existing database URLs get fixed but new HTTP ones get added in the mean time.

I'm not sure how to harmonize all of the redirect stuff, but it's important to perform any redirects _before_ we coerce `$_SERVER` to present the current pageload as HTTPS. I'd consider linking to Drupal modules or WordPress plug-ins that can handle these redirects, but I don't see value to the configuration they offer; there's no value in 2017 at being partially HTTPS (only directing certain paths and resources).


alexfornuto:
Per convo with @davidstrauss, closing in favor of future developments.

PR #3135, opened 2017-12-08
Site Networks: Add small TOC to Considerations page

danielbachhuber:


rachelwhitton:
We can add our standard TOC to this page of the doc
screen shot 2017-12-08 at 11 23 55 am

Or do you prefer to have it in paragraph form in the intro?

danielbachhuber:
> Or do you prefer to have it in paragraph form in the intro?

My preference would be paragraph form because:

1. It offers a summary of what the user should expect.
2. It's easy for the user to miss the sidebar, because they'll only be looking at it if they think to look for a navigational aid.

PR #3136, opened 2017-12-08
Site Networks: Clarify subdirectories vs. subdomains section

danielbachhuber:


alexfornuto:
@danielbachhuber  In the context of this guide, can we consider `slugs` to be a known term for all readers?

danielbachhuber:
@alexfornuto Good question. I don't know for certain but I can't think of another word to use. @rachelwhitton Any suggestions?

rachelwhitton:
I like the term.. but yeah. We need some context around it since it's feels jargon-y

I can add a tooltip. Thoughts?
screen shot 2017-12-08 at 12 46 14 pm

rachelwhitton:
+1

PR #3144, opened 2017-12-10
How Pantheon Prevents Plan Abuse: Title & Tone

allenfear:
## Effect
Shifts emphasis slightly from abuse, which suggests an intent that may not be there, toward how customers can better understand how we measure and enforce limits.

### DIFF
--- a/source/_docs/preventing-plan-abuse.md
+++ b/source/_docs/preventing-plan-abuse.md
@@ -5,17 +5,21 @@ tags: [pricing]
 categories: []
 ---
 
+Unlike traditional providers where enough requests will crash the server, Pantheon's platform can absorb massive amounts of traffic. One of the main benefits of Pantheon is our ability to provide quality service and uptime under heavy traffic loads.
+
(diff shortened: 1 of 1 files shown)


alexfornuto:
👀 👍

alexfornuto:
@joshkoenig would you like to review before we merge?

lauren-kelly:
@joshkoenig based on the questions we've seen coming in from people, I think understanding how we count page views before selecting a plan will help prevent issues going forward.
I think this bit here is absolutely critical for people to understand:

> This includes not only HTML pages, but also responses in formats such as JSON and XML that are sometimes not tracked by standard analytics suites. API calls count as traffic, and are considered part of a site's plan limit.

The rest is really FAQ related to ^^

Also, I feel REALLY weird sending a doc recommendation to people about plan level _abuse_ when they are just asking about page views and plans, etc.

allenfear:
Good point @joshkoenig. I adjusted the subhead to "Information on how Pantheon measures site traffic". 

Happy to discuss whenever you have a free moment.

joshkoenig:
Updated with some additional intro verbage. If you all think this is ok then lets go ahead and publish!

allenfear:
Looks great to me! Thanks for the excellent additions @joshkoenig and for the helpful input @lauren-kelly .

PR #3146, opened 2017-12-12
Update new-relic.md

tinefin:
- Removing the "New Relic data is not showing on the Pantheon Site Dashboard" section, as New Relic widget is no longer in use on Dashboard.

- Moving the "Who is the New Relic account associated with?" subsection up a level, as this is not related to the subject, but a single item.
Also, now it will be possible to link to it, which we often need to in tickets / chats.

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed


rachelwhitton:
+1

PR #3154, opened 2017-12-13
Symlink Refresher

### LINKED ISSUE #3098: Non-Standard Files Locations Doc Update
Re: [Non-Standard Files Locations](https://pantheon.io/docs/non-standard-file-paths/)

Priority (Low‚ Medium‚ High):

## Issue Description:
Would be great if we could use an example which is a non standard file path - and not the standard symlink which is already there.

## Suggested Resolution

alexfornuto:
Closes #3098 

## Effect
- Copy and Style updates

## Remaining Work
- [ ] 👀  from @tinefin (optional) 
- [x] Copy review.

### DIFF
--- a/source/_docs/non-standard-file-paths.md
+++ b/source/_docs/non-standard-file-paths.md
@@ -4,32 +4,53 @@ description: Learn how to address non-standard file locations from within the Pa
 tags: [debugfiles]
 categories: []
 ---
+## Standard File Path
 Pantheon provides one location for files that are part of your sites content, like those that are managed through Drupal upload forms, e.g. user profile pictures: `/sites/default/files`. For Drupal sites, this is the *only* location you can use for files that are uploaded as part of your application. For WordPress sites, `/wp-content/uploads` is the only acceptable location for files. All other locations are considered part of your codebase, and under version control.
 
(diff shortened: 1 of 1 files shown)


source/_docs/non-standard-file-paths.md
```diff
@@ -4,28 +4,32 @@ description: Learn how to address non-standard file locations from within the Pa
...
+Pantheon provides one location for files that are part of your sites content, like those that are managed through Drupal upload forms, e.g. user profile pictures: `/sites/default/files`. For Drupal sites, this is the *only* location you can use for files that are uploaded as part of your application. For WordPress sites, `/wp-content/uploads` is the only acceptable location for files. All other locations are considered part of your codebase, and under version control.
 
-If you are importing a site that has files in another location (e.g. "/files"), you will need to move the files into the standard location, and use Git to add, commit, and push a symlink from that location to the new location:  
+If you are importing a site that has files in another location, you will need to move the files into the standard location. If you have code that expects to find files at these non-standard locations, consider editing it to refer to the correct location. Barring that, you can create a **symlink** in the expected location. Symlinks can then be committed to version control and added to the codebase.
```
rachelwhitton: > consider editing it to refer to the correct location

Hacking core/modules/plugins/themes directly is not something we should recommend in any scenario since changes will not persist. 

I'm working on some structural changes locally, I'll remove this bit in the same commit.

PR #3168, opened 2017-12-20
Switch pantheon check in $_SERVER to $_ENV

wxactly:
For consistency with other PANTHEON_ENVIRONMENT checks in this code and elsewhere

### DIFF
--- a/source/_partials/redirects.twig
+++ b/source/_partials/redirects.twig
@@ -10,7 +10,7 @@
 <div class="tab-content">
 <div role="tabpanel" class="tab-pane active" id="wp">
 <p markdown="1">Add the following to <code>wp-config.php</code> (replace <code>www.example.com</code>):</p>
-<pre id="git-pull-wp"><code class="php hljs" data-lang="hljs">if (isset($_SERVER['PANTHEON_ENVIRONMENT']) && php_sapi_name() != 'cli') {
+<pre id="git-pull-wp"><code class="php hljs" data-lang="hljs">if (isset($_ENV['PANTHEON_ENVIRONMENT']) && php_sapi_name() != 'cli') {
   // Redirect to https://$primary_domain in the Live environment
   if ($_ENV['PANTHEON_ENVIRONMENT'] === 'live') {
     /** Replace www.example.com with your registered domain name */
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
I believe that on Pantheon, we prefer `$_SERVER` over `$_ENV`.

On one particular site I checked, `$_ENV` contained 19 different variables, all of which also existed in `$_SERVER`. The `$_SERVER` variable, on the other hand, contains over 70 elements. The main difference here is that not everything in `$_SERVER` is an environment variable; all of the http headers are here, for example. The variables that exist in `$_SERVER` that are not in `$_ENV` are not often needed, so perhaps this could go either way.

If we are going to standardize, though, we should be consistent throughout the documentation. Currently, we have a roughly even usage of `$_ENV` and `$_SERVER` throughout the docs.

wxactly:
Cool, thanks @greg-1-anderson. I was going off of the [documentation](https://pantheon.io/docs/read-environment-config/#using-$_server) that says:

> use `$_ENV` instead of `$_SERVER` wherever possible

But... very possible this page is out of date or inconsistent as well! I'm totally fine using `$_ENV` or `$_SERVER`, but it would be good to know which one is preferred on the platform, and make sure the docs reflect that. Definitely happy to help update other doc pages as well... can you let me know which direction to go?

greg-1-anderson:
Thanks for the reference, @wxactly, looks like I got it backwards. Drush sets `$_SERVER`, so I guess that note is primarily for wp-cli users.

Making the docs consistently use `$_ENV` would be much appreciated; the only care that must be taken is to only convert from `$_SERVER` when the variable in question is also available in `$_ENV`, e.g. for values such as:

```
FRAMEWORK
DOCROOT
FILEMOUNT
DRUPAL_HASH_SALT
php_version
DB_HOST
DB_PORT
DB_USER
DB_PASSWORD
DB_NAME
PANTHEON_SITE
PANTHEON_SITE_NAME
PANTHEON_ENVIRONMENT
PANTHEON_INFRASTRUCTURE_ENVIRONMENT
PATH
HOME
PHPRC
PANTHEON_DEPLOYMENT_IDENTIFIER
PANTHEON_ROLLING_TMP
```

alexfornuto:
@wxactly thanks for the heads up!

PR #3197, opened 2018-01-09
3168 server env

alexfornuto:
Closes #3168 

## Effect
- Switches instances of `$_SERVER` to `$_ENV`, when using a variable listing as valid for both.

## Remaining Work
- [x] Review from @greg-1-anderson


alexfornuto:
Thanks @greg-1-anderson!

PR #3209, opened 2018-01-12
Instruct customers to read Platform Considerations Docs when migrating.

dustinleblanc:
See chat #14045975809

Customer migrated site and wasn't aware of email limitations of the platform, pointed out we should document that better.


rachelwhitton:
+1

PR #3212, opened 2018-01-12
Compell Cache Clearing CLI Commands

ttrowell:
## Effect
- Just adds a bit to let us know it can only be successfully done via Terminus.

- This will not be true after the next version of WP-cli is released and we update the platform:
https://github.com/wp-cli/wp-cli/pull/4527

-

## Remaining Work
- [x] Technical Review
- [x] Copy Review


rachelwhitton:
@danielbachhuber can you take a look?

alexfornuto:
Thanks @ttrowell and @danielbachhuber!

PR #3219, opened 2018-01-16
Explicitly warn that Composer will break upstream updates.

dustinleblanc:
## Effect
- Add warning to Composer docs about upstream updates breaking.

## Remaining Work
- [x] Copy check for voice

### DIFF
--- a/source/_docs/composer.md
+++ b/source/_docs/composer.md
@@ -8,10 +8,10 @@ searchboost: 150
 <div class="alert alert-danger" role="alert">
   <h4 class="info">Warning</h4>
   <p markdown="1">
-    Managing a Drupal 8 site with Composer is an all or nothing proposition. Our Drupal 8 upstream includes a `composer.json` and a fully built `/vendor` directory. When a site developer executes a `composer require anyone/anything:anyversion` within a site repository built from [drops-8](https://github.com/pantheon-systems/drops-8){.external}, this is effectively the same thing as [overwriting Drupal core](/undo-commits/).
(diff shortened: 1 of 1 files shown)


dustinleblanc:
@alexfornuto you so fast!

dustinleblanc:
@alexfornuto The location this inserted in is probably fine, I am fine with it as is

rachelwhitton:
+1 

Made some heavy edits, @dustinleblanc can you review?

dustinleblanc:
@rachelwhitton I think it reads a lot better, but I think it could use some tweaking for technical accuracy and impact:

- I am worried it's getting buried further down the page, I'd love this warning to never be missed by anyone who fancies to use Composer.
- > Pantheon does not support using Composer to manage a subset of project dependencies, such as libraries but not core. Doing so is especially problematic for Drupal 8 sites since Composer is used by core, meaning any change to composer.json or the vendor directory would result in massive merge conflicts when trying to update core via one-click updates in the Pantheon Site Dashboard. 

The first sentence I feel discourages something that isn't problematic. The real issue being discussed here is only relevant to Drupal 8 sites that track `drops-8` where core's dependencies are tracked in our upstream, a user cannot manage those dependencies themselves and use upstream updates. All other uses of Composer are perfectly fine and I don't want to indicate that someone has to swap upstreams in situations where they don't need to do so at all.

dustinleblanc:
@rachelwhitton I'll accept the compromise, it errs on the side of caution which was my aim, I guess it is better to discourage a touch of fancy then our current state of not warning about the pitfalls. I approve this message

PR #3221, opened 2018-01-17
Explain why aggregation is good even with HTTP/2

davidstrauss:
Responds to 2018-01-12 documentation feedback, which suggested that HTTP/2 should replace the need for aggregation.


alexfornuto:
+1 Thanks @davidstrauss

PR #3236, opened 2018-01-19
Clarify nginx-access.log retention policy

xq42:
Expand "60 days of logs" (line 29) to the more precise "Up to 60 days of logs".

A customer asked how they could access all 60 days of their nginx-access.log based on the information shown here, which is misleading / inaccurate information as we only keep the logs until the appserver they're on is migrated. If their appserver has just been migrated, they may only have a few hours (or less) of logs available.

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed

### DIFF
--- a/source/_docs/logs.md
+++ b/source/_docs/logs.md
@@ -26,7 +26,7 @@ The server timezone and all log timestamps are in UTC (Coordinated Universal Tim
       </tr>
       <tr>
         <th>nginx-access.log</th>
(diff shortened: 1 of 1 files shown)


alexfornuto:
Apparently the big bold warning right next to it explaining that isn't enough for some readers...

+1

rachelwhitton:
Thanks for the pr @xq42, much appreciated! +1

PR #3239, opened 2018-01-19
New Guide: Manage Some Dependencies with Composer

### LINKED ISSUE #3238: New Guide: Manage Some Dependencies with Composer
Re: [Composer Fundamentals and Workflows](https://pantheon.io/docs/composer/)

Priority (Low‚ Medium‚ High): Med

## Issue Description:
The page discourages partial adoption for all frameworks, which is overkill for D7/WP projects that do not require fully converting to the build tools workflow. 

## Suggested Resolution 
Create a new guide to document "some composer" on D7/WP that preserves one-click core updates 

cc @dustinleblanc

rachelwhitton:
Closes #3238 

## Effect
- Adds a new guide for partial Composer adoption on D7/WP sites
- Direct users to the guide from the Composer doc in a new section called partial adoption
- Revise warning placements to reflect newly documented procedure, noting D8 is not supported.

## Remaining Work
- [x] Work with @dustinleblanc to figure out what's wrong with my D7 instructions (core is being overwritten when requiring drupal packages)
  - [x] Correct D7 tab in Require Dependencies section and replace `source/docs/assets/images/guides/partial-composer/require-papc-module.png` with a screenshot that doesn't modify core


source/_docs/guides/partial-composer.md
```diff
@@ -0,0 +1,231 @@
...
+      git push origin master
+      ```
+    </div>
+
+    <!-- 2nd pane content -->
+    <div role="tabpanel" class="tab-pane" id="d7-init" markdown="1">
+    1. Initialize composer to create a `composer.json` file with the Drupal 7 package repository:
+
+      ```command
+      composer init --repository=https://packages.drupal.org/7 --no-interaction -s=alpha
```
greg-1-anderson: Is stability: alpha the best recommendation here? Maybe we should say `-s=stable`. This is the default, but putting it in would make it clear how folks should alter this if they want to use unstable modules.

rachelwhitton: Ah, good catch.  I’ll fix this up. The guide goes through requiring the PAPC module in the following section so that’s why it’s alpha here. Planning to switch that demo to a package that doesn’t depend on drupal to avoid overwriting core, once I do that I’m not sure we’ll need to specify stability (unless we have specific suggestions, otherwise we can link to composer docs).

PR #3242, opened 2018-01-20
Update platform-considerations.md

twooten:
UTF8MB4 Convert is not a module that can be enabled. Downloading via the normal method will place the command in $HOME/.drush but it will need to be moved to the proper location so that Drush can find it.


rachelwhitton:
Thanks @twooten! +1

alexfornuto:
+1

PR #3243, opened 2018-01-20
merge clarification

allenfear:
The original text suggests that you can use the dashboard to merge code between multidev environments, which does not appear to be the case, at least I didn't see a way to use the dashboard to merge code directly from multidev branch into another multidev branch. Minor rewrite to make more explicit that merges occur to and from master.


rachelwhitton:
+1 thanks @allenfear!

PR #3253, opened 2018-01-23
clarify CNAME reqs for auto-cert renewal

### LINKED ISSUE #3216: Requirements for Automated Certificate Renewal
Re: [HTTPS on Pantheon's Global CDN](https://pantheon.io/docs/https/)

Priority (Low):

## Issue Description:
Customer wanted to be clear if A and AAAA records are actually needed for automated certificate renewal

Also, customer asked about how long and when they should check if the certificate renewal worked

## Suggested Resolution 
No need for A and AAAA records. The requirement is only to have the assigned domains and its DNS records be pointed to Pantheon, regardless of A, AAAA or CNAME records

Also, need to point out when do we renew certs, and in what frequency (number of days, months etc)

alexfornuto:
Closes #3216 

## Effect
- Adjusts CNAME requirements for certificate renewal.

The concern posed was:

> Customer wanted to be clear if A and AAAA records are actually needed for automated certificate renewal

Per @kf6nux:

> Yes, at least one A/AAAA is required, but can be found via CNAME chasing. If they CNAME to their platform-domain, our platform's A/AAAA records will be used. 

This attempts to clarify the point based on the above concern and clarification.

## Remaining Work
- [ ] Review from @eabquina and/or @kf6nux 
- [ ] Copy Review

### DIFF
--- a/source/_partials/content/https-requirements.html
+++ b/source/_partials/content/https-requirements.html
@@ -1,6 +1,6 @@
 <h3>Requirements for Automated Certificate Renewal</h3>
 <ul>
-  <li>All A/AAAA/CNAME/DNAME DNS records must point to Pantheon's servers so Let's Encrypt can verify domain ownership. MX/NS and similar records can point to other addresses.</li>
+  <li>All A/AAAA/CNAME/DNAME DNS records must point to Pantheon's servers so Let's Encrypt can verify domain ownership.</li>
   <li>Authoritative Name Servers must serve mixed-case lookups, and must not fail CAA lookups.</li>
   <li>CAA records must either 1) not exist for the domain and its parent domains or 2) authorize Let's Encrypt.</li>
   <li>AAAA records are not required, but if set must exclusively point to Pantheon.</li>
--- a/source/_partials/content/https-requirements.html
+++ b/source/_partials/content/https-requirements.html
@@ -1,8 +1,8 @@
 <h3>Requirements for Automated Certificate Renewal</h3>
 <ul>
   <li>All A/AAAA/CNAME/DNAME DNS records must point to Pantheon's servers so Let's Encrypt can verify domain ownership.</li>
+  <li>AAAA records are not required, but if set must exclusively point to Pantheon.</li>
   <li>Authoritative Name Servers must serve mixed-case lookups, and must not fail CAA lookups.</li>
   <li>CAA records must either 1) not exist for the domain and its parent domains or 2) authorize Let's Encrypt.</li>
-  <li>AAAA records are not required, but if set must exclusively point to Pantheon.</li>
 </ul>
 <p>Please check this doc periodically as requirements may be added or removed.</p>
--- a/source/_partials/content/https-requirements.html
+++ b/source/_partials/content/https-requirements.html
@@ -4,7 +4,7 @@
   <li>Authority Name Servers must serve mixed-case lookups.</li>
   <li>Authority Name Servers must not fail CAA lookups.</li>
   <li>CAA records must either 1) not exist for the domain and its parent domains or 2) authorize Let's Encrypt.</li>
-  <li>CNAMEs must follow the above requirements.</li>
+  <li>CNAMEs records must point to Pantheon's provided <a href="/docs/domains/#platform-domains">platform domain</a> for the target environment. Subdomains such as <code>www.example.com</code> do not require A/AAAA records.</li>
   <li>DNAMEs must not be used.</li>
 </ul>
 <p>Please check this doc periodically as requirements may be added or removed.</p>
--- a/source/_partials/content/https-requirements.html
+++ b/source/_partials/content/https-requirements.html
@@ -1,9 +1,8 @@
 <h3>Requirements for Automated Certificate Renewal</h3>
 <ul>
-  <li>All DNS records, whether A/AAAA/CNAME, must point to Pantheon's servers so Let's Encrypt can verify domain ownership. Verification is performed via a special file hosted on the platform (e.g., <code>http://example.com/.well-known/acme-challenge/[variableDataHere]</code>)</li>
+  <li>All A/AAAA/CNAME/DNAME DNS records must point to Pantheon's servers so Let's Encrypt can verify domain ownership. MX/NS and similar records can point to other addresses.</li>
   <li>Authoritative Name Servers must serve mixed-case lookups, and must not fail CAA lookups.</li>
   <li>CAA records must either 1) not exist for the domain and its parent domains or 2) authorize Let's Encrypt.</li>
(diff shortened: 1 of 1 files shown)


eabquina:
@alexfornuto @kf6nux 
Thanks and this clarifies the concern. 

It is safe to say that when for sub-domains (www.example.com), adding the CNAME record is enough, and doesn't need to explicitly add A or AAAA records in their DNS settings for this to work, right?

kf6nux:
@eabquina Correct.  It would be a violation of DNS specification for an individual to have both a CNAME and other resource records (such as A/AAAA).  For further information, checkout the wikipedia page for CNAMEs https://en.wikipedia.org/wiki/CNAME_record (specifically the second bullet point under "Restrictions")

kf6nux:
@eabquina Correct.  It would be a violation of DNS specification for an individual to have both a CNAME and other resource records (such as A/AAAA).  For further information, checkout the wikipedia page for CNAMEs https://en.wikipedia.org/wiki/CNAME_record (specifically the second bullet point under "Restrictions")

kf6nux:
@alexfornuto LGTM.  If you want to do it as part of this PR, we can also drop the "DNAMEs must not be used." requirement.  We just verified today this is no longer a restriction for us.

ari-gold:
I think we can simplify this a bit. Will share more detailed suggestions.

alexfornuto:
I adjusted the clarification based on (read: 100% stolen from) @rachelwhitton's suggestions.

ari-gold:
@alexfornuto   Taking a stab at it:

<h3>Requirements for Automated Certificate Renewal</h3>
<ul>
  <li>All DNS records must point to Pantheon so Let's Encrypt can verify domain ownership. 
  <li>Domain verification must complete successfully by allowing Pantheon to serve a special file (e.g., <code>http://example.com/.well-known/acme-challenge/[variableDataHere]</code>).</li>
  <li>AAAA records are not required, but if set, must exclusively point to Pantheon.
  <li>Authorititative Name Servers must serve mixed-case lookups.</li>
  <li>Authorititative Name Servers must not fail CAA lookups.</li>
  <li>CAA records must either 1) not exist for the domain and its parent domains or 2) authorize Let's Encrypt. For help generating CAA records see SSL Mate's [CAA record helper](https://sslmate.com/caa/).</li>
</ul>
<p>Please check this doc periodically as requirements may be added or removed.</p>

kf6nux:
> clarify here with
> "All DNS Records, whether A/AAAA/CNAME, must point to..."

I think that clarification would be good, as we *do allow* them to have other records (e.g. MX/NS) that do not point at Pantheon.

I also think that if we're broadening the first requirement to "All A/AAAA/CNAME" must point at Pantheon, then we can drop the following requirements as they're either redundant or then outside of the customer's control:
> Domain verification must complete successfully by allowing Pantheon to serve a special file (e.g., http://example.com/.well-known/acme-challenge/[variableDataHere]).
> AAAA records are not required, but if set, must exclusively point to Pantheon.

@ari-gold thoughts?

alexfornuto:
Assigning to @ari-gold as I'd like his input on @kf6nux's comments. By biggest concern is getting confirmation that "we * do allow* them to have other records".

My site is hosted on pantheon, and mail for that domain is handled by MX records pointing to Google, so I'm fairly sure the answer is yes; but fairly sure doesn't cut it in docs. 😉

alexfornuto:
I deleted my last comment after reading this thread with fresh eyes and realizing there wasn't a question posed as I thought there was. 

@kf6nux I've made adjustments as you've suggested, please have a look.

kf6nux:
I still think the last point is redundant and am +0 on including it.  I'm +1 on everything else.

source/_partials/content/https-requirements.html
```diff
@@ -1,10 +1,8 @@
 <h3>Requirements for Automated Certificate Renewal</h3>
 <ul>
-  <li>All A/AAAA records must point to a server that serves Pantheon's acme-challenge response (e.g., <code>http://23.185.0.1/.well-known/acme-challenge/[variableDataHere]</code>).</li>
-  <li>Authority Name Servers must serve mixed-case lookups.</li>
-  <li>Authority Name Servers must not fail CAA lookups.</li>
+  <li>All A/AAAA/CNAME/DNAME DNS records must point to Pantheon's servers so Let's Encrypt can verify domain ownership. MX/NS and similar records can point to other addresses.</li>
```
ari-gold: Let's remove the line about MX/NS. They *can't* point to Pantheon if you want things to work, so I think it just confuses things.

lauren-kelly:
@kf6nux @alexfornuto I think the redundancy there is to try to reduce the confusion around AAAA records. The dashboard is saying they need to add them, but sometime they can't with their domain manager...and other times they have existing ones and don't remove them. 
Because of those two cases, I think we need to have some very direct, clear way of saying: you do not have to have AAAA records...but if you do, they _have_ to point to Pantheon - even if it is redunant

alexfornuto:
I think @lauren-kelly accurately summed up the concerns here. I'm going to leave this line in, and resubmit this for copy review from @rachelwhitton.

source/_partials/content/https-requirements.html
```diff
@@ -1,10 +1,8 @@
...
 <ul>
-  <li>All A/AAAA records must point to a server that serves Pantheon's acme-challenge response (e.g., <code>http://23.185.0.1/.well-known/acme-challenge/[variableDataHere]</code>).</li>
-  <li>Authority Name Servers must serve mixed-case lookups.</li>
-  <li>Authority Name Servers must not fail CAA lookups.</li>
+  <li>All A/AAAA/CNAME/DNAME DNS records must point to Pantheon's servers so Let's Encrypt can verify domain ownership.</li>
+  <li>Authoritative Name Servers must serve mixed-case lookups, and must not fail CAA lookups.</li>
   <li>CAA records must either 1) not exist for the domain and its parent domains or 2) authorize Let's Encrypt.</li>
-  <li>CNAMEs must follow the above requirements.</li>
-  <li>DNAMEs must not be used.</li>
+  <li>AAAA records are not required, but if set must exclusively point to Pantheon.</li>
```
ari-gold: How about we keep and move it to bullet point 2?  @kf6nux   Alex and I discussed outside of this PR that it's probably a good idea to keep because there's some ambiguity around AAAA records being "required" from our UI.

alexfornuto: @ari-gold Do you mean merge it with the current second bullet point, or moving it into the 2nd position? I'm fine with the latter.

ari-gold: The latter =)

rachelwhitton:
+1

PR #3255, opened 2018-01-23
Troubleshoot redis installs on D8 sites using build tools

### LINKED ISSUE #3215: Installing Redis problem in Composer+CI D8
Re: [Installing Redis on Drupal](https://pantheon.io/docs/drupal-redis/)
## Issue Description

The instructions for adding Redis to a D8 site don’t include instructions that work with a [drops-8-composer](https://github.com/pantheon-systems/example-drops-8-composer) site. 

While it was easy enough to adapt them in a branch on my local development copy, CircleCI was unable to test it. In the 
`prepare database for site-under test` stage, creating a multidev environment failed with this error:
```
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException:
You have requested a non-existent service "cache.backend.redis".
```

## Suggested Resolution

I described the problem in the Pantheon Community slack, and @greg-1-anderson suggested adding the code first, then updating the settings. This approach worked.

Step 2 should be amended to mention that the Composer workflow requires the code to be deployed to Pantheon _before_ the settings can be updated.

rachelwhitton:
Closes #3215 

## Effect
- New troubleshooting section to instruct users to deploy the module to Pantheon when non-existent service error occurs

### DIFF
--- a/source/_docs/redis.md
+++ b/source/_docs/redis.md
@@ -366,6 +366,16 @@ You skipped a step; `settings.php` must include the cache\_backport files. Add t
 $conf['cache_inc'] = 'sites/all/modules/cache_backport/cache.inc';
 ```
 
+### You have requested a non-existent service
+Drupal 8 sites using Composer as recommended by the [Build Tools](/docs/guides/build-tools/) guide will encounter the following error when configuring settings before the module is deployed to Pantheon:
+
+```php
+Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException:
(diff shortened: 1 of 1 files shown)


alexfornuto:
@greg-1-anderson Would you care to take a look at this?

greg-1-anderson:
LGTM. Of course, this can happen with or without Composer, since this error will come up any time someone modifies settings without enabling the module. Composer users are more likely to try to do all of the steps in one PR, though, in which case they will disable their site until they back the changes out of settings.php. So, in short, I think the note is fine as added.

PR #3259, opened 2018-01-24
Generalize service error

rachelwhitton:
Related to #3255 

## Effect
- Make the new troubleshooting section more general, since it's not specific to Composer managed sites. 

cc @greg-1-anderson

### DIFF
--- a/source/_docs/redis.md
+++ b/source/_docs/redis.md
@@ -367,14 +367,14 @@ $conf['cache_inc'] = 'sites/all/modules/cache_backport/cache.inc';
 ```
 
 ### You have requested a non-existent service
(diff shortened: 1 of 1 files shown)


alexfornuto:
+1 for generalization and simplification, for which the ramification is better explanation.

greg-1-anderson:
Yes, that is an improvement, but it should be "the redis module", not "a module".

greg-1-anderson:
@rachelwhitton I was too slow by a minute 😛

alexfornuto:
I think it can stand as is, since the scope of the rest of the guide makes "the module" implicitly refer to Redis for readers of the guide. We then have the benefit of pointing others to this anchor for non-redis instances of the error.

PR #3271, opened 2018-01-25
Getting Started Training Footer CTA

rachelwhitton:
## Effect
- Adds a new CTA to the footer to refer traffic to the new getting started training offering

## Remaining Work
- [x] Work with @dgorton to finalize copy
- [x] Get some design input on using the blue and stacking CTAs like this (@kimby77 if you have some bandwidth to weigh in?)
     - Should change the font color to increase contrast for accessibility

### DIFF
--- a/source/_partials/footer_content.html
+++ b/source/_partials/footer_content.html
@@ -1,10 +1,9 @@
 <div class="container-fluid">
   <div class="row">
     <div class="pio-docs-footer hidden-sf" id="footer">
-      <div class="office-hours">
+      <div id="office-hours" class="office-hours">
         <h3 style="margin-top:0;">Office Hours</h3>
-        <p>Got questions? We have answers!<br>Our Agency and Community Success team holds office hours every Wednesday.</p>
-        <br><a href="https://pantheon.io/agencies/office-hours" class="cta cta-black-on-yellow">More</a>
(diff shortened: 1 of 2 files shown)


kimby77:
The two footers are confusing because they have equal visual weight, so my attention is not being drawn to one or the other.  Which one is the main one I (as a user) should be paying attention to?  Or can this be redesigned to offer both options in a single footer?

You can use this to check color contrast: http://leaverou.github.io/contrast-ratio/

dgorton:
Fair enough. How about reworking the single block to cover both options?

Something like: 
Personalized Help
Got questions? We have answers!
Our Developer Relations team holds office hours every Wednesday and regular Getting Started with Pantheon sessions. Join us!

And then link Office Hours and Getting Started. Pull the 'More' button.

Better?

rachelwhitton:
+1 @dgorton @kimby77 

But now I'm not sure anchors work best on the background.. suggestions? 

screencapture-docs-local-8000-docs-1516921437757 1

dgorton:
Ugh. Pretty sure blue on yellow is a serious legibility issue. How about just black text with underlines ?
(Taking into account that IANADesigner)
Also - minor nit - let's say "Office Hours" in Title Case (and - yes - I know I'm the culprit as that's what I wrote above)

PR #3272, opened 2018-01-25
Clarify Code VS Content

### LINKED ISSUE #3143: Describe Content VS Code Better

## Title
Understanding the difference between code, files and content

## Description
Many people get tripped up when cloning "files" to Test or Live and not understanding why their new plugin isn't showing up. I think part of the reason for the confusion is due to our usage of the word "files". The code that makes up a module, plugin, template, etc. is in a "file". Perhaps we should change "files" to "assets"?

## Outline
There are a few concepts that need a bit of separation, such as:

- code (these are the php files – plugins, themes, etc)
- files (images, text files, pdf documents, basically anything in wp-content/uploads/ or sites/default/files)
- content (this is the database – posts, pages, nodes, etc)

The database and files can be modified directly in production, without going through the dev —> test —> live workflow. This is actually recommended: you want your users to write posts and upload files directly in Live.

The code however needs to go through version control and thus whenever you modify a php file (i.e. a plugin) you need to make a commit, check it in dev, and then push to test and live. Installing a plugin directly in Live is not possible.

## Expected Audience
Anyone new to the platform.

## Path
I'm not sure about the best place for this. https://pantheon.io/docs/pantheon-workflow might be a good place but it's already a fairly long guide.

## Reason
This comes up all the time. The text in the Outline section above is our go-to snippet for when we need to explain this. Searching internally for the keywords "bit of separation" for when this was used results in 16 chats and 44 tickets. I field probably 4 or 5 of these a week and I'm sure I'm not the only one. 

Examples from Desk:
51921
76840

And Chat:
13584216747
13585479321

alexfornuto:
Closes #3143 

## Effect
- Explains in greater detail the difference between code and content.

## Remaining Work
- [ ] Technical Review
- [ ] Copy Review

### DIFF
--- a/source/_docs/pantheon-workflow.md
+++ b/source/_docs/pantheon-workflow.md
@@ -5,21 +5,24 @@ tags: [workflow, dashboard]
 categories: []
 ---
 
-Every Pantheon site comes with three environments: Dev, Test, and Live. Separate Dev, Test, and Live environments allow you to develop and test your site without impacting the live site's availability to the world. Additional development environments are available with [Multidev](/docs/multidev/).
+Every Pantheon site comes with three environments: Dev, Test, and Live. Each environment runs a version of the site on its own container. Separate Dev, Test, and Live environments allow you to develop and test your site without impacting the live site's availability to the world. Additional development environments are available with [Multidev](/docs/multidev/).
 
 <div class="enablement">
   <h4 class="info" markdown="1">[Get DevOps Training](https://pantheon.io/agencies/learn-pantheon?docs){.external}</h4>
   <p>Optimize your dev team and streamline internal workflows. Pantheon delivers custom workshops to help development teams master our platform and improve their internal DevOps.</p>
 </div>
 
-## Code VS Content
+## Components of a Site
+One of the core concepts at the heart of the Pantheon workflow is the distinction between **code** and **content**.
+### Code
+Code refers to anything version controlled by Git which includes core, custom and contributed modules or plugins, themes, and libraries.
 
-One of the core concepts at the heart of the Pantheon workflow is the distinction between **code** and **files**.
+{% include("content/code.html")%}
 
- - **Code** is the core of your CMS, plugins/modules, themes, css, javascript - anything that builds and shapes your site.
- - **Content** is your files (static images, posts, PDF files), and the database itself.
+### Content
+Content refers to your sites files and the database. In this context, files are static images and assets stored in the standard upload path `wp-content/uploads` for WordPress and `sites/all/default/files` for Drupal.
 
-### Code Moves Up, Content Moves Down
+## Code Moves Up, Content Moves Down
 <img src="/source/docs/assets/images/workflow.png" alt="Dev Test and Live icon" style="border:0;margin-left:auto;margin-right:auto;display:block;">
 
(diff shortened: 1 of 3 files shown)


source/_docs/pantheon-workflow.md
```diff
@@ -12,13 +12,17 @@ Every Pantheon site comes with three environments: Dev, Test, and Live. Separate
   <p>Optimize your dev team and streamline internal workflows. Pantheon delivers custom workshops to help development teams master our platform and improve their internal DevOps.</p>
 </div>
 
-## Code Moves Up, Content Moves Down
-<img src="/source/docs/assets/images/workflow.png" alt="Dev Test and Live icon" style="border:0;margin-left:auto;margin-right:auto;display:block;">
+## Code VS Content
```
rachelwhitton: Suggest reverting header changes - this doesn't summarize the section as effectively as the original h2

source/_docs/pantheon-workflow.md
```diff
@@ -12,13 +12,17 @@ Every Pantheon site comes with three environments: Dev, Test, and Live. Separate
...
+
+One of the core concepts at the heart of the Pantheon workflow is the distinction between **code** and **files**.
```
rachelwhitton: We should be consistent in the two components we're discussing, **code** and **content**

source/_docs/pantheon-workflow.md
```diff
@@ -12,13 +12,17 @@ Every Pantheon site comes with three environments: Dev, Test, and Live. Separate
...
 
-The core of the Pantheon workflow is to move code up from Dev to Test to Live and content down from Live to Test to Dev.
+ - **Code** is the core of your CMS, plugins/modules, themes, css, javascript - anything that builds and shapes your site.
```
rachelwhitton: Suggest reverting "anything that builds and shapes your site" to something closer to the original, which specifically noted that code is tracked by Git.

rachelwhitton: Can we show them something here? A code directory structure in a drop down panel maybe? To help connect the dots visually?

source/_docs/pantheon-workflow.md
```diff
@@ -12,13 +12,17 @@ Every Pantheon site comes with three environments: Dev, Test, and Live. Separate
...
+ - **Content** is your files (static images, posts, PDF files), and the database itself.
+
+### Code Moves Up, Content Moves Down
+<img src="/source/docs/assets/images/workflow.png" alt="Dev Test and Live icon" style="border:0;margin-left:auto;margin-right:auto;display:block;">
 
-- **Code** includes plugins, modules, themes, CSS, JS—anything that's under Git version control.
-- **Content** includes files not under Git version control, like images, PDFs, and the database.
+The main process of the Pantheon workflow is to move code up from Dev to Test to Live and content down from Live to Test to Dev. To facilitate this, we put [files](/docs/files/) into our distributed filesystem, Valhalla, and [code](/docs/code/) on to the application servers. When you build or migrate your site to Pantheon, configuring the correct paths initially will avoid complications down the road.
```
rachelwhitton: Suggest moving the definitions for code and content back to their original placement, after this paragraph. So we present the reader with the concept before breaking it down?

source/_docs/pantheon-workflow.md
rachelwhitton: Maybe these should be tabs? instead of bullets?

rachelwhitton:
Optional, but consider working #3279 into this PR

rachelwhitton:
+1

PR #3290, opened 2018-02-01
Team Permissions Matrix

### LINKED ISSUE #3167: Site Team Management Doc Update
Re: [Site Team Management](https://pantheon.io/docs/team-management/)

Priority (Low‚ Medium‚ High):

## Issue Description:

We often get questions on permissions for non organization sites - plus CSE needs this information.

## Suggested Resolution 

It would be great if we could add a permission matrix for non orgs sites too like the ones here:
https://pantheon.io/docs/change-management/#enterprise-and-partner-organizations-roles-and-permissions
Even if the matrix will only have a single role with access to everything, it would be very helpful for clarity.

It would also be great to specify whether a site need to be owned by an org to get the change management features, or it is enough to add a supporting org to get change management.
And then it would be great to specify which org types qualify and which does not (Ally, Edu, Edu+ for instance - where do these fit in?)

It would also be great to add information about which permissions org members will achieve on sites where the org is added as a "supporting org".

### This is how I _think_ it works:
(but please do double check with someone who knows for sure, cause I am highly confused too! :) )

In order to get change management, a site needs to be owned by either an enterprise or partner org? Or perhaps just be connected to the org?
All other sites (including those owned by edu / edu+ / ally orgs) get only a site owner role and a team member role.

The only difference between site owner and team member, is that the site owner pays for the site (plus perhaps only site owner can delete the site?).

Regarding org member permissions when an org is added as a "supporting org" for a site I would assume the supporting org members get the same permissions as they have on a site owned by / connected to the org. But here I am very much in doubt. 

NB: Please do note I am very (!) much in doubt about the details here - but this is my best "guess".

alexfornuto:
Closes #3167 

## Effect
- Copies the Team Permissions Table from change-management, which covers all the questions raised in the issue.

## Remaining Work
- [ ] Review from @lauren-kelly / @tinefin 
- [ ] Copy Review

### DIFF
--- a/source/_docs/team-management.md
+++ b/source/_docs/team-management.md
@@ -10,6 +10,59 @@ Pantheon has powerful workflow tools that are packed with real-time features tha
 You can add and delete users in the Team modal by clicking **Team** in the upper-right of your Site Dashboard.
 
 ![Team Management Window](/source/docs/assets/images/dashboard/team-modal.png)
+
+##Site Team Roles and Permissions
+
+<table class="table  table-bordered table-responsive">
+    <thead>
+      <tr>
+        <th>Permissions</th>
+        <th>User in Charge / Owner <a rel="popover" data-proofer-ignore data-toggle="tooltip" data-html="true" data-title="Owner" data-content="Partner orgs only"><em class="fa fa-info-circle"></em></a></th>
+        <th>Team Member </th>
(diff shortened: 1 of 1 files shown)


alexfornuto:
Replaced by #3291

PR #3297, opened 2018-02-05
Troubleshooting update for issue #114

mcdwayne:
It was not obvious from the name of he ticket this was related on github.  Adding here for clarification until fix is made

 If the user is not logged in or authentication is not valid, terminus build:project-create displays 
       ```
       BuildToolsCommand.php line 166:  
           The site name exampleuniquesitename is already taken on Pantheon.
       ```
    Make sure you are logged in with the command
    ```
    $ terminus auth:whoami
    ```
    Create and add a new [Machine Token](https://pantheon.io/docs/machine-tokens/)

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed


alexfornuto:
If the issue is that the user is not authenticated, shouldn't Terminus say that instead of saying the site name is taken? 

If so, would you agree that this should be addressed in Terminus itself?

mcdwayne:
In a perfect world,yes should just get fixed.  Given that the bug #114 has
been open since October and no fix has been attempted this will help in the
short term.  Not proposing this as a long term solution.

PR #3299, opened 2018-02-05
Remove D8 faq for launch check

rachelwhitton:
## Effect
- Removes FAQ for D8 support of launch check


alexfornuto:
:+1:

PR #3319, opened 2018-02-08
Specify unable to select time.

alexfornuto:
## Effect
- Clarifies that the user is unable to select when cron is processed.
-
-

## Remaining Work
- [ ] If needed


alexfornuto:
Saving for training example

PR #3320, opened 2018-02-09
Specify domain access above redis config

### LINKED ISSUE #3289: Reading Pantheon Environment Configuration Doc Update
Re: [Reading Pantheon Environment Configuration](https://pantheon.io/docs/read-environment-config/)

Priority (Low‚ Medium‚ High):

## Issue Description:

I am pretty sure that domain access config needs to go above fi Redis config and fast_404 etc. Could we double check and edit accordingly?
I think Tara might remember too - we had a couple of issues on this that she was in on.

## Suggested Resolution

alexfornuto:
Closes #3289

## Effect
- Specifies order between the two referenced items in `settings.php`

## Remaining Work
- [ ] :+1: from @tinefin 
- [x] Copy Review

### DIFF
--- a/source/_docs/read-environment-config.md
+++ b/source/_docs/read-environment-config.md
@@ -64,7 +64,7 @@ Pantheon uses Pressflow to automatically read the environmental configuration. I
 
 ## Domain Access
 
-Place [Domain Access setup routine](https://www.drupal.org/node/1096962) at the **end** of settings.php. For example, for Drupal 7:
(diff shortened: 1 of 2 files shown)


rachelwhitton:
+1

tinefin:
The redis.md file now reads (to me) a bit like domain access conf should be in every settings.php file.
Redis is in use on far the most sites - Domain access is a bit of a niche setup. Redis conf doesn't need to go at the end either.

Perhaps we could leave out domain access from redis.md? -Or make it something like:
NB: If using domain access, then the domain access configuration block needs to go before the redis configuration block in settings.php

PR #3325, opened 2018-02-09
removing ambiguious statements

lauren-kelly:
## Effect
* not every team member gets an multidev - there's a max of 10 per site.
* multidevs do not mimic Live  - they are another Dev environment

## Remaining Work
- [x] Verify removed content is inaccurate


alexfornuto:
+1 from me, but I'm not an expert on multidev environments.

PR #3339, opened 2018-02-13
Reinforce domain masking limitations.

### LINKED ISSUE #3310: Platform Considerations Doc Update
Re: [Platform Considerations](https://pantheon.io/docs/platform-considerations/)

Priority (Low‚ Medium‚ High):

## Issue Description:
Would be good to either take out the section on Domain Masking - or adding in there that Pantheon Support will not be able to assist with this, it is unsupported / out of scope for support.

Unsupported meaning: 
Customers can set this up if they feel confident they can make it work + maintain it themselves (including debugging any related issues later on) - but they are on their own here. We have no visibility, this is entirely in customer court yard.
Pantheon support cannot assist with guidance or debugging.

In case of any domain / redirect / etc related issues later on, Pantheon support will not be able to assist. If assistance is needed, domain masking might need to be removed at least while debugging, in order to rule this out as a cause or part of the cause.

We do not prohibit Domain masking, but we cannot recommend it either for the above reasons (unsupported).

Context:
Discussion with @ttrowell @slashzero just above this entry here:
https://pantheon.slack.com/archives/C03SRV7HB/p1517841518000536

## Suggested Resolution

slashzero:
First cut.

Closes #3310 

## Effect
- Reword to discourage use of domain masking and reiterate OOS.
-
-

## Remaining Work
- [ ] If needed

### DIFF
--- a/source/_docs/platform-considerations.md
+++ b/source/_docs/platform-considerations.md
@@ -51,7 +51,9 @@ Domain masking allows you to serve two entirely different and separate sites ove
 * Main Site: `https://www.example-site.com/`
 * Blog: `https://www.example-site.com/blog/`
 
-In and of itself, Pantheon does not support this kind of setup. Each site on the platform must have its own unique domain. However, this configuration can be achieved using an external service such as CloudFront, StackPath, Fastly, etc.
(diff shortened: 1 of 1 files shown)


alexfornuto:
+1

PR #3375, opened 2018-02-14
Add section on expiring cache on static assets

### LINKED ISSUE #3108: Long cache lifetime and correct cache expiry, for static assets
Priority (Low‚ Medium‚ High):

## Description

How to manage assets with long cache lifetime / correctly do cache expiry.

Pantheon sets a max-age header of 366 days on static assets. These can be expired from Pantheon Edge, but an end users browser / a CDN in front of Pantheon will cache the asset for a year (cannot be expired).
Therefore it is needed to change the name of a static file when updated. This is somewhat impractical for assets like pdf's, favicons, or even images updated without changing the name, so could be good to document.

See BUGS-1669 for context and a list of some of the tricky scenarios that might need an effort / workaround to handle.

## Path

xq42:
Closes #3108 

## Effect
- Add new section on handling static asset caching

## Remaining Work
- [ ] Add more strategies for assets whose file names cannot simply be changed
- [ ]

### DIFF
--- a/source/_docs/caching-advanced-topics.md
+++ b/source/_docs/caching-advanced-topics.md
@@ -36,7 +36,7 @@ If your site or application requires Facebook authentication, we have added exce
 
 ## Manually Expiring Cache for Static Assets (e.g. CSS, JS, PDFs, Images)
 
-Pantheon sets a max-age header of 366 days on static assets. While these can be expired from the Pantheon Edge, an end user's browser or a CDN in front of Pantheon will cache the asset for a year - something neither site owners, site developers, or Pantheon can control.
+Pantheon sets a max-age header of 366 days on static assets. While these can be expired from the Pantheon Edge, an end user's browser or a CDN in front of Pantheon will cache the asset for a year - something neither site owners, site developers, nor Pantheon Support can control.
 
 The simplest method for dealing with this is to ensure any new version of a static asset has a new name. This will force the user's browser to grab the new version of the asset.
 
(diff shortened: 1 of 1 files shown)


source/_docs/caching-advanced-topics.md
```diff
@@ -34,6 +34,12 @@ For more information, see [PANTHEON_STRIPPED GET Parameter Values](/docs/pantheo
 
 If your site or application requires Facebook authentication, we have added exceptions for this to allow users to register and log in. In the event you are having problems with another external authentication service, please contact us and let us know what service you are having issues with.
 
+## Manually Expiring Cache for Static Assets (e.g. CSS, JS, PDFs, Images)
+
+Pantheon sets a max-age header of 366 days on static assets. While these can be expired from the Pantheon Edge, an end user's browser or a CDN in front of Pantheon will cache the asset for a year - something neither site owners, site developers, or Pantheon can control.
```
alexdicianu: > something neither site owners, site developers, or Pantheon can control.
We've actually set this long expiration period from the VCL. We should remove the Pantheon part.

alexfornuto: @alexdicianu I think the point is that we can't control the behavior of the browser.

alexfornuto:
+1 Thanks @xq42. Unfortunately no SMEs contributed any additional strategies, but we won't let that continue to be a blocker. If you think it's important that we still do that step, please make a new issue for it.

PR #3381, opened 2018-02-15
Remove global_redirect

### LINKED ISSUE #3379: Evaluate Removal of Global Redirect from Modules and Plugins with Known Issues
Re: Modules and Plugins with Known Issues (Global Redirect)
## Issue Description
**Issue**: Too many redirects error when site is in maintenance mode.
**Solution**: Ensure that the "Frontpage Redirect Handler" is not checked in the Global Redirect administration page. Alternatively, [apply a patch to the module](https://www.drupal.org/node/1399024) to correct the issue.

## Suggested Resolution
Global Redirect should be able to be removed from the list of Modules and Plugins with Known Issues as [Issue 1399024](https://www.drupal.org/project/globalredirect/issues/1399024), "d7 version uses incorrect variable name for maintenance mode", was fixed in 7.x-1.6 that was released today.

https://www.drupal.org/project/globalredirect/releases/7.x-1.6

slashzero:
fixed in latest release

Closes #3379 

## Effect
- Remove Global Redirect section


alexfornuto:
+1 thanks for testing @slashzero, and thanks for reporting @cityofoaksdesign

PR #3382, opened 2018-02-16
Start State Scope

thomas-thackery:
Mention that start states outside of WordPress, Drupal 8, Drupal 7 are outside of platform support scope.

## Effect
- Statement of scope related to start states


ttrowell:
I might actually make a further distinction- 

"Custom upstreams and distributions not maintained by Pantheon..."

slashzero:
I agree with @ttrowell  - we need to reinforce custom upstreams aren't maintained by Pantheon.

alexfornuto:
Thanks @ttrowell / @slashzero, I added custom upstreams to the note.

PR #3384, opened 2018-02-16
Scope for Public Distributions

thomas-thackery:
## Effect
- Sets scope for Public Distributions 

## Remaining Work
- [x] Verify limitation of scope "please be aware that we cannot provide platform support for it."


alexfornuto:
+1 Thanks @thomas-thackery!

PR #3369, opened 2018-02-14
Remove Kalabox Doc.

dustinleblanc:
Closes #3365

PR #3385, opened 2018-02-16
Remove reference to Kalabox.

sarahg:
## Effect
- Remove reference to Kalabox.

## Remaining Work
- none


alexfornuto:
Looks like this change was accidentally duplicated (by me) in #3369

PR #3386, opened 2018-02-16
Fix initial Git tag name.

sarahg:
## Effect
- Updates name of Git tag used when a new site is created.

## Remaining Work
- [x] Technical review


alexfornuto:
+1 confirmed, thanks @sarahg!

PR #3388, opened 2018-02-16
Clear cache doesn't include Redis

thomas-thackery:
## Effect
- Mention that Clear Caches does not clear Redis cache


alexfornuto:
+1 Thanks @thomas-thackery!

PR #3389, opened 2018-02-16
[WIP]Start Composer conversion doc.

dustinleblanc:


source/_docs/guides/composer_convert.md
```diff
@@ -0,0 +1,118 @@
...
+```sh
+git checkout -b composify
+```
+
+## Create a New Project with Composer
+
+On your local workstation, from the site directory of your Pantheon site, execute the following command:
+
+```command
+cd .. && composer create-project pantheon-systems/example-drops-8-composer --stability=alpha my-site-composer
```
greg-1-anderson: I don't think `--stability=alpha` is needed any longer. This should be tested, though.

source/_docs/guides/composer_convert.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Checkout a New Branch
+We're about to make some massive changes to the codebase, it is wise to do this work on a branch as it might take you sometime to complete and rolling back can be complicated:
+```sh
+git checkout -b composify
+```
+
+## Create a New Project with Composer
+
+On your local workstation, from the site directory of your Pantheon site, execute the following command:
```
greg-1-anderson: `the site directory` is unclear. Do you mean the Drupal root, or `sites/default`? From context it seems that you mean the Drupal root.

source/_docs/guides/composer_convert.md
```diff
@@ -0,0 +1,118 @@
...
+#### Modules
+A Composer managed site should be able to include all custom code via Composer. Begin by reviewing your existing site's code. Check for contributed modules in `/modules`, `/modules/contrib`, `/sites/all/modules`, and `/sites/all/modules/contrib`. Add these modules to your new codebase using Composer by doing the following for each module in the `my-site-composer` directory:
+```sh
+composer require drupal/module_name
+``` 
+Where `module_name` is the machine name of the module in question
+
+#### Themes
+Repeat the same process with themes, checking `/themes`, `/themes/contrib`, `/sites/all/themes`, and `/sites/all/themes/contrib`. Use Composer in the same way as above to require these.
+
```
greg-1-anderson: Do we want to consider advising `https://www.drupal.org/project/composer_generate` or similar? Someone made a similar but better-maintained version of this, but I don't remember where it is, and drupal.org is 5xx-erroring right now so I can't search for it.

source/_docs/guides/composer_convert.md
```diff
@@ -0,0 +1,118 @@
...
+#### Libraries
+Libraries can be handled in the same way, the specifics dependent on how your library code was included in the source site. If using the Libraries API, you may have to do additional work to ensure the libraries function properly.
+
+### Custom Code
+#### Modules and Themes
+
+Custom code should be manually copied from the existing site repository.
```
greg-1-anderson: "from the existing site repository to the new site's repository"

source/_docs/guides/composer_convert.md
```diff
@@ -0,0 +1,118 @@
...
+```
+theme:
+```sh
+me@my-computer:~/Sites/my-site(composify) $ cp -r /themes/custom/great_theme ../my-site-composer/web/themes/custom
+```
+
+Follow suit with any other custom code you need to carry over.
+
+#### Settings.php
+Your existing site may have customizations to `settings.php` or any other config files. Review these carefully and extract relavent changes from these files to copy over, always review any file paths referenced in the code as these may change in the transition. It is not wise to completely overwrite the file with the old one as there are customizations for moving the configuration directory you don't want to overwrite.
```
greg-1-anderson: Is this guide always going to be used to go from a Pantheon site to a Composer-managed Pantheon site? We might want to explicitly mention that there should be no `$databases` array if we consider that some might start with a non-Pantheon site.

source/_docs/guides/composer_convert.md
```diff
@@ -0,0 +1,118 @@
...
+```
+my-site-composer
+|-web
+|-config    <--Here!
+|-vendor
+|-composer.json
+|-etc...
+```
+
+Locate the configuration files in your existing site and move them here. If they are stored in the files directory on your existing site, you will want to retreive them via SFTP (as the Git clone would not contain them).
```
greg-1-anderson: Perhaps mention somewhere that the example project is already configured to put the config in this location.

source/_docs/guides/composer_convert.md
```diff
@@ -0,0 +1,118 @@
...
+cd ../my-site-composer
+git add .
+git commit -m "Convert to Composer Based install
+```
+
+You should see a large amount of files committed to the new branch we created earlier.
+
+## Deploy &#128640;
+
+Now that we've committed the code on a branch, if the site has multi-dev, you can deploy that branch directly to a new multidev and test the site in the browser. If there are any issues, utilize your site's logs via `terminus drush my-site.composify -- wd-show` to inspect the watchdog logs, or follow the directions on (our documentation on log collect)[/docs/logs]
```
greg-1-anderson: There's a caution here. If the original site did not have a relocated document root (and it almost certainly will not), then creating a multidev site from an existing branch (as created and pushed above) with a pantheon.yml that DOES contain a relocated document root (as the example project does), then a bug will prevent the docroot from being relocated. The user should create the multidev first, and then push the new code with the new pantheon.yml + relocated docroot as a second step. If that is done, then the docroot will relocate correctly.

If the user ends up with a broken site (created a multidev with an existing commit, so docroot did not move), then adding a change to pantheon.yml (e.g. add a comment) and pushing it to the multidev will fix things up.

source/_docs/guides/composer_convert.md
greg-1-anderson: `composer_generate` is a Drush extension; it should be installed to somewhere like `~/.drush`

Snugug:
An initial thought:

It's possible that someone has followed the existing documentation before getting to this documentation. Given that the existing documentation and this documentation differ greatly, including instructions for existing projects that have done that transition already and that haven't is IMO important

source/_docs/guides/composer_convert.md
dustinleblanc: punted on this to GSD

source/_docs/guides/composer_convert.md
dustinleblanc: Addressed above in new revision

alexfornuto:
Tested! Technical +1

greg-1-anderson:
LGTM

dustinleblanc:
@alexfornuto awesome job getting this into shape, just read it over, magnificent sir!

ElijahLynn:
For anyone who finds this, published docs are at https://pantheon.io/docs/guides/composer-convert/

PR #3390, opened 2018-02-16
Include "git add" with steps to commit script.

sarahg:
## Effect
- Include "git add" step with commands needed to commit Quicksilver script.

## Remaining Work
- copy review


alexfornuto:
Nice catch @sarahg!

PR #3402, opened 2018-02-21
Update HR class for HTML 5

alexfornuto:
## Effect
- We've been using `<hr>` tags to create horizontal lines in our docs. As of HTML 5, `<hr>` no longer implicitly creates this line, but rather represents a thematic break. From [w3schools.com](https://www.w3schools.com/tags/tag_hr.asp):

    > All the layout attributes are removed in HTML5. Use CSS instead.

- This PR adds the required CSS for `<hr>` tags to behave as expected.

## Remaining Work
- [x] Review from either @rachelwhitton or @allenfear


alexfornuto:
Reviewed by @allenfear over slack.

PR #3427, opened 2018-02-26
Assessing Alerts Partial File

### LINKED ISSUE #3396: Pingdom Uptime Check Doc Update
Re: [Pingdom Uptime Check](https://pantheon.io/docs/guides/pingdom-uptime-check/)

Priority (Low‚ Medium‚ High):

## Issue Description:

We frequently get tickets from customers that got an alert from New Relic, Pingdom, an internal monitoring system etc.
These often merely state that customer received xx number of alerts, asking if there is an issue.
Pantheon support has very little to go on here. We rarely have the time to spend an hour digging through the site, unless there is an issue.

Please see internal reference:
Ticket 99555

## Goal
It would be great if we could help customers learn how to assess alerts. Ideally monitoring should help gain peace of mind, rather than additional worry.

## Suggested Resolution 

Could we add a section in each for something like:
- What do alerts mean
- What to do when you get an alert

### LINKED ISSUE #3397: Incident Management Doc Update
Re: [Incident Management](https://pantheon.io/docs/guides/pagerduty/monitor/)

Re: [Pingdom Uptime Check](https://pantheon.io/docs/guides/pingdom-uptime-check/)

Priority (Low‚ Medium‚ High):

## Issue Description:

We frequently get tickets from customers that got an alert from New Relic, Pingdom, an internal monitoring system etc.
These often merely state that customer received xx number of alerts, asking if there is an issue.
Pantheon support has very little to go on here. We rarely have the time to spend an hour digging through the site, unless there is an issue.

Please see internal reference:
Ticket 99555

Also see identical ticket for Pingdom: #3396 

## Goal
It would be great if we could help customers learn how to assess alerts. Ideally monitoring should help gain peace of mind, rather than additional worry.

## Suggested Resolution 

Could we add a section in each for something like:
- What do alerts mean
- What to do when you get an alert

alexfornuto:
Closes #3396 
Closes #3397

## Effect
- Creates partial file on assessing alerts
- Adds said partial to Incident Response and Pingdom guides

## Remaining Work
- [ ] @pantheon-systems/cse review (ie @tinefin )
- [ ] Copy Review

### DIFF
--- a/source/_docs/debug-slow-performance.md
+++ b/source/_docs/debug-slow-performance.md
@@ -127,17 +127,6 @@ If your site is already optimized to the best of your knowledge, including elimi
 
 If you've reached this point, it's probably time to consider upgrading your [Pantheon plan](/docs/select-plan/). We have a number of self-service options for scaling to your needs, but if you've already maxed out a self-service plan, then [Elite](https://pantheon.io/pricing#elite) is a good option.
 
-## Site Monitoring Alerts
-Services like Pingdom periodically reach out to sites with ping or HTTP requests to gauge availability. These types of tests can create false positives for a number of reasons, including:
-
-### Internet Routing
-A temporary routing issue between a particular monitoring service endpoint and one of our CDN Points of Presence (POPs) can result in an alert. The internet is a big place, and there's a lot of potential for intermittent packet loss between any two points, depending in the route taken. A single loss of access from one point to another does not always correlate to an issue with the site itself.
-
-Tools like [MTR](https://en.wikipedia.org/wiki/MTR_(software)){.external} can help to diagnose a routing issue. 
-
-### Idle Containers
-As we discuss in our doc [All About Application Containers](/docs/application-containers/#idle-containers), containers will spin down after approximately twelve hours of idle time (for a live environment on a paid plan). A web request will trigger a "wake up" on the environment, which usually occurs within 30 seconds. Monitoring pings sent between the initial request and the completed spin-up may fail, but do not necessarily indicate an issue.
-
 ## Running Cron Over Web Traffic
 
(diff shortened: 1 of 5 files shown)


alexfornuto:
I've reached out to @tinefin to review the latest version of this PR.

tinefin:
Would be good to include the points in the original issue(s).
This is described in internal ticket reference + PM with Alex F.

alexfornuto:
update: @tinefin is writing up some additional info to add to this PR.

tinefin:
# Monitoring alerts
Remember, these are automated alerts. The monitoring tool can't actually know whether the site is up or down. It merely checks if the site responds within xx seconds and if not, sends out an alert.
Some level of false positives should be expected.
The alert could just mean that this once, the site responded a bit slower than the xx seconds time limit. Or it could mean that the site is literally down.

If alerts are too "noisy" and generates many false positives, check if you can adjust the alert threshold or email preferences. Acceptable response time on sites vary and might not fit the default alert threshold.
Some monitoring tools also the option only alert on a number of consequtive failures. This helps false positives caused by a single request being slow.

## What to do when receiving monitoring alerts
Don't panic! This could just be internet winds, a few minutes of heavy load - or something else =)
As these are automated alerts, it is necessary to first verify that there actually is an issue.

All-the-things can be monitored. For websites, 2 popular parameters to monitor are whether the site responds and how quickly the site responds.

### Verify an alert for whether the site responds (uptime):
- Load the site in a browser and check if it loads.
- Login to the site.
If both are ok, the site is up. 

In case the site doesn't load, check whether the site is up outside your local network. Tools like these can be helpful:
https://downforeveryoneorjustme.com
https://geopeeker.com/fetch/?url=pantheon.io

### Verify an alert for site response time:
Application response time can be checked in New Relic. If response time is spiking compared to historically normal response time, dig into New Relic traces and site logs to find out why.
This is where to find the site logs on Pantheon:
https://pantheon.io/docs/logs/

Pantheon support can assist in case there is a verified issue on a site.

PR #3429, opened 2018-02-26
Correct -L flag in comment

acobster:
According to `man rsync`, there's `-l, --links` (copy symlinks as symlinks) and there's `-L, --copy-links` (transform symlink into reference file/dir).

Closes ...well, there's a sense of closure.

## Effect
- Literally a one-character change in the docs

## Remaining Work
- Gosh, I hope not!

:smile:

### DIFF
--- a/source/_docs/rsync-and-sftp.md
+++ b/source/_docs/rsync-and-sftp.md
@@ -56,7 +56,7 @@ Substitute your target environment and site UUID to connect; copying/pasting thi
 
 
     # -r: Recurse into subdirectories
-    # -l: Check links
+    # -L: Check links
     # -v: Verbose output
     # -z: Compress during transfer
     # Other rsync flags may or may not be supported


alexfornuto:
@acobster first of all, kudos for not only using the template, for for using it amusingly.

I think in this case we want to stick with `-l` because we want it to keep symbolic links and not create hard copies, but I think the description here could definitely be clearer. I'll take a stab at that tomorrow.

acobster:
Thanks, yeah, it definitely threw me through a loop! I had to guess at what it was aiming for, but I went with `-L` because the sample commands already use that flag. Is there something specific to that command that means we want to do something different there?

alexfornuto:
@acobster so I looked into the history of this file to see which jerk changed that to a `-L` without updating the comments... it was me.

Now we're looking into some other options like `--copy-unsafe-links` that might be a better choice. The switch to `-L` is good to make sure you get files that aren't in the directory being transferred but symlinked, but if there are symlinks to files with _are_ also being copied, you would end up with duplicates.

PR #3445, opened 2018-03-06
Add Missing Notes on Symlink

albertcausing:
Modification in relation to this explanation workflow:

Switch to GIT mode
$ git clone yoursite
$ cd yoursite

We then remove the directory or backup first before we create our symlink, usually, this is an existing directory and creating symlink will fail or will be created inside that directory and will look similar to this wp-content/my_resource_directory/my_resource_directory
$ rm wp-content/my_resouce_directory or mv wp-content/my_resouce_directory ~/backup/

We use -sf (force) so even the target directory does not exist, the symlink will get created.
$ ln -sf  ../uploads/my_resource_directory wp-content/my_resouce_directory
$ git add .; git commit -m 'adding symlink'; git push origin master;

To complete the symlink, we have to create the target directory in DEV via SFTP, the plugin thinks the directory exists, hence, the plugin will not create the directory anymore. That makes our symlink incomplete since it does not have a target directory.
#Switch to SFTP mode and create wp-content/uploads/my_resource_directory

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed

### DIFF
--- a/source/_docs/assuming-write-access.md
+++ b/source/_docs/assuming-write-access.md
@@ -19,15 +19,17 @@ The best solution is to communicate with the maintainer of the module or plugin 
 
  ```bash
  # The first path will be used as the new file destination instead of whatever path the plugin assumed write access to
- ln -s ./wp-content/uploads/new-directory ./wp-content/path/plugin-expects-to-write-to
+ ln -sf ./wp-content/uploads/new-directory ./wp-content/path/plugin-expects-to-write-to
  # Stage your changes
  git add .
  ```
-
     <div class="alert alert-info">
     <h4 class="info">Note</h4>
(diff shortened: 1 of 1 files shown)


alexfornuto:
Hey @albertcausing, thanks for the contribution! You'll notice I make a lot of changes to this one:

 - I broke up each step into a distinct command.
 - I took some of your warnings and made them required steps, including a step to `mv` the directory we're replacing with the symlink.
 - Because of the above, I removed the `-f` flag. If they don't follow the previous step they will still end up with a nested symlink, but I feel like this is a safer method that avoids potential loss of data.

What do you think?

alexfornuto:
+1 from @albertcausing over slack.

PR #3446, opened 2018-03-06
Update resetting dashboard password instruction

rubelyn:
logout before visiting reset password link so you will not redirected to User Dashboard


rachelwhitton:
+1! Thanks for the PR @rubelyn! Much appreciated!

PR #3447, opened 2018-03-06
Additional databases

thomas-thackery:
Closes #

## Effect
- Fixed typo: persevered > preserved
- Using table prefixes as an alternative to using additional database with mention of scope and responsibility 

## Remaining Work
- [ ] Get technical review +1


alexfornuto:
Closing without merge after discussion with @thomas-thackery, in which we determined this as a bad idea to suggest.

PR #3453, opened 2018-03-07
troubleshoot custom upstream checks

alexfornuto:
## Effect
- New troubleshooting section on using the **Check now** button to get custom upstream updates.

## Remaining Work
- [ ] Review from @lauren-kelly 
- [ ] Copy Review


source/_docs/upstream-updates.md
```diff
@@ -160,6 +160,14 @@ It's important to relay the need for updating core to maintainers, even if you p
 
 Once you have communicated the issue, you can [manually apply updates from the command line](#apply-upstream-updates-manually-from-the-command-line-and-resolve-merge-conflicts).
 
+### Custom Upstream Updates Not Available
+
+If you know your site's Custom Upstream has updated code, but it's not visible on your Site Dashboard, click on **Check now**:
+
+![Check Now](/source/docs/assets/images/dashboard/check-for-updates.png)
+
+This will trigger a "Code Cache Clear" to verify that the Site Dashboard has fetched the most recent commit. Please note that even after the workflow compeltes, it may take up to a minute before updates appear on the dashboard.
```
rachelwhitton: What does "Code Cache Clear" refer to in this context? If it's a workflow, suggest leaving it out so it's not distracting. Maybe something like: 
> This triggers the Site Dashboard to look for new commits made in the Custom Upstream's remote repository.

rachelwhitton:
Suggest including additional info on troubleshooting / workarounds in the event this doesn't resolve the issue. For example, the token was revoked for a private repo - which requires an org admin to edit custom upstream settings with new auth creds. 

Are there any more scenarios like that we should include @lauren-kelly? Trying to think if we should tell them to test multiple sites to make sure the custom upstream is connected properly - but not sure what would cause this on a single site.

source/_docs/upstream-updates.md
```diff
@@ -160,6 +160,14 @@ It's important to relay the need for updating core to maintainers, even if you p
 
 Once you have communicated the issue, you can [manually apply updates from the command line](#apply-upstream-updates-manually-from-the-command-line-and-resolve-merge-conflicts).
 
+### Custom Upstream Updates Not Available
```
rachelwhitton: I like the subheader here, but it conflicts with  [line 156](https://github.com/pantheon-systems/documentation/pull/3453/files#diff-2d8862e1a7f9fb609a84a234b6181faeR156)

That line was a messy change leftover from when I scrubbed "public distributions" from the docs. Can you take a swing at renaming [line 156](https://github.com/pantheon-systems/documentation/pull/3453/files#diff-2d8862e1a7f9fb609a84a234b6181faeR156) @alexfornuto? It's not clear that these scenarios refer to two completely different products

source/_docs/upstream-updates.md
rachelwhitton: Or if it's a workflow and we want to keep it, suggest adding supporting context in a tooltip or something

alexfornuto:
@rachelwhitton: @lauren-kelly kindly provided us with a catch-all troubleshooting solution. I'm comfortable :shipit:ing it now, but since you requested the additional work I thought you might want to review.

PR #3455, opened 2018-03-07
Recognise secure session prefix in Drupal config

davidsparks:
When PHP session.cookie_secure is on, the session cookie is prefixed with 'SSESS' instead of 'SESS'.

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed


davidsparks:
Pleasure!

PR #3457, opened 2018-03-08
Markdown Tables

### LINKED ISSUE #3249: Replace Markdown Table for Add On Service
Re: [Apache Solr on Pantheon](https://pantheon.io/docs/solr/)

## Issue Description:
Set a standard for illustrating access to features based on service level
Improve readability 

## Suggested Resolution 
- [x] Convert to a bordered, responsive table in favor of existing markdown table
- [x] Use a tooltip in the **Supported** table header cell to clarify that this refers to all environments
- [x] Do the same for similar tables throughout, such as redis.md
- [x] Add template to the style guide with intended usage

alexfornuto:
Closes #3249

## Effect
- New CSS so markdown tables will use the styles from `table-bordered` and `table-responsive`
- Defined tables and tooltips in style guide
- Used tooltip and check marks in solr table
- Aligned markdown table in Solr guide (not required, but easier on my eyes)

## Remaining Work
- [x] Review from @rachelwhitton

### DIFF
--- a/source/_docs/php-errors.md
+++ b/source/_docs/php-errors.md
@@ -15,6 +15,7 @@ For more in-depth information, see [Error Handling and Logging](https://secure.p
 Each of the PHP errors are handled differently depending on the site environment. On Dev, they are shown directly to the user in the browser. On Test and Live, PHP errors are not displayed to users, but they'll still be logged. Notices and warnings are logged in the database logs if `db\_log` is enabled for Drupal. The PHP constants `WP_DEBUG` and `WP_DEBUG_LOG` can be enabled for WordPress to save errors to wp-content/debug.log. PHP errors are also logged on the application server at `logs/php-error.log`.
 
 Here's a breakdown of what errors are shown and where:
+
 <table>
 <thead>
 		<tr>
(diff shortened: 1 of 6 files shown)


source/_docs/php-errors.md
```diff
@@ -15,90 +15,90 @@ For more in-depth information, see [Error Handling and Logging](https://secure.p
 Each of the PHP errors are handled differently depending on the site environment. On Dev, they are shown directly to the user in the browser. On Test and Live, PHP errors are not displayed to users, but they'll still be logged. Notices and warnings are logged in the database logs if `db\_log` is enabled for Drupal. The PHP constants `WP_DEBUG` and `WP_DEBUG_LOG` can be enabled for WordPress to save errors to wp-content/debug.log. PHP errors are also logged on the application server at `logs/php-error.log`.
 
 Here's a breakdown of what errors are shown and where:
-<table>
```
alexfornuto: Note: The only changes here are Y => ✓ and N => " ". The entire block is red because I removed CRLFs.

rachelwhitton: Any reason not to go ahead and convert this to markdown while we're here? (defer to you on if converting all tables to markdown is in scope of this PR but seems like we could do it quick since this file is in the diff now.)

rachelwhitton: Any reason not to go ahead and convert this to markdown while we're here? (defer to you on if converting all tables to markdown is in scope of this PR but seems like we could do it quick since this file is in the diff now.)

source/_docs/solr.md
```diff
@@ -14,16 +14,15 @@ Apache Solr is a system for indexing and searching site content. Pantheon provid
...
-| ------------- |:-------------:|
-| Sandbox      | **Yes, for all environments** |
-| Personal      | No      |
-| Professional | **Yes, for all environments**      |
-| Business | **Yes, for all environments**      |
-| Elite | **Yes, for all environments**      |
-
-
+
+| Plans         | Supported <a rel="popover" data-proofer-ignore data-toggle="tooltip" data-html="true" data-title="Environments" data-content="Solr is enabled for all environments"><em class="fa fa-info-circle"></em></a> |
```
rachelwhitton: Let's be more specific in the second th - maybe **Solr Support** instead of Supported (keep the tooltip though)

source/_docs/style-guide.md
```diff
@@ -469,6 +469,35 @@ Use panels for extraneous but useful information such as troubleshooting section
 ```
 </div>
 
+## Tables
+
+Markdown tables are defined like this:
```
rachelwhitton: The goal of this entry should be to add a standard for our practice of noting feature access based on service level, so we need more context here. 

Can you revise this intro to describe our new standard including where on the page to place a feature access table (towards the top)?

source/_docs/style-guide.md
```diff
@@ -469,6 +469,35 @@ Use panels for extraneous but useful information such as troubleshooting section
...
+
+<div class="style-example" markdown="1">
```
rachelwhitton: I revised this table to refer to the corresponding HTML entity (e.g., table header, table data) but once we have the context improved in the intro.. it would be best to show an example of actual usage. Like what we do in the [error messages section](https://pantheon.io/docs/style-guide/#error-messages).

source/_docs/style-guide.md
```diff
@@ -469,6 +469,35 @@ Use panels for extraneous but useful information such as troubleshooting section
...
+```markdown
+| Table Header 1     | Table Header 2 |
+| ------------------ | -------------- |
+| Table Data 1.1     | Table Data 2.1 |
+| Table Data 1.2     | Table Data 2.2 |
+| Table Data 1.3     | Table Data 2.3 |
+```
+</div>
+
+## Tooltips
```
rachelwhitton: Can you make the same adjustments to the tooltips section? 
- Improved inro that describes the intended usage. (doesn't need to mention alternatives specifically, but after reading it should be clear when to use tooltips vs something else like a definition list)
- Align the style of the example with the rest of the page - [refer to this commit for example](https://github.com/pantheon-systems/documentation/pull/3457/commits/434c4dfb697978fdaa831e053299f69af15610ba#diff-135923fe09a22b7615f023b1270a8a6e).


Also, I suggest replacing the example tooltip here with an example showing how to include links. Such as [the tooltip here](https://pantheon.io/docs/guides/multisite/considerations/): 
screen shot 2018-03-13 at 1 26 10 pm

source/_docs/redis.md
```diff
@@ -19,15 +19,15 @@ Redis provides an alternative caching backend, taking that work off the database
 All plans except for a Personal plan can use Redis. Redis is available to Sandbox plans for developmental purposes, but Redis will not be available going live on a Personal plan.
 
 | Plans         | Supported
```
rachelwhitton: Suggest **Redis Support** instead of Supported header

PR #3458, opened 2018-03-09
No need to export, build-tools will cache

ari-gold:

### DIFF
--- a/source/_docs/guides/build-tools/02-create-project.md
+++ b/source/_docs/guides/build-tools/02-create-project.md
@@ -18,15 +18,7 @@ image: buildToolsGuide-thumb
 ---
 In this section we'll use the Terminus Build Tools Plugin to create a new Pantheon Site, a corresponding GitHub repository, and configure CircleCI to run tests.
 
-1. Create a [CircleCI personal API token](https://circleci.com/account/api){.external}
-
-2. Create a [GitHub personal access token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/)
-    <div class="alert alert-info">
-      <h4 class="info">Note</h4>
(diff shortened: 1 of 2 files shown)


ari-gold:
@greg-1-anderson

greg-1-anderson:
I think it would be good to retain the info about the GITHUB_TOKEN and CIRCLE_TOKEN, perhaps as an optional step.

ari-gold:
IF there are legitimate use cases for the extra steps I'm +1 on keeping, but maybe move down to an FAQ or something to streamline the "quickstart" . @rachelwhitton or @alexfornuto would you be able to pick this up?

greg-1-anderson:
The largest use-case is for folks using version 1.x of the Terminus Build Tools, which does not cache the credentials. I'm going to tag 2.0.0 stable shortly, though.

greg-1-anderson:
Oh nevermind I re-read my reminder above that we are not caching in 1.0, and I haven't tagged 2.0 stable yet. Nevermind. 👍

source/_docs/guides/build-tools/02-create-project.md
```diff
@@ -49,10 +30,29 @@ In this section we'll use the Terminus Build Tools Plugin to create a new Panthe
...
+
+    ```bash
+    export GITHUB_TOKEN=yourGitHubToken
+    export CIRCLE_TOKEN=yourCircleCIToken
+    ```
+
+    If you need to replace a token, navigate to your [project settings page in CircleCI](https://circleci.com/docs/2.0/env-vars/#adding-environment-variables-in-the-app){.external}.
+    </div>
+    </div>
+
```
rachelwhitton: Two drop down panels in row like this makes it hard to tell what we want to user to do

Let's move the token info out of the drop down and place it either 1) top of this page or 2) before you begin on the intro page -- just so we give the info on tokens before the user has to interact with them.

Issue #3462, opened 2018-03-09
Dig For Windows


alexfornuto:
Re: [Domains and Redirects](https://pantheon.io/docs/domains/)

Priority (Low‚ Medium‚ High):low

## Issue Description:
From Feedback:
>I don't have dig because I'm on a windows client

## Suggested Resolution
Add documentation here (any anywhere else dig is a requirement) that dig fow Windows can be [downloaded here](https://www.isc.org/downloads/)

**OR**

Instruct Windows users to use `nslookup`.

alexfornuto:
Resolved with #3464

PR #3464, opened 2018-03-12
Replace dig with whatsmydns.net

ari-gold:
Based on feedback in two separate doc status reports that users on Windows couldn't use dig

## Remaining Work
- [x] Copy review

### DIFF
--- a/source/_docs/domains.md
+++ b/source/_docs/domains.md
@@ -195,15 +195,15 @@ Replace `http://` in the site's database and configure your CMS to assume users 
 ### CNAME Record Workaround
 If your domain's DNS configuration relies on an existing MX or TXT record that intentionally disallows CNAME records, you'll need to use A and AAAA records to configure DNS for subdomains (e.g., `www.example.com`) instead of CNAMEs:
 
-1. Identify required A record value by querying the target environment's [platform domain](#platform-domains) (e.g., `live-example.pantheonsite.io`) using a free online tool: [https://www.whatsmydns.net/#A/{live-example.pantheonsite.io}](https://www.whatsmydns.net/#A/{live-example.pantheonsite.io}).
+1. Identify required A record value by querying the target environment's [platform domain](#platform-domains) (e.g., `live-example.pantheonsite.io`) using a free online tool: [https://www.whatsmydns.net/#A/live-example.pantheonsite.io](https://www.whatsmydns.net/#A/live-example.pantheonsite.io).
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
@ari-gold I'm not comfortable with our documentation depending on the availability and possibly changing status of an external resource like that. I think the better solution, which I've put into #3462 would be to either:

A) Offer a link to the Dig for Windows binary
B) Change the docs to use `nslookup` which is pretty universal to all OSes, including Windows. 

Right now I'm leaning towards B. Thoughts?

ari-gold:
@alexfornuto  I think for someone not comfortable with the command line that an easy to use web tool is the best option and that installing dig or using nslookup introduces a lot of unnecessary friction to an already multi-step process of going live. Yes, it's a dependency, but not one I'm concerned about, and if it becomes a problem, we'll always be able to update the one or two lines with another easy to use web tool.

alexfornuto:
> installing dig or using nslookup

nslookup is built in to Windows.

>  if it becomes a problem

I'd prefer it never became a problem.

source/_docs/domains.md
```diff
@@ -195,21 +195,15 @@ Replace `http://` in the site's database and configure your CMS to assume users
 ### CNAME Record Workaround
 If your domain's DNS configuration relies on an existing MX or TXT record that intentionally disallows CNAME records, you'll need to use A and AAAA records to configure DNS for subdomains (e.g., `www.example.com`) instead of CNAMEs:
 
-1. Identify required A record value by querying the target environment's [platform domain](#platform-domains) (e.g., `live-example.pantheonsite.io`) from the command line using dig:
+1. Identify required A record value by querying the target environment's [platform domain](#platform-domains) (e.g., `live-example.pantheonsite.io`) using a free online tool: [https://www.whatsmydns.net/#A/{live-example.pantheonsite.io}](https://www.whatsmydns.net/#A/{live-example.pantheonsite.io}).
```
rachelwhitton: Suggest removing `{` and `}` from the URL to reduce confusion around how to structure the URL. `live-example.pantheonsite.io` should be enough to signal example platform domain.

rachelwhitton:
fwiw I think this makes sense based on feedback received, but I could be convinced that dig or nslookup is needed for those who prefer to work from the command. Not saying we need to do one or the other, but for the web tool does feel more accessible.

alexfornuto:
@rachelwhitton edited based on feedback. Feel free to merge.

PR #3470, opened 2018-03-15
Update drupal-cron.md

xq42:
Add a note in the troubleshooting section that cron will not run on an idle application container.

## Effect
- Adds a "Note" in the troubleshooting section that cron will not run on an idle application container, with a link to the relevant section in the "All About Application Containers" doc

## Remaining Work
- [x] Formatting / editing for publication


rachelwhitton:
+1! Thanks for the PR @xq42, much appreciated!

PR #3471, opened 2018-03-15
Update ImageMagick version.

sarahg:
## Effect
- Updates ImageMagick version used on the platform.

This can be verified on a PHP info page, like this:
https://v71-php-info.pantheonsite.io

(or by running `convert --version` on an application server)

## Remaining Work
- [x] Technical review


alexfornuto:
Thanks @sarahg!

PR #3474, opened 2018-03-16
Clarify creating and calling settings.local.php

thomas-thackery:
## Effect
- Further clarify the file that needs to be created and then included

### DIFF
--- a/source/_docs/local-development.md
+++ b/source/_docs/local-development.md
@@ -178,7 +178,7 @@ This file is ignored by the `.gitignore` file  in [WordPress](https://github.com
 
 ### Drupal 7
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
@thomas-thackery is it not true that there's nothing requiring this file to be called `settings.local.php`? Couldn't a user call it whatever they want, so long as that's reflected in the code snippet?

alexfornuto:
Edited based on discussion with @thomas-thackery, will merge if he's happy with the edits.

thomas-thackery:
+1 and yes the file could be named anything

alexfornuto:
Thanks @thomas-thackery!

PR #3476, opened 2018-03-16
Update 05-redirects.md

xq42:
Add a note near the top of the page reminding users to wait until https has been successfully provisioned before deploying any code that will redirect traffic to https (based on a chat that simply followed the launch guide and did the things it recommended in order, without considering what each step would require to work).

## Effect
- Adds a "note" / reminder to be sure https has been provisioned before adding the code in the doc

## Remaining Work
- [ ] Editing / word-smithing...


alexfornuto:
+1 Thanks @xq42!

PR #3479, opened 2018-03-20
Edit description

allenfear:
Reference to Drupal seems misplaced since git revert isn't CMS-specific.


alexfornuto:
Thanks @allenfear!

PR #3489, opened 2018-03-21
WordPress Incorrect foreach loop

### LINKED ISSUE #3487: Environment-Specific Configuration for WordPress Sites Doc Update
Re: [Environment-Specific Configuration for WordPress Sites](https://pantheon.io/docs/environment-specific-config/)

Priority (Medium):

## Issue Description:
A customer followed the steps here and it seems to be no longer working. It may be because of the new version updates at core where it is no longer compatible.

## Suggested Resolution
Review and revamp if this docs is still applicable or deprecated.

eabquina:
Closes #3487 

Incorrect foreach loop, only shows or views the LAST array value
mu-config/site-config.php  (lessen one folder back)

## Effect
- New PHP code for loops, and use is_plugin_inactive
- duplicate mu-plugins folder
-

## Remaining Work
- [ ] Peer Review

### DIFF
--- a/source/_docs/environment-specific-config.md
+++ b/source/_docs/environment-specific-config.md
@@ -17,11 +17,11 @@ if ( in_array( $_ENV['PANTHEON_ENVIRONMENT'], array( 'test', 'live' ) ) && ! def
 endif;
 ```
 
-In the same way, you can conditionally set `'WP_DEBUG', true` based on the given Pantheon environment. For example, the following configuration enables WP_DEBUG for development environments (Dev and Multidevs), while disabling it on production environments (Test and Live): 
+In the same way, you can conditionally set `'WP_DEBUG', true` based on the given Pantheon environment. For example, the following configuration enables WP_DEBUG for development environments (Dev and Multidevs), while disabling it on production environments (Test and Live):
 
 ```php
 /**
- * For developers: WordPress debugging mode.
+ * WordPress debugging mode.
  *
  * Sets WP_DEBUG to true on if on a development environment.
  *
--- a/source/_docs/environment-specific-config.md
+++ b/source/_docs/environment-specific-config.md
@@ -26,7 +26,7 @@ In the same way, you can conditionally set `'WP_DEBUG', true` based on the given
  * Sets WP_DEBUG to true on if on a development environment.
  *
  */
-if (!defined('WP_DEBUG')) {
+if (!defined('WP_DEBUG') && isset($_ENV['PANTHEON_ENVIRONMENT'])) {
     if(in_array( $_ENV['PANTHEON_ENVIRONMENT'], array('test', 'live'))) {
       define('WP_DEBUG', false);
     } else {
--- a/source/_docs/environment-specific-config.md
+++ b/source/_docs/environment-specific-config.md
@@ -10,10 +10,30 @@ To quickly see which environment you are on, consider installing the [Pantheon H
 ## Customizations to wp-config.php
 
 Pantheon's default version of `wp-config.php` contains a nice example that sets `DISALLOW_FILE_MODS` to `true` on Test and Live environments.
-<script src="//gist-it.appspot.com/https://github.com/pantheon-systems/wordpress/blob/master/wp-config.php?footer=minimal&slice=88:91"></script>
+
+```php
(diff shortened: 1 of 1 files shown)


alexfornuto:
@eabquina who would you suggest for peer review?

eabquina:
@alexfornuto   sorry don't have an idea right now who.. but calling folks  @pantheon-systems/cse on your free time. Thanks! i'll mention it in the next huddle

alexfornuto:
@eabquina I've reached out to try to get an SME to review this.

source/_docs/environment-specific-config.md
```diff
@@ -9,11 +9,36 @@ To quickly see which environment you are on, consider installing the [Pantheon H
...
+if ( in_array( $_ENV['PANTHEON_ENVIRONMENT'], array( 'test', 'live' ) ) && ! defined( 'DISALLOW_FILE_MODS' ) ) :
+    define( 'DISALLOW_FILE_MODS', true );
+endif;
+```
+
+In the same way, you can conditionally set `'WP_DEBUG', true` based on the given Pantheon environment. For example, the following configuration enables WP_DEBUG for development environments (Dev and Multidevs), while disabling it on production environments (Test and Live): 
+
+```php
+/**
+ * For developers: WordPress debugging mode.
```
rachelwhitton: Let's remove `For developers:`

source/_docs/environment-specific-config.md
```diff
@@ -9,11 +9,36 @@ To quickly see which environment you are on, consider installing the [Pantheon H
...
+ * Sets WP_DEBUG to true on if on a development environment.
+ *
+ */
+if (!defined('WP_DEBUG')) {
+    if(in_array( $_ENV['PANTHEON_ENVIRONMENT'], array('test', 'live'))) {
+      define('WP_DEBUG', false);
+    } else {
+      define( 'WP_DEBUG', true );
+    }
+}
```
rachelwhitton: I'd like to get another technical review on this snippet to answer:  

1) Do we need to check the var is defined before calling it? Using `isset($_ENV['PANTHEON_ENVIRONMENT'])` -- otherwise this would cause a fatal error if they weren't on Pantheon (I think..) 
2) Do we need to check for command line condition, via `(php_sapi_name() != "cli")` ?

In context of the changes, I'm not sure if number 2 is a big deal but number 1 seems like a required standard we should adhere to any time we recommend conditional configurations like this

source/_docs/environment-specific-config.md
```diff
@@ -22,7 +47,7 @@ Filters or functions running in an mu-plugin can enable development plugins that
 
 ### Create the Plugin
 
-Copy this plugin file to `wp-content/mu-plugins/mu-plugins/site-config.php` and edit accordingly.
+Copy this plugin file to `wp-content/mu-plugins/site-config.php` and edit accordingly.
```
rachelwhitton: Not a blocker, just would like to understand this change -- is this a preference or are there broader implications of mu vs standard plugin?

source/_docs/environment-specific-config.md
alexfornuto: @dustinleblanc I'm tagging you in on this since this is mostly your handiwork.

The only other use of `isset($_ENV['PANTHEON_ENVIRONMENT'])` is in an `else` statement that happens when we know it is set, but we know that by checking to see if it's _not_ set (`!isset($_ENV['PANTHEON_ENVIRONMENT'])`).

source/_docs/environment-specific-config.md
alexfornuto: @rachelwhitton the issue here is that it was doubled, easy to miss.

source/_docs/environment-specific-config.md
dustinleblanc: @rachelwhitton @alexfornuto I think 29 as 
```php
if (!defined('WP_DEBUG') && isset($_ENV['PANTHEON_ENVIRONMENT'])) {
``` 
is probably the way to go?

alexfornuto: @dustinleblanc does moving the check for `PANTHEON_ENVIRONMENT` up into the main `if` statement protect from fatal errors if it isn't defined?

alexfornuto:
@rachelwhitton  Requested changes implemented.

PR #3515, opened 2018-03-28
Stage new files

ari-gold:


source/_docs/drupal-security-patches.md
```diff
@@ -88,6 +89,7 @@ You'll need a local git clone of your site code.
 
     ```bash
     rm 119.patch
+    git add includes/request-sanitizer.inc
```
greg-1-anderson: This is incomplete -- there are more files that need to be added. Can we just say `git add .` here?

ari-gold: the next line is `git commit -am`

source/_docs/drupal-security-patches.md
```diff
@@ -53,6 +53,7 @@ You'll need a local git clone of your site code.
 
     ```bash
     rm 206.patch
+    git add core/lib/Drupal/Core/Security/
```
greg-1-anderson: There are files in other locations that also need to be added. Can we just say `git add .` instead?

greg-1-anderson:
+1 LGTM.

Looks like the test failure was transient, perhaps caused by high load from folks updating.

PR #3518, opened 2018-03-29
Remove terminus auth:logout step for account switching.

sarahg:
## Effect
- Remove auth:logout step for account switching.

Running auth:logout deletes *all* your tokens (https://github.com/pantheon-systems/terminus/issues/1828) and isn't actually needed to switch between two accounts.

## Remaining Work
- [ ] technical review


alexfornuto:
+1 TY!

PR #3535, opened 2018-04-03
Amend -l flag

alexfornuto:
Closes #3429 

## Effect
- Amends `-l` with `--copy-unsafe-links`

## Remaining Work
- [x] Review from @acobster (optional)
- [ ] Technical Review
- [ ] Copy review.

### DIFF
--- a/source/_docs/rsync-and-sftp.md
+++ b/source/_docs/rsync-and-sftp.md
@@ -56,7 +56,7 @@ Substitute your target environment and site UUID to connect; copying/pasting thi
 
 
     # -r: Recurse into subdirectories
-    # -l: Check links
+    # -L: Check links
     # -v: Verbose output
     # -z: Compress during transfer
     # Other rsync flags may or may not be supported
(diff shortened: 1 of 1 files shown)


alexfornuto:
@acobster thoughts?

acobster:
Looks good! My only nitpick would be to replace the term "reference" with something more explicit, like "symlink target" [here](https://github.com/pantheon-systems/documentation/pull/3535/commits/73cfec87f42653dd1a8c01738fe96a78ff2b2e55#diff-3e0cf9c130f5fad65aa69aa17dbbf683R62), since "reference" could be construed as either the file being pointed to or the file doing the pointing. Whenever possible, I like being able to reason about `rsync` without thinking about [semiotics](https://en.wikipedia.org/wiki/Signified_and_signifier). ;)

alexfornuto:
Done, thanks for the feedback!

alexfornuto:
BTW for anyone interested, [this](https://gist.github.com/alexfornuto/a575b1a2a3492b632dfc94a5336a0af4) is how I confirmed that these flags will work as we expect them to.

alexfornuto:
Assigning to @lauren-kelly to confirm the behavior.

PR #3540, opened 2018-04-04
Add reference to search-replace troubleshoot doc

thomas-thackery:
## Effect
- Adds a link to the wp-cli search-replace troubleshooting guide. 

## Remaining Work
- [ ] Consider also referencing to the similar multidev guide at https://pantheon.io/docs/guides/multisite/workflows/

### DIFF
--- a/source/_docs/wp-cli.md
+++ b/source/_docs/wp-cli.md
@@ -43,7 +43,7 @@ The first part of the output is Terminus telling you which command it's running,
 
 Feeling comfortable with WP-CLI? Here are a [few of many commands](https://developer.wordpress.org/cli/commands/) you may find helpful in your journeys:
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
@thomas-thackery thanks for the submission! Can you please clarify what your item under *Remaining Works* means?

thomas-thackery:
Yes, there's a search-replace guide also on this page for Site Networks https://pantheon.io/docs/guides/multisite/workflows/. Was wondering if it would be worth it to mention it as well.

alexfornuto:
@thomas-thackery thanks for clarifying! I'm gonna go with "no" on that, since those instructions are site-network-specific. But I think the addition you made here is pretty great. Thanks again!

PR #3542, opened 2018-04-04
add-incommon

### LINKED ISSUE #3539: Establish best practices for dealing with Let's Encrypt rate-limiting issues
Re: https://pantheon.io/docs/https/#addressing-lets-encrypt-rate-limits
## Issue Description
Some primarily EDU client have mentioned facing "rate limiting" when trying to issue Let's Encrypt certificates. ref: https://letsencrypt.org/docs/rate-limits/

I believe that this occurs in cases where they are trying to generate certs for sites, not hosted on Pantheon, that share domains with sites that are on Pantheon. I can provide references to cases where clients have faced this issue, but to understand the underlying problem space, please refer to DTS, who has been involved with at least one of the instances.
 
## Suggested Resolution
Some solutions have bee proposed, but I believe may no longer be applicable:
- Requesting their own high-limit account with Let's Encrypt; But if I understand correctly, this may not help us, since if Pantheon+client together hit this cap, then we are just increasing the problem
- Applying for inclusion on the Public Suffix List: But the attempt to do so that I've followed appears as though it will be refused (more details can be provided internally).

Another solution proposed was to:
- Choose a different CA that supports ACME; For example, GlobalSign supports it. This would not be free though, so requires cost and procurement flexibility. Client could, however, switch to Let's Encrypt once/if one of the other solutions is in place. 

I believe that DTS is evaluating other possible solutions/scenarios.

allenfear:
Closes #3539 

## Effect
Addresses possible workaround for edu use case related to #3539

### DIFF
--- a/source/_docs/https.md
+++ b/source/_docs/https.md
@@ -114,10 +114,11 @@ The text challenge to pre-provision HTTPS on Pantheon requires adding a `.well-k
 
 Pantheon requests new certificates frequently in order to add domains to existing certificates. This can potentially expose  organizations managing many domains to Let's Encrypt rate limits. While sites hosted on Pantheon are not subject to these lower limits, sites hosted off the platform may experience request failures.
 
-If you encounter rate limits we recommend either of two approaches:
+If you encounter rate limits, we recommend the following approaches:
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
+1 Thanks @allenfear!

PR #3550, opened 2018-04-09
Update cookies.md

Jorenm:
The only time this crucial requirement is mentioned on this page it sounds as if it's talking about a specific use case for cookies related to serving cached content. I propose making it very prominent at the top that this is required, since any use of cookies seems to depend upon that naming scheme.

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed

## Post Launch
- [ ] Redirect `/docs/old-path/` => `/docs/new-path/`
- [ ] Include/exclude pages ^ respectively within docs search service provider

### DIFF
--- a/source/_docs/cookies.md
+++ b/source/_docs/cookies.md
@@ -5,6 +5,9 @@ tags: [cacheedge]
 categories: []
 ---
 
+## Cookie naming requirements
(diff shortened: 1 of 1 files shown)


alexfornuto:
Thanks @Jorenm!

PR #3571, opened 2018-04-16
Correct target not responding error code to 504.

dustinleblanc:


alexfornuto:
+1 Thanks @dustinleblanc!

Issue #3208, opened 2018-01-12
Redis log access


sarahg:
https://pantheon.io/docs/drupal-redis/#frequently-asked-questions

## Issue Description:
Under the FAQ section, we've got "If Redis hits the upper limit of memory usage, is this logged on Pantheon?" with directions to connect to the cacheserver via SFTP to view the log. The Dashboard does not provide SFTP connection info to Redis servers.

## Suggested Resolution 
Remove this section or suggest that users contact support for log data.

PR #3584, opened 2018-04-17
Remove outdated section from Redis FAQ

sarahg:
## Effect
- Removes section in Redis FAQ about retrieving `redis.log` via SFTP. SFTP access is not available to Redis servers.

(Similar to this PR on the original Drupal-specific Redis doc: https://github.com/pantheon-systems/documentation/issues/3208)

## Remaining Work
- none

### DIFF
--- a/source/_docs/redis.md
+++ b/source/_docs/redis.md
@@ -401,15 +401,5 @@ set-max-intset-entries 512
 activerehashing yes
 ```
 
-### If Redis hits the upper limit of memory usage, is this logged on Pantheon?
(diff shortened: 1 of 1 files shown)


alexfornuto:
So I did a little more testing, and discovered that this log file is actually accessible. The dashboard doesn't really describe it, but if you use the SFTP instructions combined with the IP address from the redis-cli instructions, you can get your redis.log file.

alexfornuto:
Thanks @sarahg!

PR #3586, opened 2018-04-18
stronger intros

alexfornuto:


alexfornuto:
@rachelwhitton I think my latest commit addresses all your concerns in your first comment.

Regarding definition lists, I'm afraid I'm not following. Would you be willing/able to fork from this branch and provide an example?

source/_docs/domains.md
```diff
@@ -93,7 +93,7 @@ It's a best practice for SEO and security to standardize all traffic on HTTPS an
     <div class="panel-inner" markdown="1">
 When using multiple snippets, be sure to step through the logic. This is particularly important when redirecting to a common domain while also incorporating redirects for specific pages. All <code>if</code> conditional statements need to be in the correct order. For example, a wholesale redirect executed <em>prior</em> to redirects for specific pages would likely prevent the second statement from being evaluated.
 
-#### Redirect to HTTPS
```
rachelwhitton: These need to stay H4s because the styling of H3 associates the section with a corresponding TOC entry 

Our use of info class is restricted to elements like note titles (warning, note, etc.) and drop down panel titles, unless we want that drop down panel to be included in the TOC as is the case in this doc. We want the panel to be anchored in this scenario for easy sharing. The headers contained within the drop down panel do not get corresponding TOC entries so they shouldn't be styled as such

alexfornuto:
@rachelwhitton I liked H3s for increased font size and weight, and thought that adding the info class would prevent them from showing up in the TOC, as H4s are. Happy either way.

PR #3598, opened 2018-04-23
D8composer no ci

### LINKED ISSUE #3251: New Guide: Managing Drupal Core and Contrib with Composer (and without an external repo)
Re: [Composer](https://pantheon.io/docs/composer/)

## Issue Description

Using Terminus Build Tools with an external repo (like GitHub) and a 3rd party CI service (like CircleCI) is more overhead than many Pantheon customers want while still using Composer with Drupal 8.

## Suggested Resolution

Write a guide documenting how to set up and use a copy[ example-drops-8-composer](https://github.com/pantheon-systems/example-drops-8-composer) without putting it on GitHub.

This issue came up in Power Users and I wrote an outline of what the guide could be. https://groups.google.com/a/pantheon.io/d/msg/power-users/Yf7su41Shrc/LGnkY2GHBQAJ

mcdwayne:
Closes #3251 

## Effect
New document for using Composer and Pantheon without using a CI cool like CircleCI.  All manual like.  

## Remaining Work
- [ ] If needed

## Docs Team Work
- [ ] Technical Review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/guides/drupal-8-composer-no-ci.md
+++ b/source/_docs/guides/drupal-8-composer-no-ci.md
@@ -11,9 +11,7 @@ contributors:
   - davidneedham
 ---
 
-In an ideal situation, you might ignore the vendor directory and use [an automated build step with continuous integration](https://pantheon.io/docs/guides/build-tools/) to install your site with composer on the fly. But sometimes your project doesn’t have the budget or time needed to justify this workflow.
-
-In this guide, we’re going to simplify things and run through the bare necessities to use [Composer](https://getcomposer.org/) to install your Drupal 8 site on your local machine and push that to Pantheon. We’ll then demonstrate how to add a module using the same process.
+In this guide, we’re going to run through the bare necessities to use [Composer](https://getcomposer.org/) for managing a Drupal 8 site on your local machine and pushing to Pantheon.
 
 Using a Composer managed site **removes** the ability to [apply Drupal core updates via the site dashboard](https://pantheon.io/docs/core-updates/). This is for advanced users who are comfortable taking complete responsibility for the management of site updates with Composer.
 


ataylorme:
Sorry for not thinking of this earlier. We should tell folks that adopting Composer, specifically using the non-default Drupal upstream, will mean they are in charge of Drupal updates via Composer (e.g. no more dashboard updates).

greg-1-anderson:
This document recommends using the [drops-8-composer](https://github.com/pantheon-systems/drops-8-composer) repo, which is obsolete and unmaintained.  I [created some issues](https://github.com/pantheon-systems/drops-8-composer/issues) that should be considered before reverting this repo to supported status. We should also consider what our update plan is to keep [example-drops-8-composer](https://github.com/pantheon-systems/example-drops-8-composer) in sync with this project. We should have product buy-in before adding another upstream to maintain.

I think that the simplest thing to do would be to recommend that folks use [example-drops-8-composer](https://github.com/pantheon-systems/example-drops-8-composer) as their starting point. It may have some directories (`tests`, `.circleci` etc.) that the user will simply need to ignore, but the advantage here is that if a user thinks they do not want a build step when they start, and later change their mind, then it might be easier to add build + test if they started with example-drops-8-composer instead of some other repo.

Another alternative to consider would be to start the user with just `drops-8`, and explain how to use and update this repo using Composer. This means that the user would not have a relocated document root, which would be a disadvantage; also, the user might have to go to a little extra effort to update their site every minor update. Using drops-8 directly means the user would have to update core with Drush or the Dashboard, not Composer, so this option might not meet the goals of this guide. On the upside, though, it wouldn't require maintaining a new upstream.

ataylorme:
@mcdwayne I updated the guide to use `example-drops-8-composer` at @greg-1-anderson's request since `drops-8-composer` is not maintained. I ran through all the steps again. If you or @davidneedham could test as well that would be great.

@greg-1-anderson mind giving the updates guide a review?

source/_docs/guides/drupal-8-composer-no-ci.md
```diff
@@ -0,0 +1,194 @@
...
+
+* Remove all dependencies in the `require-dev` section
+* Remove the `lint` sub-section of the `scripts` section
+* Remove the `code-sniff` sub-section of the `scripts` section
+* Remove the `unit-test` sub-section of the `scripts` section
+* Remove the `"find .circleci/scripts/pantheon/ -type f | xargs chmod 755",` line of the `post-update-cmd` sub-section of the `scripts` section
+* Remove the `"find tests/scripts/ -type f | xargs chmod 755"` line of the `post-update-cmd` sub-section of the `scripts` section
+* Remove the comma at the end of the `"DrupalProject\\composer\\ScriptHandler::createRequiredFiles"` line of the `post-update-cmd` sub-section of the `scripts` section to make the JSON valid
+
+The `scripts` section should now be:
```
ataylorme: @mcdwayne added a snippet of what the `scripts` section should look like after the modifications. Does this help?

ataylorme:
@alexfornuto can you help fix the image errors in the CircleCI build?

alexfornuto:
@ataylorme https://github.com/pantheon-systems/documentation/pull/3598/commits/220dfd9d3127ead756664e50adb7ebb21a1b0022 should resolve.

ataylorme:
> I know it is 'yet another thing to maintain' but if we are going to encourage people down this path we need to ether

I think we have a fourth option of leaving the automated testing components in place. This would mean no modifying `composer.json` (until address module is added). The negative effect would be downloading and committing additional, likely unused, code for the test suites. While my guess is folks who don't want CI probably aren't going to do any testing on the site either.

@mcdwayne @davidneedham what are your thoughts on leaving PHP Unit, etc. in the `composer.json` and not asking the user to modify it at all? I'm sort of torn. It is easier for them initially to set things up but I also feel if you are using Composer then being comfortable modifying the `composer.json` isn't too much of an ask.

ataylorme:
Correct link for review is https://d8composer-static-docs.pantheonsite.io/docs/drupal-8-composer-no-ci/

davidneedham:
> @mcdwayne @davidneedham what are your thoughts on leaving PHP Unit, etc. in the composer.json and not asking the user to modify it at all? I'm sort of torn. It is easier for them initially to set things up but I also feel if you are using Composer then being comfortable modifying the composer.json isn't too much of an ask.

That seems reasonable to me. Making it easy to start is important here. Maybe we add a section at the end of the lesson mentioning that they can make that edit if they don't expect to use it?

alexfornuto:
Not sure who to tag on this, but feel free to assign me and remove the WIP label when this is ready for Docs :eyes:

ataylorme:
> Maybe we add a section at the end of the lesson mentioning that they can make that edit if they don't expect to use it?

The steps to do that are what @mcdwayne thought was too much.

alexfornuto:
>> Maybe we add a section at the end of the lesson mentioning that they can make that edit if they don't expect to use it?
>
>The steps to do that are what @mcdwayne thought was too much. 

Sounds like a perfect use for a [panel](https://pantheon.io/docs/style-guide/#panels)

davidneedham:
Exactly.

alexfornuto:
@ataylorme Yup, that's a fine looking panel!

ataylorme:
@davidneedham a swell suggestion from Power Users

> My main suggestion would be to add a section at the end (you could do a dropdown like you did for removing the tests or another page) about what do if there is an update (drupal update, module update or theme change are the three main ones I usually do).  Does the user need to do the change locally or can it be done with terminus (I assume local).  That should be explicit.

ataylorme:
@davidneedham can you give this a final review before we pass off to docs?

ataylorme:
> Add Andrew to the authors
I don't know my docs author name. Both `andrew` and `andrewtaylor` in the markdown front matter did nothing. Maybe I don't have one? @alexfornuto can you look into that?

ataylorme:
I've addressed David's feedback. @davidneedham and @mcdwayne can either (or both) of you give this a final review?

ataylorme:
> I don't know my docs author name. Both andrew and andrewtaylor in the markdown front matter did nothing. Maybe I don't have one? @alexfornuto can you look into that?

Cancel that, found it on the [contributor archive](https://d8composer-static-docs.pantheonsite.io/docs/contributors/).

mcdwayne:
Error occurs when you export a site name with any capital letters.  such as 
`export PANTHEON_SITE_NAME="d8-composer-no-CI-5-3-2018"`
site create auto converts to lower case,  BUT This later causes issues for commands like as `terminus connection:info` is case sensitive.  

Suggestion: Add a reminder that the site name should be all lower case.

mcdwayne:
On line 67 it uses a uri for the remote repo that is quite frankly hard to notice it is not a real URI if you are skimming.  
Suggestion
Either
1.add a step to export the git uri as SITE-GIT-URI, then mod code to be `git remote set-url origin $SITE-GIT-URI`
or
2. mod the line to be `git remote set-url origin <git URI from previous step>`

davidneedham:
More feedback:

1. Do we need this text if we immediately go on to add the variable?  "Note this doc uses the $PANTHEON_SITE_NAME site slug. This will need to be replaced with your unique site slug." I suggest we remove the text and add a small note where the variable is being added. Maybe see what we've said in other guides where we do this.
2. I agree with Dwayne about the git URL. This works, but I don't know if it's the best:
```export SITE_GIT_URL=`terminus connection:info $PANTHEON_SITE_NAME.dev --field=git_url`
git remote set-url origin $SITE_GIT_URL```
3. .gitignore file section: I don't know if the important heading is necessary, but I suggest breaking this out.
 - Explain what the problem is after we run the first git status. (That too many things are being ignored. We're not seeing the web folder).
 - Then keep most of the paragraph the same. Edit the .gitignore to remove everything CUT and above.
 - Line break before "Now let's run git status again to make sure everything is included."
4. When I do the terminus commit in the dashboard it says "There is no code to commit." Weird. Maybe there's something particular to this way of installing Drupal that doesn't require a change to settings.php?

**Comment by ataylorme** [comment]
> gitignore file section: I don't know if the important heading is necessary, but I suggest breaking this out.

@davidneedham can you clarify what you mean here? Or better yet just update this section directly with what you hope to see.

**Comment by ataylorme** [comment]
> When I do the terminus commit in the dashboard it says "There is no code to commit." Weird. Maybe there's something particular to this way of installing Drupal that doesn't require a change to settings.php?

@davidneedham  I added this due to your earlier feedback. Would you like to remove the section? We could also add a `--force` to make a commit, though it may be empty and doesn't seem like the best solution.

File: source/_docs/guides/drupal-8-composer-no-ci.md [hunk 2]
```diff
@@ -0,0 +1,242 @@
...
+        },
+    ```
+  </div>
+</div>
+
+## Managing Drupal with Composer
+
+### Downloading Drupal Dependencies with Composer
+
+Normally the next step would be going through the standard Drupal installation. But since we’re using Composer, none of the core files exist yet. Let’s use Composer to install Drupal core.
```
stevector: Can we change this to `Let’s use Composer to download Drupal core.` since `installation` was used in the previous sentence to mean the database operation of installing core and here we mean downloading files.

source/_docs/guides/drupal-8-composer-no-ci.md
```diff
@@ -0,0 +1,242 @@
...
+
+Since we modified `composer.json` we will need to update Composer to download the new dependencies.
```
stevector: Are there new dependencies at this point?

source/_docs/guides/drupal-8-composer-no-ci.md
```diff
@@ -0,0 +1,240 @@
...
+
+![image of git status showing the changed files in red](/source/docs/assets/images/guides/drupal-8-composer-no-ci/drops-8-composer-git-status-after-installing-d8.png)
+
+Set the site to git mode.
+
+```
+terminus connection:set $PANTHEON_SITE_NAME.dev git
+```
+
+Add and commit the code files. A git force push is necessary initially but subsequent pushes should not need `--force`.
```
stevector: Let's add an explanation for why. 

> A git force push is necessary because we are writing over the empty repository on Pantheon with our new history that was started on the local machine. Subsequent pushes should not use ` --force`.

source/_docs/guides/drupal-8-composer-no-ci.md
```diff
@@ -0,0 +1,240 @@
...
+
+```
+git add .
+
+git commit -m 'Drupal 8 and dependencies'
+
+git push --force
+```
+
+**Note** the `vendor` directory is being committed to Pantheon. This is because Pantheon needs the full site artifact. If you prefer to ignore the `vendor` directory then take a look at our documentation on the more advanced automated workflow with a build step.
```
stevector: We should link to that documentation.

source/_docs/guides/drupal-8-composer-no-ci.md
```diff
@@ -0,0 +1,236 @@
...
+terminus drush $PANTHEON_SITE_NAME.dev -- site-install -y
+```
+
+Log in to your new Drupal 8 site to verify it is working. You can get a one-time login link using Drush.
+
+```
+terminus drush $PANTHEON_SITE_NAME.dev -- uli
+```
+
+**Note:** there is no need to commit changes as the `example-drops-8-composer` project that we started with has already defined `$settings['install_profile'] = 'standard';` in `settings.php`.
```
stevector: Do we need to the switch back and forth at all? I was just able to do a `drush si` on a site in git mode that had the `settings.php` piece in place.

source/_docs/guides/drupal-8-composer-no-ci.md
```diff
@@ -0,0 +1,236 @@
...
+categories: [drupal]
+type: guide
+permalink: docs/guides/:basename/
+contributors:
+  - ataylorme
+  - dwayne
+  - davidneedham
+---
+
+In an ideal situation, you might ignore the vendor directory and use [an automated build step with continuous integration](https://pantheon.io/docs/guides/build-tools/) to install your site with composer on the fly. But sometimes your project doesn’t have the budget or time needed to justify this workflow.
```
stevector: I think this paragraph would benefit from a little more exposition. It starts with assuming knowledge of a vendor directory and Composer. Do we need to explain what those are?

ataylorme: Perhaps we remove the first paragraph altogether and rework the second paragraph to be the into.

> In this guide, we’re going to run through the bare necessities to use [Composer](https://getcomposer.org/) for managing a Drupal 8 site on your local machine and pushing to Pantheon. 

We talk about the `vendor` directory later with

> **Note** the `vendor` directory is being committed to Pantheon. This is because Pantheon needs the full site artifact. If you prefer to ignore the `vendor` directory then take a look at our build tools guide for documentation on the more advanced automated workflow with a build step.

davidneedham: I agree. That is much clearer.

ataylorme:
@alexfornuto this is good to go on our side and ready for you to review/merge.

sarahg:
In the "Removing Automation Pieces" section, maybe we should recommend that people remove the Quicksilver hooks from pantheon.yml as well. If you leave those in place, some PHP notices get logged on sync_code workflows:

[11-May-2018 22:10:28 UTC] PHP Notice:  Undefined index: token in /srv/bindings/5af1c9fe92c34521a1666543cdd65dca/code/web/private/scripts/quicksilver/quicksilver-pushback/push-back-to-github.php on line 38

alexfornuto:
Interesting feedback came in [on the issue](https://github.com/pantheon-systems/documentation/issues/3251#issuecomment-389995503).

PR #3615, opened 2018-04-25
serialize doc and repeat for newest patches

alexfornuto:
Closes #NA

## Effect
- Serializes the Drupal Security Patch doc so that it can be amended
- Amended with patches for SA-CORE-2018-004

## Docs Team Work
- [x] Technical Review (@greg-1-anderson and/or @ari-gold)
- [ ] Copy Review (@rachelwhitton)

## Post Launch
- [ ] Update Status Report


alexfornuto:
Update to remove from archive in Waffle

PR #3622, opened 2018-04-26
Make the load testing steps CMS-agnostic

sarahg:
## Effect
- Remove the references to Drupal when suggesting load testers run cron and clear caches — these would apply to WP, too.

## Remaining Work
- n/a

## Docs Team Work
- [ ] Review


alexfornuto:
Nice catch, thanks @sarahg!

PR #3625, opened 2018-04-27
site plans update

allenfear:
## Effect
Accounts for the fact that Basic sites do not include New Relic.


alexfornuto:
Thanks @allenfear!

PR #3635, opened 2018-04-30
git mode required for restore

allenfear:
## Effect
Explains that Git mode is required to initiate an environment restore. The restore button is greyed out when the environment is in SFTP mode.


alexfornuto:
@rachelwhitton I'd like your input; specifically, if that's the link I added is the most appropriate place to send readers to learn about the Git connection mode.

PR #3637, opened 2018-04-30
Add Out of Scope header

### LINKED ISSUE #3409: Alert users when docs beyond scope of support
Re: [Style Guide](https://pantheon.io/docs/style-guide/)

Priority (Low‚ Medium‚ High): medium / high

## Issue Description:
Some guides go beyond our scope of support, making it difficult for CSE to manage expectations in chat and tickets.
## Suggested Resolution 
Put some type of notice / alert on guides that go beyond the scope of support (like the build tools and frontend performance guides) to help manage users expectations when they're chatting with support. 

Consider directing towards agency enablement page / power users

alexfornuto:
Closes #3409 

## Effect
- Creates a new Header note warning users when the contents of a guide or doc are outside our scope of support
- Includes a link to report issues with the doc to GitHub (duplicate I know, but worth it I think)
- Uses Pantheon-brand colors

**Example:**
image

## Remaining Work
- [ ] Discuss with appropriate parties to lock down language.
- [ ] Discuss / Review color choices (@kimby77 I'm :eyes: at you).
- [ ] Discuss moving the Note below the title & description
- [ ] Identify docs & guides this note should be applied to (looking to @pantheon-systems/cse for input on that)

## Docs Team Work
- [ ] Technical Review
- [ ] Copy Review

## Post Launch
- [ ] ~~Redirect `/docs/old-path/` => `/docs/new-path/` (if applicable)~~
- [ ] ~~Include/exclude pages ^ respectively within docs search service provider (if applicable)~~
- [ ] Update Status Report

### DIFF
--- a/source/_partials/content/oos.html
+++ b/source/_partials/content/oos.html
@@ -1,7 +1,6 @@
 <div class="row" style="margin-bottom:15px;">
   <div class="col-md-12">
     <div id="earlyaccessoutter" class="alert oos" role="alert" style="border-radius: 4px;">
-     <h3 class="oos">Out of Scope</h3>
      <p class="oos">This page covers topics that go above and beyond our <a href="/docs/getting-support/#scope-of-support">Scope of Support</a>. While we provide these instructions to help you better develop your site, it's your responsibility to maitain your site's code.</p>
      <p class ="oos">If you find this doc to be incorrect or out of date, please <a href="https://github.com/pantheon-systems/documentation/issues/new?title={{page.title|url_encode }}%20Doc%20Update%20&body=Re%3A%20%5B{{page.title|url_encode }}%5D(https%3A%2F%2Fpantheon.io{{page.url}})%0A%0APriority%20(Low‚%20Medium‚%20High)%3A%0A%0A%23%23%20Issue%20Description%3A%0A%0A%23%23%20Suggested%20Resolution" target="blank" class="external">report the issue</a> on GitHub</p>
    </div>


slashzero:
All of the debugging docs are technically out of scope - as we may want to think about those tto:

https://pantheon.io/docs/debug-slow-performance/

Plus maybe Quicksilver, we can't really support that:

https://pantheon.io/docs/quicksilver/

Perhaps

https://pantheon.io/docs/guides/phpstorm-composer/
https://pantheon.io/docs/guides/wordpress-automated-testing/

kimby77:
Are the two bold phrases in the Out of Scope container links? They need to look more like they're clickable.  

I think it's good the out of scope note is at the top so that it doesn't break the flow of the actual doc.  I think it could be less out of the way though so that it isn't so dominant on the page and acts as more of an aside.  The two paragraphs could probably be combined so that would make it shorter, and you could also try without the "Out of Scope" header.  Having it take more of a backseat will also help communicate that while this is out of our scope of support if the user runs into trouble, it is still possible to do this on Pantheon.  The large header may be making the message sound more dire than it actually is.

alexfornuto:
Thanks @kimby77! The only part of your feedback I haven't yet applied is combining the two paragraphs. I want to make a special point of letting readers know that they can and should turn to GitHub with any problems they have with the doc, and I feel it's worth the extra space.

@rachelwhitton Feel free to take a pass at the copy.

rachelwhitton:
Still blocked. [See comment on related issue](https://github.com/pantheon-systems/documentation/issues/3409#issuecomment-391060710) for what we need to move forward.

PR #3638, opened 2018-05-01
Update alternative-php-cache.md

twooten:
This doc states that clearing the cache via the Dashboard will clear Redis. This is not true according to David S.

Please see David's explanation in BUGS-1779


alexfornuto:
Thanks @twooten!

PR #3660, opened 2018-05-07
Added a bit about ENV value being case-sensitive

twooten:


rachelwhitton:
+1! Thanks for the PR @twooten, much appreciated

PR #3682, opened 2018-05-16
Explicitly defining code style for the code blocks

davidneedham:
Closes #

## Effect
- The formatting / coloring of code blocks used for the various `terminus` commands was inconsistent. This manually defines that the format should be bash for every terminus code block.
- An experiment to test breaking long lines of code with `\`

## Remaining Work
- [ ] If needed

## Post Launch
- [ ] Update Status Report


davidneedham:
Adding explicit call to the `bash` syntax highlighting on all terminus commands. 

Decided not to add line breaks with `\` since we don't use that standard elsewhere in docs ATM.

rachelwhitton:
Thank you so much @davidneedham! This looks great +1

PR #3683, opened 2018-05-16
Fix Drush "aa" command example

sarahg:
## Effect
- Removes unneeded alias declaration from the Drush "aa" command example. Running it as shown in the current example yields "Could not find the alias" errors.

## Remaining Work
- [ ] technical review


rachelwhitton:
thanks for the PR @sarahg! much appreciated 

ping @greg-1-anderson - thoughts?

PR #3689, opened 2018-05-18
Temporary File Management Refactor

### LINKED ISSUE #3121: Temporary File Management with Multiple Application Containers Doc Update
Re: [Temporary File Management with Multiple Application Containers](https://pantheon.io/docs/temp-files/)

Priority (Low‚ Medium‚ High):

## Issue Description:

It might just be me - but I am finding it a little confusing to understand which info applies to which use case / framework / type of tmp directory.
Ideas for making the doc easier to translate into "what do I need to do":

In general:
- What is the difference between tmp and system tmp directory ?
- Which sections are about system tmp, which are about tmp?
- Which tmp directory paths are sync'ed between containers, which are not?

Drupal 8:
> Pantheon's upstream for WordPress and Drupal 8 configures the temporary directory path dynamically 
- Is any of the info in this doc relevant for Drupal 8 - or only for Drupal 7?

Drupal 7:
> Drupal 7 sites do not have the option to set or override the system /tmp path.
- Is any of the info in this doc relevant for Drupal 7 - or only for Drupal 8?

Drupal 8 / WordPress:
>Pantheon's upstream for WordPress and Drupal 8 configures the temporary directory path dynamically 
- Which tmp directory path does Pantheon automatically assign? /tmp? ~/tmp?

Sync'ed tmp files:
- Which directory do I need to use in order to get tmp files sync'ed between containers? ~/tmp? ~/files/private/tmp? Is this system tmp dir or tmp dir?
(This is a frequent use case for this doc. Example: Importing data / batch jobs).
Which tmp folders do not get sync'ed between environments?

Red warning box:
- The warning is under the "Using Valhalla, Pantheon's Networked Filesystem" section. The section under "Drupal" uses the same directory  which is used as above the warning box too ('private://tmp').

Private Temporary Files section:
Could be relevant to link this section and the below doc to each other:
https://pantheon.io/docs/private-paths/

Suggestion to add info:
Might be good to add info about the slight delay in sync'ing static files between application containers (I can't remember how much this is).
The solution is to add a delay between each batched request. 

## Reference

Motivated by this: 95343

alexfornuto:
Closes #3121 

## Effect
- Refactor doc for better readability
- Consider users will most likely find this guide after encountering one of two scenarios: 
  - Fix an unsupported tmp path 
  - Support persistent tmp path across multiple app containers

## Post Launch
- [ ] Redirect `/docs/temp-files/` => `/docs/tmp/`
- [ ] Update Status Report


alexfornuto:
Input from @pantheon-systems/cse requested over Slack.

xq42:
The doc looks like an improvement to me over the old version.

One question though - the original doc has this line in it (3rd paragraph):

> Drupal 7 sites do not have the option to set or override the system tmp path.

This version doesn't make a distinction between D7 and D8.

Upon a second reading, it appears this version only deals with WP and D8. If that is correct, do we have somewhere else we'll be sending folks who have multiple appservers on D7?

alexfornuto:
@xq42 very astute observation. The old reference to D7 referred to overriding the main temp variable for the entire CMS. Since this version only covers overriding module/plugin specific variables, it no longer applies.

When I'm back "in" I'll replace references to D8 with Drupal, since the current content should cover both.

rachelwhitton:
This update should also cover the following, [taken from the description of a related issue in drops-8 repo](https://github.com/pantheon-systems/drops-8/issues/114)
> admin/config/media/file-system will still display the directory set in the configuration system, and this displayed directory will never be used. This may lead to some confusion.

This has been mentioned in several tickets/chats and it does lead to confusion

rachelwhitton:
@greg-1-anderson if/when convenient can you help get us pointed in the right direction? 😬 Sorry for the mega comments but this one has us quite confused! 

## Context
We're rewriting our tmp files doc which is commonly shared by support to debug file not found error that references a tmp file while doing some kind of upload/import operation on production env that uses multiple application containers. 

The resolution provided is to override the standard tmp path (`$_SERVER[HOME] .'/tmp'`) with a filesystem path (`sites/default/files/tmp` or `sites/default/files/private/tmp`)

I'm speaking in terms of D8 here in this comment but the page is intended to provide a consistent recommendation across all three frameworks.

## Questions
1) Should we document a system override (`$config['system.file']['path']['temporary']`) or a specific settings override (`$config['plupload.settings']['temporary_uri']`)? 

   - If specific settings override, what about media module? Isn't the "specific settings" override in that case the same as the system? 

   - If systems override, can the configuration be visually confirmed anywhere (since the admin interface will report standard path)?

2) What are the implications of using a private (`sites/default/files/private/tmp`) vs public path (`sites/default/files/tmp`)?  How would a customer know which one fits their site's use case?

3) When configuring tmp path to the filesystem, will tmp files pile up overtime? Or are they cleaned up?

Any clarifications you can provide would be appreciated, thanks in advance!

rachelwhitton:
Summarizing my notes after meeting with Greg: 
- D8 and WP set appropriate tmp path by default, D7 can do so similarly via `$conf['file_temporary_path'] =  $_SERVER['HOME'] . '/tmp'; in `settings.php
- it's better to use an alternative module than to take performance hit from filesystem path, when it's unavoidable use the private files path 
   - public files path not recommended because it is web accessible; standard tmp files are treated as if they have sensitive information that should only be read by php and not by the browser
- option specific should be enough; avoid overriding the default system wide
- D8 should update the module settings and export config as described in previous comments using Terminus


Work still in progress on this PR - not yet ready for review

rachelwhitton:
No additional work remaining; ready for final review.

source/_docs/tmp.md
```diff
@@ -0,0 +1,166 @@
+---
+title: Temporary File Management
+description: Understand Pantheon's default temporary path and learn how to debug .tmp file errors.
+tags: [debugcode, infrastructure]
+categories: []
+---
+## Default Temporary Path  
+Pantheon configures an appropriate temporary path for [WordPress](https://github.com/pantheon-systems/WordPress/blob/master/wp-config.php#L83-L86){.external} and [Drupal 8](https://github.com/pantheon-systems/drops-8/blob/master/sites/default/settings.pantheon.php#L146-L154){.external}. Drupal 7 sites can achieve the same configuration by adding the following to `settings.php`:
```
greg-1-anderson: The contents of GitHub links on the `master` branch may change as more commits are added. To avoid this, link to a specific tag:

- [WordPress](https://github.com/pantheon-systems/WordPress/blob/4.9.6/wp-config.php#L84:L86)
- [Drupal 8](https://github.com/pantheon-systems/drops-8/blob/8.5.3/sites/default/settings.pantheon.php#L152:L154)

PR #3697, opened 2018-05-22
Add troubleshooting suggestions for infinite redirect loops

### LINKED ISSUE #3646: "Redirect to HTTPS and the primary domain" example can cause a redirect loop on multilingual Drupal sites
Re: [Switching Sites from HTTP to HTTPS](https://pantheon.io/docs/http-to-https/)

## Issue Description:

The example HTTPS/primary domain redirect code can conflict with Drupal multilingual settings when detecting languages via domain name, resulting in an endless loop.

Internal Desk ref: #103730

## Suggested Resolution

Add a section to https://pantheon.io/docs/domains/#troubleshooting or perhaps "more redirect scenarios" (if it requires different PHP code that the main example) to illustrate working settings for multilingual sites.

sarahg:
Closes #3646

## Effect
- Adds a section to Troubleshooting > Infinite Loops that notes that this problem can come from modules/plugins that allow you to add redirects in the CMS.

## Remaining Work
- [ ] look at Desk tickets to see if there are other plugins/modules we should list here
- [ ] copy review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/domains.md
+++ b/source/_docs/domains.md
@@ -101,9 +101,13 @@ All redirect logic should include the `php_sapi_name() != "cli"` conditional sta
 [error]
 ```
 ### Infinite Redirect Loops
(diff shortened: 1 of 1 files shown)


rachelwhitton:
Nice! Thanks @sarahg quick spot check looks good but would like to link out to plugins/modules mentioned to provide context. I can do that tomorrow

rachelwhitton:
Decided against linking out to the modules/plugins mentioned - felt distracting. Like how you listed it out so I left it as is. 

Thanks again for the PR! +1

Issue #3699, opened 2018-05-22
Clone existing Pantheon Site with Terminus Plugin


ari-gold:
Re: [Migrate Sites to Pantheon](https://pantheon.io/docs/migrate/)

Priority (Low‚ Medium‚ High): High

## Issue Description:

"How do I clone an existing Pantheon site?" section  can be improved

## Suggested Resolution

- Include instructions on how to use new @ataylorme  terminus plugin to clone a site
- Recommend when you'd use this method vs WP Plugin
- Remove instructions to use terminus drush, if appropriate

rachelwhitton:
Closed by #3701

PR #3701, opened 2018-05-22
List the Terminus site clone plugin under 'How do I clone an existing Pantheon site?'

ataylorme:
This is my attempt at the update suggested by @ari-gold in #3699.

## Effect
- Edits the "How do I clone an existing Pantheon site?" FAQ in the `/migrate` doc
- Removes manual site archive/import instructions for Drupal using `drush ard` in favor of the Terminus site clone plugin

### DIFF
--- a/source/_docs/migrate.md
+++ b/source/_docs/migrate.md
@@ -251,13 +251,19 @@ Go the to files directory of your existing site and check if the site archive wa
 
 ## Frequently Asked Questions (FAQs)
 ### How do I clone an existing Pantheon site?
-You can make a copy of a WordPress site on Pantheon by following the [standard migration procedure](#migrate-existing-sites) described above. The procedure does not deviate for WordPress sites already hosted on Pantheon and is preferred as it built into the dashboard.
+You can make a copy of a WordPress site on Pantheon by following the [standard migration procedure](#migrate-existing-sites) described above. The procedure does not deviate for WordPress sites already hosted on Pantheon and is preferred since it's built into the Site Dashboard.
 
-Drupal 7, Drupal 8 and WordPress sites may also use [the Terminus site clone plugin](https://github.com/pantheon-systems/terminus-site-clone-plugin) to clone code, database and/or media files from one Pantheon site to another. This method requires use of [Terminus](https://pantheon.io/docs/terminus/) and the command line tool.
+Drupal 7, Drupal 8 and WordPress sites can use Terminus to clone one Pantheon site to another from the command line. This method requires you to [install and authenticate Terminus](/docs/terminus/install), then install the [Terminus Site Clone](https://github.com/pantheon-systems/terminus-site-clone-plugin){.external} plugin. 
+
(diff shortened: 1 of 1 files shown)


ari-gold:
👍  💯 :shipit: 

@rachelwhitton @alexfornuto Can you please review?

ataylorme:
Thanks for suggesting it @ari-gold. In my announcement blog post, currently a draft, I highlight the use case for a "starter" site where agencies can keep a skeleton build, like [the Underscores theme for WordPress](https://underscores.me/), with some demo content. They can then clone it for new projects and start hacking.

This is a separate use case than custom upstreams as it's not a distro or shared code but something meant to diverge from the original after clone.

You could also use it in conjunction with a custom upstream to just clone demo content, files and settings from a demo site on the distribution to a new site spun up from the upstream as well.

It has me wondering if that is worth mentioning in the custom upstream doc or creating a separate doc/guide for the site clone plugin.

rachelwhitton:
Exciting! Thanks for the PR @ataylorme! Thanks for reporting the issue and your review @ari-gold! 
+1! 
Note that this introduces a new layer of complexity by requiring a plugin. The previous method relied on Terminus so it's not a big hit to accessibility but this is "more advanced" since it requires extending Terminus. For ease of use it would be best to support this operation as a dashboard feature.  

> You could also use it in conjunction with a custom upstream to just clone demo content, files and settings from a demo site on the distribution to a new site spun up from the upstream as well.

> It has me wondering if that is worth mentioning in the custom upstream doc or creating a separate doc/guide for the site clone plugin.

@ataylorme I would avoid noting the plugin on the custom upstream doc - just because custom upstreams is a concept users struggle with at first, so new content needs to be carefully considered to make sure we don't distract/overwhelm. A separate guide sounds interesting! Would prefer mentioning a related guide over the plugin without supporting context

ataylorme:
> I would avoid noting the plugin on the custom upstream doc - just because custom upstreams is a concept users struggle with at first, so new content needs to be carefully considered to make sure we don't distract/overwhelm

Copy that.

> The previous method relied on Terminus so it's not a big hit to accessibility but this is "more advanced" since it requires extending Terminus.

The previous steps required getting a manual dump from `drush ard` and importing it along with "Adjust paths as needed for sites with a nested docroot". While I agree requiring a Terminus plugin adds complexity I don't think it is more complex than what was previously required.

rachelwhitton:
Only consider it more complex because previous approach relies on a drush command that's available out of the box w/Terminus and the new approach requires extending. 

Agree the command you use in the end is easier this way, but getting there may be more challenging for users new to the command line. 

Especially since we don't provide instructions to install the plugin on the page and they have to deduce that themselves from /docs/terminus/plugins/ or from the install instructions in the plugin's readme on Github -- which are different.

Not a blocker though, just an observation of small risk

rachelwhitton:
It has been removed as an option in context of duplicating a Pantheon site -- it's still the path for guided Drupal migrations via the Site Dashboard. The instructive copy for that operation is served by the product and not repeated directly in the migrate documentation (see Drupal tab of guided migration doc).

PR #3774, opened 2018-06-11
Replace June 28th with August 1st

rachelwhitton:
## Effect
- Replace June 28th with August 1st


rachelwhitton:
+1 received from lizzie via slack

PR #3790, opened 2018-06-15
Fix Perf Small resource specs

jfussion:
Base on this bug ticket, Performance Small gets same configuration with Professional plan so 256MB is the correct PHP Memory Limit 

Ref: BUGS-2037

Closes #

## Effect
- Change PHP memory Limit of Performance Small


alexfornuto:
Holding pending confirmation that this is a docs issue not a platform issue.

alexfornuto:
Fixed in #3848, thanks!

PR #3858, opened 2018-07-10
Update d7-composer-init.html

jfussion:
## Effect
- Modify the suggested composer.json and to include 'sites/all/vendor' as `vendor-dir`.

By default, composer downloads packages to `vendor` directory but `Composer Vendor` module only autoloads the content of `sites/all/vendor`.  Specify vendor-dir to `sites/all/vendor` will fix the issue.

Test site: http://sendgrid-jts.pantheonsite.io/

### DIFF
--- a/source/_partials/content/d7-composer-init.html
+++ b/source/_partials/content/d7-composer-init.html
@@ -29,6 +29,9 @@
     "post-update-cmd": [
       "@remove-git-submodules"
     ]
+  },
+  "config": {
+    "vendor-dir": "sites/all/vendor"
   }
 }</code></pre>
 </li>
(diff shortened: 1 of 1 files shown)


alexfornuto:
@greg-1-anderson I can do A), you can do B). Which is preferred?

greg-1-anderson:
@ari ^^?

greg-1-anderson:
The current guide is missing a step, and won't work without it. If the user figures out how to correctly relocate the vendor directory, then they should also realize (maybe) that they should protect the location. However, if we instruct the user to relocate vendor, then we should definitely tell them to protect the destination location.

So, while I do not think we need to do an emergency removal of this document (an unprotected vendor is undesirable, but not always a vulnerability), I do think that you have made a good argument for doing a) now, and optionally b) later, since that will take longer.

greg-1-anderson:
I do believe that `code/vendor` is protected, actually.

alexfornuto:
My latest commit creates this warning:

warning

@greg-1-anderson does this satisfy your concerns?

CC @jfussion

alexfornuto:
@greg-1-anderson I'm hesitant to show `pantheon.yml` examples within this partial because:

 - The partial itself and the docs it's used in are already so dense, I don't want to inhibit readability. 
 - If changes are made to the way protected web paths are created, I want to know that it's only explained in one place (i.e. reduce stale content in the future).

PR #3865, opened 2018-07-12
update rsync in deployment scripts

alexfornuto:
## Effect
- Hopefully by removing --size-only rsync will properly use --checksum and only copy updated files to the multidev

## Post Launch
- [ ] Update Status Report


alexfornuto:
These updates were implemented in #3875 and #3886.

PR #3885, opened 2018-07-18
Implement Dockerize tool

alexfornuto:
## Effect
- Implements `dockerize` a tool build in to the CircleCI docker image ours is based from.
- Used to wait until services come online.
- More reliable than using `sleep`.


alexfornuto:
It works!

image

PR #3886, opened 2018-07-18
Rsync Improvements

alexfornuto:
## Effect
- Re-instituted checksum comparison for rsync transfers. Testing showed that not only did skipping checksum _not_ keep us from hitting the BUGS-1877 issue, but it also resulted in updated files being skipped sometimes.

- Adjusted the `rsync` flags controlling output, piped output to a logfile, and grep the logfile for entries where files are modified. This reduces the output to only what's relevant to debugging.

- print more script output to CI shell.

- Several HTML => Markdown adjustments, as part of testing that rsync saw file changes.


joemiller:
👍

PR #3888, opened 2018-07-18
Change double quotes to single

rick02840:
## Effect
- in this:
 terminus site:create terminus-cli-create 'Terminus CLI Create' 21e1fada-199c-492b-97bd-0b36b53a9da0

## Post Launch
- [ ] Update Status Report


alexfornuto:
@rick02840 is it not the case that either quote symbol will work? Additionally, the `"` will support variables, which some may use when scripting these commands.

alexfornuto:
Closing for reasons notes above, but thanks for contributing!

PR #3889, opened 2018-07-19
update dev.static-docs after deployment

alexfornuto:
## Effect
- Updated the `deploy-live.sh` script to update the dev.static-docs environment, from which multidev environments are cloned.

## Remaining Work
- [ ] Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/scripts/deploy-live.sh
+++ b/scripts/deploy-live.sh
@@ -7,7 +7,7 @@
 #=====================================================#
 
 
-# Migrate paginated files to avoid .html within the URLs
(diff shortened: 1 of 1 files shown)


alexfornuto:
Actually, this is pointless. When we build static files for the multidev environment we first change the templates to adjust Links for the new environment. This will change the checksum of every page, so having an updated dev environment won't reduce the amount of files `rsync` transfers when we first create a multidev environment. 

Work on this branch can continue if/when we change the CTA links in the template files to be dynamically generated for the proper environment.

PR #3901, opened 2018-07-25
Fix CircleCI caching.

felicianotech:
This PR fixes caching for the CircleCI build.

The current config is using CircleCI 1.0 caching syntax which no longer works on 2.0.

The Node and Composer caches each get cached in their own step. A checksum of each of their lock files is used as the cache key. This means, as long as the lock file doesn't change, all the dependencies are the same and we can restore the cache.

The third directory seems to be an erroneous one and was removed.

Also added a comment at the end for some FYI info that might be useful sometime in the future.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,4 @@
 version: 2
-dependencies:
-  cache_directories:
-    - ~/.composer/cache
-    - ~/build/vendor
-    - ~/build/node_modules
 jobs:
   build:
     docker:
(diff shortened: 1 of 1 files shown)


alexfornuto:
I made a copy of this branch on the local fork to kick off testing with our environment variables:

 - The first build completed successfully, and created the cache: https://circleci.com/gh/pantheon-systems/documentation/11548
 - The second build found and used the cache: https://circleci.com/gh/pantheon-systems/documentation/11549

PR #3908, opened 2018-07-27
Add Force Login to the list

jfussion:
## Effect
- Added Force Login WordPress plugin to the [modules-plugins-known-issues](https://pantheon.io/docs/modules-plugins-known-issues) page

### DIFF
--- a/source/_docs/modules-plugins-known-issues.md
+++ b/source/_docs/modules-plugins-known-issues.md
@@ -339,6 +339,11 @@ For more details, see [SERVER_NAME and SERVER_PORT on Pantheon](/docs/server_nam
 **Solution**: See [Caching: Advanced Topics](/docs/caching-advanced-topics/) for details on how to bypass the platform page cache.
 <hr>
 
+### [Force Login](https://wordpress.org/plugins/wp-force-login/){.external}
+**Issue**: This plugin appends a port number using `$_SERVER['SERVER_PORT']` at the end of the URL when the user logs in to the site.
+
+**Solution**: See [Set SERVER_PORT Corectly](https://pantheon.io/docs/server_name-and-server_port/#set-server_port-correctly)
+
 ### [Instashow](https://elfsight.com/instagram-feed-instashow/){.external}
 **Issue**: The Instashow plugin relies on query parameters that are not compatible with Pantheon's Edge Cache. See [PANTHEON_STRIPPED](https://pantheon.io/docs/pantheon_stripped/){.external} for more information. This inhibits the ability to set the authorization token required to make the plugin function.
 <hr>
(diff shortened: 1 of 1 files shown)


source/_docs/modules-plugins-known-issues.md
```diff
@@ -339,6 +339,12 @@ For more details, see [SERVER_NAME and SERVER_PORT on Pantheon](/docs/server_nam
 **Solution**: See [Caching: Advanced Topics](/docs/caching-advanced-topics/) for details on how to bypass the platform page cache.
 <hr>
 
+### [Force Login](https://wordpress.org/plugins/wp-force-login/){.external}
+**Issue**: This plugin appends a port number using `$_SERVER['SERVER_PORT']` at the end of the URL when the user logs in to the site.
```
alexfornuto: Is it always the same port number? What number is it? What error does this cause, if any?

source/_docs/modules-plugins-known-issues.md
```diff
@@ -339,6 +339,12 @@ For more details, see [SERVER_NAME and SERVER_PORT on Pantheon](/docs/server_nam
...
+
+**Solution**: See [Set SERVER_PORT Corectly](https://pantheon.io/docs/server_name-and-server_port/#set-server_port-correctly).
```
alexfornuto: Would the customer need to replace any values in the example snippet linked? Which values, and replaced to what?

source/_docs/modules-plugins-known-issues.md
jfussion: It depends on what the port number the appserver is using. You can check that port number on the debug pannel of the site
children s hospital of philadelphia pantheon dashboard 2018-07-28 09-48-37

source/_docs/modules-plugins-known-issues.md
jfussion: No need to modify the snippet.

jfussion:
Here's the line from Force login plugin code where it uses the said environment variable:
`wp-force-login/wp-force-login.php:31:      $url .= in_array( $_SERVER['SERVER_PORT'], array('80', '443') ) ? '' : ':' . $_SERVER['SERVER_PORT'];`

source/_docs/modules-plugins-known-issues.md
jfussion: @alexfornuto It's not the same port. The port number is random based on the port number used by the appserver as shown in the image. The result is a white page.

You can test it here: http://force-login-jts-wp.pantheonsite.io/hello-world
user: josephpantheon
pass: WbjgZ(Bt#hVKq3zDUT)uekR2

PR #3909, opened 2018-07-27
Update create-custom-upstream.md

tinefin:
## Effect
Adds info that changing repository password also requires to make a new upstream, just like changing the repo url.

https://pantheon-systems.desk.com/agent/case/108506


alexfornuto:
+1 Thanks @tinefin!

PR #3937, opened 2018-08-07
Define issue with wp_filesystem->get_contents

### LINKED ISSUE #3674: Add WP_Filesystem→get_contents() workaround
Re: Add WP_Filesystem→get_contents() workaround

## Issue Description:

Calls to WP_Filesystem→get_contents() fail if the site is in Git mode even if the file being called is contained in wp-content/uploads.

The ticket has replicated the issue and confirmed the workaround - initially reported by Astra theme author.

Surfaced in ticket 104691

## Suggested Resolution

Workaround: add `define('FS_METHOD', 'direct');` to wp-config.php.

alexfornuto:
Closes #3674 

## Effect
- Defines the issue with `wp_filesystem->get_contents` in git mode
- Explains why it shouldn't be used
- Provides a workaround 

## Remaining Work
- [ ] Technical Review from someone who knows WP better than I. 
- [ ] Copy Review (maybe)

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/modules-plugins-known-issues.md
+++ b/source/_docs/modules-plugins-known-issues.md
@@ -526,6 +526,18 @@ The list of [WordPress roles and capabilities](https://codex.wordpress.org/Roles
 * `edit_plugins`
 * `edit_themes`
 
+### [wp_filesystem->get_contents()](https://developer.wordpress.org/reference/classes/wp_filesystem_base/get_contents/){.external}
+
+**Issue**: The function `wp_filesystem->get_contents()` can fail when an environment is in Git mode (as Test and Live always are) because it is aware of filesystem-level permissions which are restricted in this mode.
+
+**Solution**: As described in [this StackExchange answer](https://wordpress.stackexchange.com/questions/166161/why-cant-the-wp-filesystem-api-read-googlefonts-json/166172#166172){.external}, for cases where file ownership doesn't matter this function could be replaced with `file_get_contents()`. This is true of most cases where the file in question is only being read, not written to.
+
+If the function cannot be replaced, you can create a workaround by adding to your `wp-config.php` file:
+
+```php
+define('FS_METHOD', 'direct');
+```
+
 ## PHP Libraries
 Due to the cloud-based infrastructure of the Pantheon platform, certain PHP libraries are not available on the platform.
 


slashzero:
LGTM +1

source/_docs/modules-plugins-known-issues.md
```diff
@@ -526,6 +526,18 @@ The list of [WordPress roles and capabilities](https://codex.wordpress.org/Roles
 * `edit_plugins`
 * `edit_themes`
 
+### [wp_filesystem->get_contents()](https://developer.wordpress.org/reference/classes/wp_filesystem_base/get_contents/){.external}
+
+**Issue**: The function `wp_filesystem->get_contents()` can fail when an environment is in Git mode (as Test and Live always are) because it is aware of filesystem-level permissions which are restricted in this mode.
+
+**Solution**: As described in [this StackExchange answer](https://wordpress.stackexchange.com/questions/166161/why-cant-the-wp-filesystem-api-read-googlefonts-json/166172#166172){.external}, for cases where file ownership doesn't matter this function could be replaced with `file_get_contents()`. This is true of most cases where the file in question is only being read, not written to.
+
+If the function cannot be replaced, you can create a workaround by adding to your `wp-config.php` file:
```
danielbachhuber: What are the other implications of this workaround? I'm not sure it's something I'd recommend, as it has broader impact on WordPress' behavior.

Rather than including this workaround, you could simply leave this section as the "Issue" and "Solution.

danielbachhuber: Another way of phrasing: including this workaround becomes the "official Pantheon supported solution". I don't think it's quite that, so better to leave unmentioned.

alexfornuto:
@danielbachhuber I've been digging in  deeper on this, and it looks like wordpress.org has taken a stance against `file_get_contents` in themes they host:
https://github.com/WordPress/theme-check/issues/55
http://ottopress.com/2011/tutorial-using-the-wp_filesystem/

In light of this, should we reconsider the "workaround" of defining `FS_METHOD`?

danielbachhuber:
> In light of this, should we reconsider the "workaround" of defining `FS_METHOD`?

I think my original assessment is still relevant:

> I'm not sure it's something I'd recommend, as it has broader impact on WordPress' behavior.
>
> [...]
>
> Another way of phrasing: including this workaround becomes the "official Pantheon supported solution". I don't think it's quite that, so better to leave unmentioned.

The theme check guidelines seem less relevant to Pantheon in this context.

alexfornuto:
@danielbachhuber it's not just guidelines, according to the theme author in the original ticket:

>Unfortunately, `file_get_contets()` is not allowed for themes hosted o W.org due to security concerns. It gets flagged with the automatic checks when uploading the theme o W.org.

danielbachhuber:
>  It gets flagged with the automatic checks when uploading the theme o W.org.

This only means https://wordpress.org/themes/, which isn't relevant to a Pantheon customer installing the theme on their site. A standard WordPress installation doesn't perform code scanning when uploading a theme through the WordPress dashboard.

alexfornuto:
@danielbachhuber isn't that the same as saying that themes allowed on wordpress.org/themes won't work on Pantheon, and those that work on Pantheon aren't allowed on wordpress.org?

danielbachhuber:
@alexfornuto Sure, although I'm not sure that's relevant. Can you describe in greater detail what you're trying to accomplish?

alexfornuto:
I'm just trying to self-enlighten, and also raise awareness to potential issues that should be addressed.

In this case the issue is: if Pantheon and the major repository of WordPress themes disagree on the functions to use in those themes, to the point of being mutually exclusive, it adversely affects those using the platform by drastically limiting the themes available to them.

PR #3955, opened 2018-08-11
Uploading files over 256MB is not possible

twooten:
The doc states "Pantheon cannot be used to host files over 256MB, no matter how the file is uploaded." which seems to imply that it _is_ possible to use SFTP or some other method to upload files over 256MB. 

This change is to clarify that the transfer will fail due to the large file size.


alexfornuto:
Thanks @twooten!

PR #3958, opened 2018-08-14
Add Redirect Using Wildcards section to redirects.md

uberhacker:
## Effect
- Add `Redirect Using Wildcards` section to redirects.md

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/redirects.md
+++ b/source/_docs/redirects.md
@@ -134,6 +134,42 @@ if (isset($_ENV['PANTHEON_ENVIRONMENT']) && ($_ENV['PANTHEON_ENVIRONMENT'] === '
 }
 ```
 
+### Redirect Using Wildcards
+The following configuration will redirect requests for `example.com` using wildcard regex patterns:
+
+```php
+// Add wildcard redirects (old_path(s) => new_path)
+$redirects = array(
+  "/any/old/path/(.*)" => "/new",
+  "/starts-with(.*)" => "/new/path",
+  "/(.*)ends-with-this" => "/new/path",
+  "/(.*)contains-this(.*)" => "/new/path",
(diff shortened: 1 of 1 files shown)


source/_docs/redirects.md
```diff
@@ -134,6 +134,42 @@ if (isset($_ENV['PANTHEON_ENVIRONMENT']) && ($_ENV['PANTHEON_ENVIRONMENT'] === '
...
+  "/starts-with(.*)" => "/new/path",
+  "/(.*)ends-with-this" => "/new/path",
+  "/(.*)contains-this(.*)" => "/new/path",
+  "/node/(\d)+" => "/all-nodes-go-here",
+  "/user/(a-z)+/blog" => "/new-user-blogs",
+);
+
+$match = FALSE;
+foreach ($redirects as $old_path => $new_path) {
+  if ($match = preg_match("#^($old_path)$#", $_SERVER['REQUEST_URI'])) {
```
greg-1-anderson: This should be optimized to a single preg_match without a `foreach`, e.g.:
```
if (preg_match('#^(/old/path|/other/path)$', $_SERVER['REQUEST_URI']))) {
```

uberhacker: This will work if you wanted all the old paths to redirect to a single new path but what if you wanted each old path to redirect to a unique new path?

uberhacker: I imagine the current code could get bogged down if you had a lot of redirects but I can't imagine it would make much of a difference in performance with just a handful of redirects.

alexfornuto:
Closing as too edge-case / specific, per @greg-1-anderson.

uberhacker:
What does too edge-case mean?  Whatever...

PR #3963, opened 2018-08-15
Update modules-plugins-known-issues.md

tinefin:
Expire module's functionality is not to circumvent caching, but rather to help cached items automatically being expired/purged.

Adding a more relevant destination for a Pantheon solution.


alexfornuto:
Thanks @tinefin!

PR #3979, opened 2018-08-17
Use valid syntax for example variables.

sarahg:
## Effect
- Adjusts example cookie code so that the example variables will work if actually used. $path and $value need to be strings (http://php.net/manual/en/function.setcookie.php).

## Remaining Work
- none

## Post Launch
- [ ] Update Status Report


alexfornuto:
Thanks @sarahg!

PR #3998, opened 2018-08-23
Preferred pricing GA update

allenfear:
Adjusting timing to align with new anticipated sunset date.


allenfear:
Awaiting confirmation of the planned sunset window before merging.

Issue #3690, opened 2018-05-21
Getting the Client IP Address Doc Update 


eabquina:
Re: [Getting the Client IP Address](https://pantheon.io/docs/client-ip/)

Priority (Low):

## Issue Description:
The customer claims that this method will not be compatible because of Varnish cache.  

```
1. Use the system environment variable $_SERVER["REMOTE_ADDR"]. One benefit is that on Pantheon this takes into account if the X-Forwarded-For header is sent in cases when a request is filtered by a proxy.
2. Use Drupal’s ip_address() function. The function leverages the $_SERVER["REMOTE_ADDR"] variable to get the client IP, as well as parse other meta information like the remote proxy IP (if available), trim the forwarded IPs if they contain any additional commas or spaces, and filter out any untrusted IPs.
```

## Suggested Resolution
Provide comment FAQ that this will be ok with caching layer

alexfornuto:
@eabquina a few questions:

1)
> The customer claims that this method 

Which of the two methods listed is this referring to? Or is it both?

2) Have you tested this?

3) Are you saying that we need to make an addition to this doc saying that the doc does indeed work even if you (the reader) thinks it may not?

alexfornuto:
Marking this as closed by #4134, please re-open with the requested additional details if this is not a full resolution.

PR #4040, opened 2018-08-31
Update client-ip.md - Domain Access & Cloudflare

daggerhart:
Possibly related to #3690 

## Effect
- Details for getting the user's real IP address when dealing with multiple proxies

### DIFF
--- a/source/_docs/client-ip.md
+++ b/source/_docs/client-ip.md
@@ -27,6 +27,18 @@ We offer various methods for you to interact with your Pantheon site, so it is i
 
 When using Cloudflare as a stacked CDN or proxy, use the variable `$SERVER["HTTP_CF_CONNECTING_IP"]` instead of `$_SERVER["REMOTE_ADDR"]`.
 
+#### Drupal Domain Access module with Cloudflare
+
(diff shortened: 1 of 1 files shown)


davidneedham:
I have sites on Pantheon using Domain Access + CloudFlare that haven't had this problem, but I could be doing things differently. I don't have a way to test per se, but the suggestion appears to be right.

@daggerhart your use case was for D7, yes? Per this change record (https://www.drupal.org/node/1969794) `ip_address()` was removed in D8. I can help create a D8 version of this users can tab between, but I don't have as much experience in D8. The change record suggests `Drupal::request()->getClientIp()` - does that seem right from in this context?

daggerhart:
@davidneedham yes, it is for D7.

I'm not sure this is needed for D8. The problem with the `ip_address()` function is that it caches the user's ip address the first time it is run using `drupal_static()`, So if it gets the wrong IP on it's first attempt, every subsequent call to `ip_address()` will return that wrong IP. I don't see any similar cache in D8's `Drupal::request()->getClientIp()`, so it may not be necessary.

alexfornuto:
Thanks @daggerhart! I was hoping that by including all your commits in my branch this would show as "Merged" instead of "Closed" :frowning_face:.

Issue #4014, opened 2018-08-28
October Changelog Candidates


alexfornuto:
When entering product updates, please use the following format:

```
### Subject (i.e WP-CLI Upgrades)
Short description, like: WP-Cl has been upgraded to [1.5.0](https://make.wordpress.org/cli/2018/01/31/version-1-5-0-released/).
```

pantheon-upstream:
### Drupal 8.6.0 
Update to Drupal 8.6.0. For more information, see https://www.drupal.org/project/drupal/releases/8.6.0

greg-1-anderson:
### SPF records now supported with Pantheon's built-in MTA

Using a third-party email solution is still recommended; however, for those who choose to use Pantheon's built-in MTA, you may now set up an SPF record for your domain and include Pantheon's mail relays for improved delivery. For details, see: https://docs.pantheonsite.io/docs/email/

pantheon-upstream:
### Drupal 8.6.1 
Update to Drupal 8.6.1. For more information, see https://www.drupal.org/project/drupal/releases/8.6.1

ari-gold:
Fixed an issue where image derivatives were not generating successfully in Drupal maintenance mode.

greg-1-anderson:
Update to php-5.6.38, php-7.0.32, php-7.1.22 and php-7.2.10

alexfornuto:
Add new items to #4131

PR #4046, opened 2018-09-05
Document how to use Pantheon's SPF record

greg-1-anderson:
Closes #n/a

## Effect
- Describes how to use SPF records with Pantheon's built-in MTA

## Remaining Work
- [ ] Engineering review
- [ ] Documentation review

## Post Launch
- [ ] Redirect `/docs/old-path/` => `/docs/new-path/` (n/a)
- [ ] Update Status Report

### DIFF
--- a/source/_docs/email.md
+++ b/source/_docs/email.md
@@ -140,9 +140,14 @@ No, mail logs are not available for download and we do not recommend using the l
 ### What ports are recommended by Pantheon?
 Configuring mail to use port 25, 465 or 587 is strongly discouraged because they attract SPAM activities. Make sure that your service provider allows traffic on a port other than those mentioned and that you have correctly configured your site to use that port.
 
-### Can Pantheon provide, publish, or support SPF records?
+### Are there SPF records for Pantheon's local MTA (postfix)?
+If you are using Pantheon's local MTA ([not recommended](#outgoing-email)), and your domain contains an SPF record, then you should include Pantheon's SPF record, as shown below:
 
-As consumers of cloud infrastructure, we don’t have control over our IP ranges and they are subject to change without our notice. Publishing an SPF record would imply assurance on our end that it can work, which would be very difficult to guarantee given these circumstances. We take the decision of what we support and what we don’t very seriously, and at this time we’re not in a position to support SPF records.
+```
+v=spf1 include:spf.pantheon.io -all
+```
+
+Adjust the above example records as needed for your domain. Add the `include:spf.pantheon.io` to your existing record, keeping the rest unchanged. To craft a new SPF record for a domain that does not yet have one, use the [SPF Record Generator](https://mxtoolbox.com/SPFRecordGenerator.aspx?domain=example.com), and enter `spf.pantheon.io` in the **3rd party mail systems** text box.
 
 ### Why does my Gmail user name and password not work?
 


source/_docs/email.md
```diff
@@ -140,9 +140,14 @@ No, mail logs are not available for download and we do not recommend using the l
...
-### Can Pantheon provide, publish, or support SPF records?
+### Are there SPF records for Pantheon's local MTA (postfix)?
+If you are using Pantheon's local MTA ([not recommended](#outgoing-email)), and your domain contains an SPF record, then you should include Pantheon's SPF record, as shown below:
 
-As consumers of cloud infrastructure, we don’t have control over our IP ranges and they are subject to change without our notice. Publishing an SPF record would imply assurance on our end that it can work, which would be very difficult to guarantee given these circumstances. We take the decision of what we support and what we don’t very seriously, and at this time we’re not in a position to support SPF records.
+```
+v=spf1 include:spf.pantheon.io -all
+```
+
+Adjust the above example records as needed for your domain. Add the `include:spf.pantheon.io` to your existing record, keeping the rest unchanged. To craft a new SPF record for a domain that does not yet have one, use the [SPF Record Generator](https://mxtoolbox.com/SPFRecordGenerator.aspx?domain=example.com), and enter `spf.pantheon.io` in the **3rd party mail systems** text box.
```
ari-gold: Should we add a sentence like ....

"Make sure to configure your site to send email from the same domain that you've configured SPF records for."

greg-1-anderson:
Added an entry to the [Changelog](https://github.com/pantheon-systems/documentation/issues/4014#issuecomment-419517855) for this.

source/_docs/email.md
```diff
@@ -140,9 +140,14 @@ No, mail logs are not available for download and we do not recommend using the l
...
 
-As consumers of cloud infrastructure, we don’t have control over our IP ranges and they are subject to change without our notice. Publishing an SPF record would imply assurance on our end that it can work, which would be very difficult to guarantee given these circumstances. We take the decision of what we support and what we don’t very seriously, and at this time we’re not in a position to support SPF records.
+```
+v=spf1 include:spf.pantheon.io -all
+```
+
+Adjust the above example records as needed for your domain; be sure that you are configuring the SPF record for the domain that your site is set up to use in the `From` address of outgoing messages. If an SPF record exists for that domain, then add just the `include:spf.pantheon.io` part to whatever is already there, keeping the rest unchanged. To craft a new SPF record for a domain that does not yet have one, use the [SPF Record Generator](https://mxtoolbox.com/SPFRecordGenerator.aspx?domain=example.com), and enter `spf.pantheon.io` in the **3rd party mail systems** text box.
```
alexfornuto: >Adjust the above example records as needed for your domain;

@greg-1-anderson @ari-gold the only record displayed is the one we are suggesting they add for pantheon.io. How would that be adjusted based on the end-user's domain?

greg-1-anderson: @alexfornuto In addition to the text we show above, the user will need to add `mx` if they want to send email from the same servers that their mx records point at, or they might want to add the fqdn of their outgoing email servers if they are different than their receiving email servers, or they might want to add another `include` if they are sending through Google GSuite, etc.

Would it be clearer if we said "as needed for your use case" instead?  My feeling is that folks who really feel that they need spf records are already going to know what they need, and since this is not a recommended configuration, we shouldn't give too much of a tutorial to lead folks down a path that they don't need.

alexfornuto: It sounds like the clarification I'm looking for would come by replacing 'Adjust the above' with something like 'Include this record, along with any others needed for your configuration...'.

Thoughts?

greg-1-anderson: The key thing that we are trying to convey is that the user should add `include:spf.pantheon.io` to whatever else they need in their SPF record. I gave it as an example because this has to come after the `v=spf1` and before the `-all` (or `~all` or whatever form they wish to use).

Your suggestion sounds good.

alexfornuto:
Requesting final review from @greg-1-anderson and @ari-gold before deploying.

ari-gold:
LGTM. 

@garydylina  -- can you please review?

source/_docs/email.md
```diff
@@ -140,9 +140,18 @@ No, mail logs are not available for download and we do not recommend using the l
 ### What ports are recommended by Pantheon?
 Configuring mail to use port 25, 465 or 587 is strongly discouraged because they attract SPAM activities. Make sure that your service provider allows traffic on a port other than those mentioned and that you have correctly configured your site to use that port.
```
garydylina: This is.  just.  wrong!
> Configuring mail to use port 25, 465 or 587 is strongly discouraged because they attract SPAM activities. Make sure that your service provider allows traffic on a port other than those mentioned and that you have correctly configured your site to use that port.

I propose the following replacement:

> Pantheon strongly encourages using ports other than 25, 465 or 587 to send email because those ports are often blocked by service providers as an anti-spam measure. Here’s a list of popular email providers and the alternate ports which Pantheon recommends:

source/_docs/email.md
```diff
@@ -140,9 +140,18 @@ No, mail logs are not available for download and we do not recommend using the l
...
 
-### Can Pantheon provide, publish, or support SPF records?
+### Are there SPF records for Pantheon's local MTA (postfix)?
+If you are using Pantheon's local MTA ([not recommended](#outgoing-email)), and your domain contains an SPF record, then you should include Pantheon's SPF record, as shown below:
```
garydylina: > If you are using Pantheon's local MTA ([not recommended](#outgoing-email)), and your domain contains an SPF record, then you should include Pantheon's SPF record, as shown below:

I propose the following...

> If you are using Pantheon's local MTA and your domain contains an SPF record, then you should include Pantheon's SPF record, as shown below:

source/_docs/email.md
```diff
@@ -140,9 +140,18 @@ No, mail logs are not available for download and we do not recommend using the l
...
 
-As consumers of cloud infrastructure, we don’t have control over our IP ranges and they are subject to change without our notice. Publishing an SPF record would imply assurance on our end that it can work, which would be very difficult to guarantee given these circumstances. We take the decision of what we support and what we don’t very seriously, and at this time we’re not in a position to support SPF records.
+```
+v=spf1 include:spf.pantheon.io -all
```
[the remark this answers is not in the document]
greg-1-anderson: I left an example include out for fear of folks blindly copying the whole spf record. To that end, I think that `example-domain` is better than `_spf.google.com`, as the later might make folks think that Pantheon needs the google spf info.

I'm fine with including the example (it's probably better that way); there should also be an explicit call-out, "do not put `include:spf.example-domain.com` in your SPF record; replace with the mail servers your domain uses in its place", or something to that effect.

source/_docs/email.md
garydylina: @ari-gold Please review my proposed changes with an eye toward improving the doc.  I offer these proposed changes in that spirit, but I don't feel strongly that my proposed versions should be accepted.  Your call.

garydylina: Are we going to document the name of the A and AAAA records somewhere so interested parties can use them?   dig mail-relay.pantheon.io any

greg-1-anderson: We could, but what would they use them for? They should send mail via the php `mail()` function, so I don't think there is a common use for these. If there is a use case that necessitates using this, then we should document it in the context of that use case.

source/_docs/email.md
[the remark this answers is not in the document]
alexfornuto: I'm OK with including `example.com`, which is explicitly reserved for documentation purposes, especially in light of the list of adjustments required right below.

alexfornuto:
I made adjustments based on @garydylina's input. Unless there's anything else to discuss, I can push this tomorrow.

source/_docs/email.md
```diff
@@ -138,11 +138,20 @@ We strongly recommend that you do not use the local MTA (postfix) as described [
...
+Pantheon strongly encourages using ports other than `25`, `465` or `587` to send email because those ports are often blocked by service providers as an anti-spam measure.  Make sure that your service provider allows traffic on a port other than those mentioned and that you have correctly configured your site to use that port.
 
-### Can Pantheon provide, publish, or support SPF records?
+### Are there SPF records for Pantheon's local MTA (postfix)?
+If you are using Pantheon's local MTA ([not recommended](#outgoing-email)), and your domain contains an SPF record, then you should include Pantheon's SPF record, as shown below:
 
-As consumers of cloud infrastructure, we don’t have control over our IP ranges and they are subject to change without our notice. Publishing an SPF record would imply assurance on our end that it can work, which would be very difficult to guarantee given these circumstances. We take the decision of what we support and what we don’t very seriously, and at this time we’re not in a position to support SPF records.
+```
+v=spf1 include:spf.example.com include:spf.pantheon.io -all
```
greg-1-anderson: @garydylina suggested that most folks will probably use `~all`, so perhaps we should use that here as well.

source/_docs/email.md
```diff
@@ -138,11 +138,20 @@ We strongly recommend that you do not use the local MTA (postfix) as described [
...
+```
+
+Adjust the above example record as needed for your domain:
+
+ - Be sure that you replace `spf.example.com` with the SPF record for the domain that your site is set up to use in the `From` address of outgoing messages.
```
greg-1-anderson: How about:

Be sure that you replace `include:spf.example.com` with the appropriate list of mail relays that also send email for your domain.

source/_docs/email.md
```diff
@@ -138,11 +138,20 @@ We strongly recommend that you do not use the local MTA (postfix) as described [
...
+ - If an SPF record exists for that domain, then add just the `include:spf.pantheon.io` part to whatever is already there, keeping the rest unchanged.
```
greg-1-anderson: Maybe this should just be merged in with the previous bullet, or removed. It seems redundant as a separate bullet.

source/_docs/email.md
alexfornuto: I didn't realize that `-all` and `~all` were different, will adjust.

greg-1-anderson:
👍 LGTM!

PR #4050, opened 2018-09-06
Note the error shown in D8 when there's a trusted host patterns mismatch

sarahg:
## Effect
- Notes the error shown in Drupal 8 when there's a trusted host patterns mismatch.
- GitHub decided to standardize line endings here as well, hence the big diff (the addition I made was the "Error: "The provided host name is not valid for this server" note in the Troubleshooting section)

## Remaining Work
- [x] copy review

## Post Launch
- [ ] Update Status Report


alexfornuto:
@sarahg isn't the information you've added already covered [here](https://pantheon.io/docs/settings-php/#trusted-host-setting)?

sarahg:
@alexfornuto Sort of — what's already documented is how to configure settings.php to prevent the error in the first place. 

I'm hoping to have something people could uncover via search if they actually run into the "The provided host name is not valid for this server" error. We get a lot of support requests from people who see that error since it sounds sort of hosting-related (even though it's just Drupal).

PR #4056, opened 2018-09-06
Add Cygwin as an option for Windows users.

sparklingrobots:
We document Cygwin as a way to install Terminus on Windows, but it wasn't previously linked here.

Closes # n/a

## Effect
- Adds Cygwin as an option, with a link to our documentation.

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/terminus/install.md
+++ b/source/_docs/terminus/install.md
@@ -14,7 +14,7 @@ searchboost: 100
 ---
 Terminus is available for Mac OS X and Linux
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
@sparklingrobots we removed references to Cygwin in the docs because so many people have trouble with it and we don't want to support that implementation.

sparklingrobots:
@alexfornuto Ah, no worries. Should we remove this page, then? https://pantheon.io/docs/cygwin-windows/ Or perhaps add a note that Cygwin isn't supported?

PR #4064, opened 2018-09-11
Update ldap-and-ldaps.md

ttrowell:
Further clarify the use of LDAPS via PEG

Closes #

## Effect
- add FAQ about the code patch needed when using LDAPS/PEG
- add comments to code snippet
- change 'never' to Allow per Gary for putenv() statement
- add an alternate settings array when testing through PEG

## Remaining Work
- [ ] Get a review by Gary D.

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/ldap-and-ldaps.md
+++ b/source/_docs/ldap-and-ldaps.md
@@ -57,8 +57,8 @@ Ensure that your certificates do **not** have a password. There is an extremely 
 
 You can also specify additional configurations with putnev, such as whether to perform server certificate checks.
 
-    // LDAP - Never perform server certificate check in a TLS session.
-    putenv('LDAPTLS_REQCERT=never');
+    // LDAP - Allow server certificate check in a TLS session.
+    putenv('LDAPTLS_REQCERT=allow');
 
 
 ## Frequently Asked Questions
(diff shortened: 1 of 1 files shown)


ttrowell:
@garydylina This is a follow-up to our PEG debugging.  Could you please review to see if I included this properly?

garydylina:
Who holds these opinions at Pantheon? 
- We do not recommend using LDAP for single sign-on authentication as it is not secure.
- Pantheon does not support IP authentication schemes.
I don't feel these are true and would like to strike them and reword the sections that contain them.
Though it is possible to implement insecure authentication using LDAP, there is nothing inherently insecure about using LDAP for authentication.
I am not sure what IP authentication is other than whitelisting that is possible for customers that use PEG.  What would support for that be if it isn't using PEG in conjunction with IP whitelisting - which we clearly do support.  Again, it's possible for customers to use PEG in ways that result in insecure implementations, but there is nothing inherently wrong with IP whitelisting.

Also, the section about applying the patch for D7 does not presently support LDAPS without further patching.  Seems like we ought to patch the patch or call out that the patch doesn't support LDAPS.

ttrowell:
@alexfornuto passing this to you now to incorporate Gary's feedback.  Thanks!

source/_docs/ldap-and-ldaps.md
```diff
@@ -7,20 +7,16 @@ categories: []
...
 PHP on Pantheon includes LDAP using OpenLDAP, so no changes to the platform are necessary in order to enable LDAP on your Pantheon site.
 
-<div class="alert alert-info" role="alert">
-<h4 class="info">Note</h4>
-<p>Pantheon does not support IP authentication schemes. We recommend certificate-based authentication to be compatible with distributed application containers.</p>
-</div>
-
+<div class="alert alert-info" role="alert" markdown="1">
+#### Note {.info}
+Pantheon does not support IP authentication schemes, except when used as part of a [Pantheon Enterprise Gateway](/docs/pantheon-enterprise-gateway) configuration. We recommend certificate-based authentication to be compatible with distributed application containers.
```
garydylina: Where possible we should write about what Pantheon "does" vs what it "does not" do.  Can we try something like,
"Pantheon supports IP authentication schemes when implemented as part of a [Pantheon Enterprise Gateway](/docs/pantheon-enterprise-gateway) configuration."

PR #4074, opened 2018-09-14
Exclude double quotes from TOC anchors

rachelwhitton:
## Effect
- Fixes TOC anchors for headers that include double quote character by striping the character via JS

For example: `/docs/settings-php/#error-"the-provided-host-name-is-not-valid-for-this-server"` => `/docs/settings-php#error-the-provided-host-name-is-not-valid-for-this-server`

Note: Affected TOC links require the data-proofer-ignore HTML attribute when linked from another doc page. Otherwise, htmlproofer will report the link as broken due to the nature of the test (server side vs client side)

### DIFF
--- a/source/docs/assets/js/jquery.tocify.js
+++ b/source/docs/assets/js/jquery.tocify.js
@@ -376,7 +376,7 @@
 
             }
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
Thanks @rachelwhitton!

PR #4100, opened 2018-09-21
Remove request to notify support in advance of tests.

sparklingrobots:
We recently stopped asking customers to notify us in advance of pen testing, and I suspect this is also true of load testing. I removed the relevant line from the documentation, but it would probably be best to have @newtoid or @shakhossain review to ensure this is correct.

Closes # n/a

## Effect
- Removes request for customers to let support know in advance of load testing

## Remaining Work
- none

## Post Launch
- [ ] Update Status Report


newtoid:
+1 - the load testing we need to know about is scheduled during onboarding

PR #4104, opened 2018-09-25
Add <site> to header in Terminus Examples doc

rachelwhitton:
## Effect
- Rewrite first section of terminus example usage page based on feedback received from a client


alexfornuto:
We'll want to replace throughout Terminus documentation the use of `<site>.<env>` with `$site.$env` per our [Style Guide](https://pantheon.io/docs/style-guide/#variables), as the use of `<>` is confusing for some readers, who include it in their commands.

rachelwhitton:
@alexfornuto this work is based off feedback from a new user looking for `<site>.<env>` on this page after reading the [command reference table](https://pantheon.io/docs/terminus/commands/) and seeing it in terminus help output. I think enforcing variables is well served (to quote the style guide) "When writing multi-step processes" - this section is not a process and variables would be confusing in this context. From the perspective of a new user, seeing a header that aligns with Terminus output is a stronger indicator of relevance.

alexfornuto:
@rachelwhitton I take your point, and agree that variables are not needed in this example. What I _would_ like to consider is removing `<>` symbols from all Terminus documentation, both here and in Terminus itself.

The use of the `<>` symbols causes confusion for some users, who think their site name and environment go inside them instead of replacing them.

CC @TeslaDethray @kimby77 for input. If everyone's on board, I can make a sweep through both repos and update.

kimby77:
👍  I'm good with whatever change makes it work better for users!

alexfornuto:
Discussion with the WriteTheDocs community has taught me that the use of `<>` is more common than I thought. Rather than try to remove them, I will just make more clear wherever I can that the `<>` is to be replaced, not included.

With that aside, @rachelwhitton this PR is fantastic!

PR #4108, opened 2018-09-27
Adding info on data export limitations

tinefin:
## Effect
PR includes the following changes:
- Additional info on data exports. Tested, reproduced myself. Discussed and confirmed w dts.

Internal reference:
Ticket 111676

## Remaining Work
- Needs review, ping @davidstrauss 

## Post Launch
To be completed by the docs team upon merge: 
- [ ] Update Status Report
- [ ] Archive from **Done** in Waffle

PR #4141, opened 2018-10-04
Update platform-considerations.md

tinefin:
Clarifying that it is not the testing which is difficult - it is getting this to work at all.

Closes #

## Effect
-
-
-

## Remaining Work
- [ ] If needed

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/platform-considerations.md
+++ b/source/_docs/platform-considerations.md
@@ -36,9 +36,9 @@ In this or similar instances, consider performing larger operations locally, the
 
 ## Batched Data Export to File
 
-In Test and Live environments on plans with multiple application containers, it is difficult to test batched exports to a file. Many contrib plugins and modules are not designed to support multiple application containers. It might be possible to get data export working, but that is likely to require additional effort and custom code.
+For plans which have multiple application containers, bacthed data export to file is difficult. Many contrib plugins and modules are not well architected to support multiple application containers. It might be possible to get data export working, but is likely to require effort and custom code.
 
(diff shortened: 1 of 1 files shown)


source/_docs/platform-considerations.md
```diff
@@ -36,9 +36,9 @@ In this or similar instances, consider performing larger operations locally, the
 
 ## Batched Data Export to File
 
-In Test and Live environments on plans with multiple application containers, it is difficult to test batched exports to a file. Many contrib plugins and modules are not designed to support multiple application containers. It might be possible to get data export working, but that is likely to require additional effort and custom code.
+For plans which have multiple application containers, bacthed data export to file is difficult. Many contrib plugins and modules are not well architected to support multiple application containers. It might be possible to get data export working, but is likely to require effort and custom code.
```
alexfornuto: @tinefin I removed "well architected" from the original PR because "architected" is not _really_ an English word, and can be confusing, especially when run through translators.

alexfornuto:
@EdwardAngert  See #4108 for reference to original submission.

source/_docs/platform-considerations.md
tinefin: Ah. Well, it is not just that word which has changed though. The meaning of the paragraph has changed so it is now saying something fairly different / incorrect.

alexfornuto: Well I can see that er added "test" and should remove that, but a lot of it was just switching from passive to active voice. I've assigned @EdwardAngert to take a pass at this to do so again, while keeping the message the same.

PR #4159, opened 2018-10-10
Cancellation: No partial refund

thomas-thackery:
## Effect
- Mention of no partial refund for unused service. Language comes from similar mention on annual plan. 

## Remaining Work
- [ ] Our TOS §11.1 Fees fleshes this out, but there is no direct anchor to this section, https://pantheon.io/terms-of-service. 
- [ ] Consider leaving the mention at the section head of Cancellation and possibly also at the Downgrade section as this will be where CSE will send customers to make the change. Were the mention lower in the doc, it runs the risk of being missed entirely. 
- [] Reach out to Product to get this information on the Downgrade screen itself. 

## Post Launch
- [ ] Update Status Report


alexfornuto:
P.S. @thomas-thackery I reached out to the website team to have ID values added to the TOS section headers so that we can link to them in the future.

PR #4164, opened 2018-10-11
D8 Composer WO CI - Style alignment updates.

alexfornuto:
Closes: NA

## Effect
- Update guide copy for alignment with the style guide.
- Update / Replace terminal screenshots

## Remaining Work
- [ ] Copy review (@EdwardAngert)

## Post Launch
- [ ] Expire cache for updated images.
- [ ] Update Status Report


source/_docs/guides/drupal-8-composer-no-ci.md
```diff
@@ -112,37 +112,35 @@ Normally the next step would go through the standard Drupal installation. But si
...
 
-5. Set the site to git mode:
+5. Set the site to `git` mode:
 
    ```bash
    terminus connection:set $PANTHEON_SITE_NAME.dev git
    ```
 
-6. Add and commit the code files. A git force push is necessary because we are writing over the empty repository on Pantheon with our new history that was started on the local machine. Subsequent pushes after this initial one should not use `--force`:
+6. Add and commit the code files. A Git `force` push is necessary because we are writing over the empty repository on Pantheon with our new history that was started on the local machine. Subsequent pushes after this initial one should not use `--force`:
```
alexfornuto: If you're gonna encapsulate `force` as a command, should you include the `--` prefix?

PR #4190, opened 2018-10-23
Add overflow to ToC CSS

### LINKED ISSUE #3196: Make guides navigation independently scrollable
Re: Guides navigation

## Issue Description
On the Terminus Manual I noticed that the navigation on the left side, you can't see the options on the bottom without scrolling all the way down the main page. This would apply to any Doc Guide using this pattern (Quick Start guide, etc...). 

Example page: https://pantheon.io/docs/terminus/commands/

image

## Suggested Resolution
To improve usability of navigation, I suggest we make the navigation area independently scrollable. When a user has their mouse hovered over the navigation, I'd expect they'd be able to scroll up/down on the navigation without scrolling the main content.  Open to other suggestions as well!

alexfornuto:
Closes #3196 

## Effect
- Makes ToC div scroll independently of main doc when the browser window is shorter than the ToC.

## Remaining Work
- [ ] Review from @kimby77 
- [ ] Review from @rachelwhitton 

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/docs/assets/css/main.css
+++ b/source/docs/assets/css/main.css
@@ -3355,6 +3355,7 @@ div.col-md-3.manual-guide-toc {
   padding-right: 0px;
   color: #1e1e2b;
   max-height: 100vh;
+  overflow: auto;
 }
 .invisible-manual-toc {
   background: #fff !important;
@@ -3468,6 +3469,7 @@ padding-left:30px;
 }
 .manual-guide-toc {
   padding:0px;
+  overflow: auto;
 }
 
 .manual-body {
--- a/source/docs/assets/css/main.css
+++ b/source/docs/assets/css/main.css
@@ -3355,7 +3355,8 @@ div.col-md-3.manual-guide-toc {
   padding-right: 0px;
   color: #1e1e2b;
   max-height: 100vh;
-  overflow: auto;
+  overflow-y: auto;
+  overflow-x:hidden;
 }
 .invisible-manual-toc {
   background: #fff !important;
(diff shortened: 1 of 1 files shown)


alexfornuto:
@kimby77 the [Terminus Command Reference](https://3196scroll-static-docs.pantheonsite.io/docs/terminus/commands/) page in the Multidev environment should demonstrate the behavior your were looking for.

kimby77:
🎉  You have made my day!  Except it's not working for me on the multidev :/  I'm on the latest version of chrome.

drainpip:
@alexfornuto - the problem is the dynamic height of the `.manual-guide-toc` container. In order for the browser to know when to show a scrollbar, it needs a solid height to base itself on. You can play with it by making `.manual-guide-toc` have a `height: 300px` or something to see the scrollbar working properly.

The solution isn't so straightforward I don't think. Basically what you need is another element that contains just the menu and title with overflow on and a set height. Of course the set height will change so you can't just pick a number.

There's likely JS libraries that help out with this exact thing, but you could probably hack together something that adds up the heights of all the parts of the menu and set that height to the new child of `.manual-guide-toc` with the overflow... If I were you, I'd probably find a JS library that specializes in this and call it a day.

drainpip:
A JS library will affect performance in some way, how much I can't say as it would need to be measured. Might be negligible.

A quick script that calculates the height and such will be very minimal... 

There's already a bit of JS stuff going on with the menu setting things active and whatnot... page complete isn't the fastest as it is so it's likely not going to be noticed either way.

alexfornuto:
The team as decided to merge this PR as is, and make a note to more fully resolve this issue as part of a larger project to update our static site generator and refactor the code.

PR #4197, opened 2018-10-24
align with pricing page

allenfear:


alexfornuto:
Confirmed with marketing, thanks @allenfear!

PR #4204, opened 2018-10-30
Fix tabs in custom upstream doc

rachelwhitton:
## Effect
- Remove active class from GitLab tab pane so that it is not displayed until the tab is selected


alexfornuto:
Nice fix, TY @rachelwhitton.

PR #4213, opened 2018-10-31
Scripting Terminus

### LINKED ISSUE #3086: New Doc Proposal: Add "Scripting with Terminus" to Terminus Manual
We have a lot of people writing scripts with Terminus, and a lot of best practices are known, but there's no doc page for it!

It would be a nice thing to add to the terminus guide, and we could probably source an outline based on the ACE team's experiences in Jump Starts and such.

cc @ataylorme @stevector @tessak22

alexfornuto:
Closes #3086 

## Effect
- Adds a new page to the Terminus Manual covering scripting Terminus

## Remaining Work
- [x] Review / additions / suggestions from @pantheon-systems/devrel
- [ ] Copy Review

## Post Launch
- [ ] Include/exclude pages ^ respectively within docs search service provider
- [ ] Update Status Report

### DIFF
--- a/source/_docs/terminus/scripting.md
+++ b/source/_docs/terminus/scripting.md
@@ -1,0 +1,81 @@
+---
+title: Terminus Manual
+subtitle: Scripting Terminus
+description: Automate your workflow with Terminus.
+terminuspage: true
+type: terminuspage
+layout: terminuspage
+nexturl: terminus/plugins/
+previousurl: terminus/commands/
+permalink: docs/terminus/:basename/
+image: terminus-thumbLarge
+searchboost: 100
+---
+
+While Terminus makes it easy to directly interact with Pantheon from your command line, its real value is in scripting. By adding Terminus to your automated tasks, you can truly make Pantheon a part of your standardized workflow.
+
+Consider the repetitive tasks you perform using the Pantheon Dashboard:
+
+ - Can those tasks be executed by Terminus commands?
+ - Can the values required by the commands be derived programmatically?
+
+If so, you consider how you can turn the task into a script.
+
+## Authentication
+
+Terminus needs to be authenticated in order to execute most commands. Before running any script you must ensure Terminus is authenticated with a [machine token](/docs/terminus/install#machine-token) that has the proper permissions
+
+## Examples
+
+ - The [Pantheon Example Terminus Auto Update Script](https://github.com/pantheon-systems/example-terminus-auto-update-script){.external} demonstrates how you can use Terminus to automate plugin and theme updating for multiple sites.
+
+
+ - The [Example WordPress Composer](https://github.com/pantheon-systems/example-wordpress-composer){.external} repository (used by our [Build Tools](/docs/guides/build-tools/) guide) uses Terminus to [deploy staged changes](https://github.com/pantheon-systems/example-wordpress-composer/blob/46ff34e2b9f421a1c0eae72ade80376e8dd42f31/.circleci/deploy-to-pantheon.sh){.external} to multidev environments.
+
+
+
+## Bash Variables
+
+One of the ways Terminus can be used in scripting is the generation of variables. In the example below, we use the output of `terminus multidev:list` to create an environment variable with all our multidev environments:
+
+```bash
+PANTHEON_MULTIDEV_LIST="$(terminus multidev:list -n ${TERMINUS_SITE} --format=list --field=Name)"
+```
+
+This example assumes the variable `TERMINUS_SITE` is already set. Now you can iterate through `$PANTHEON_MULTIDEV_LIST` using something like a `while read` loop to perform tasks on each multidev environment.
+
+
+Here's a more complete example. This script goes through every site in an organization, skips any that are frozen, and takes a backup of the Live environment:
+
+```bash
+#!/bin/bash
+
+# Exit on error
+set -e
+
+# Stash org UUID
+ORG_UUID="722c5f3b-....-6c8a"
+
+# Stash list of all Pantheon sites in the org
+PANTHEON_SITES="$(terminus org:site:list -n ${ORG_UUID} --format=list --field=Name)"
+
+# Loop through each site in the list
+while read -r PANTHEON_SITE_NAME; do
+    # Check if the site is frozen
+    IS_FROZEN="$(terminus site:info $PANTHEON_SITE_NAME --field=frozen)"
+
+    # If the site is frozen
+    if [[ "true" == "${IS_FROZEN}" ]]
+    then
+        # Then skip it
+        echo -e "Skipping a backup of the site '$PANTHEON_SITE_NAME' because it is frozen...\n"
+    else
+        # Otherwise create a backup of the live environment of the site
+        echo -e "Creating a backup of the live environment for the site '$PANTHEON_SITE_NAME'...\n"
+        terminus backup:create $PANTHEON_SITE_NAME.live
+    fi
+done <<< "$PANTHEON_SITES"
+```
+
+This script requires that you set the variable `ORG_UUID` within the script itself. You can find the UUID using `terminus org:list`.
+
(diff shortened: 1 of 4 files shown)


source/_docs/terminus/scripting.md
```diff
@@ -0,0 +1,44 @@
...
+type: terminuspage
+layout: terminuspage
+nexturl: terminus/plugins/
+previousurl: terminus/commands/
+permalink: docs/terminus/:basename/
+image: terminus-thumbLarge
+searchboost: 100
+---
+
+While Terminus makes it easy to manually manipulate the Pantheon environment from your command line, its real value is in scripting. By adding Terminus to your automated tasks, you can truly make Pantheon a part of your standardized workflow.
```
ataylorme: > manually manipulate the Pantheon environment

Can we change to "interact with Pantheon"

source/_docs/terminus/scripting.md
```diff
@@ -0,0 +1,44 @@
...
+## Authentication
+
+To streamline your use of Terminus in scripts, make sure that it's authenticated first. The recommended method is to use a [machine token](/docs/terminus/install#machine-token).
+
+## Examples
+
+ - The [Pantheon Example Terminus Auto Update Script](https://github.com/pantheon-systems/example-terminus-auto-update-script){.external} demonstrates how you can use Terminus to automate plugin and theme updating for multiple sites.
+
+
+ - The [Example WordPress Composer](https://github.com/pantheon-systems/example-wordpress-composer){.external} uses Terminus to [deploy staged changes](https://github.com/pantheon-systems/example-wordpress-composer/blob/master/.circleci/deploy}to-pantheon.sh){.external} to multidev environments.
```
ataylorme: > `[deploy staged changes](https://github.com/pantheon-systems/example-wordpress-composer/blob/master/.circleci/deploy}to-pantheon.sh)`

There seems to be an incorrect character with a `}` instead of `-` in the URL. With GitHub links we should probably also link to a specific commit hash, rather than `master` in case the repository changes in the future.

> The Example WordPress Composer uses

It seems like this is missing a word. Perhaps "The Example WordPress Composer project uses". I think this is also a good opportunity to link folks to our [build tools guide](https://pantheon.io/docs/guides/build-tools/). Following that guide will allow them to deploy the example.

source/_docs/terminus/scripting.md
```diff
@@ -0,0 +1,44 @@
...
+
+
+## Bash Variables
+
+One of the ways Terminus can be used in script is the generation of environment variables. In the example below, we use the output of `terminus multidev:list` to create an environment variable with all our multidev environments:
+
+```bash
+PANTHEON_MULTIDEV_LIST="$(terminus multidev:list -n ${TERMINUS_SITE} --format=list --field=Name)"
```
ataylorme: We should probably expand on this. _Why_ would you want to store a list of multidev environments in a bash variable? We need to show the value here - being able to leverage that list to complete a larger task.

I think looping through multiple sites would make an even better example. Here is an example script looping through all sites in an org and making a backup on the live environment. It could be extended to download the backup and store it offsite.

```bash
#!/bin/bash

# Exit on error
set -ex

# Stash list of all Pantheon sites in the org
PANTHEON_SITES="$(terminus org:site:list -n ${ORG_UUID} --format=list --field=Name)"

# Loop through each site in the list
while read -r PANTHEON_SITE_NAME; do
    # Create a backup of the live environment of the site
    echo -e "Creating a backup of the live environment for the site '$PANTHEON_SITE_NAME'...\n"
    terminus backup:create $PANTHEON_SITE_NAME.live
done <<< "$PANTHEON_SITES"
```

This example assumes `ORG_UUID` is set to an org UUID retrieved from `terminus org:list`

source/_docs/terminus/scripting.md
```diff
@@ -0,0 +1,44 @@
...
+Consider the repetitive tasks you perform using the Pantheon Dashboard:
+
+ - Can those tasks be executed by Terminus commands?
+ - Can the values required by the commands be derived programmatically?
+
+If so, you consider how you can turn the task into a script.
+
+## Authentication
+
+To streamline your use of Terminus in scripts, make sure that it's authenticated first. The recommended method is to use a [machine token](/docs/terminus/install#machine-token).
```
ataylorme: > To streamline your use of Terminus in scripts, make sure that it's authenticated first. 

Maybe something like "Terminus needs to be authenticated in order to execute most commands. Before running any script you must ensure Terminus is authenticated with a machine token that has the proper permissions"

source/_docs/terminus/scripting.md
alexfornuto: > With GitHub links we should probably also link to a specific commit hash, rather than master in case the repository changes in the future.

I see how linking to that file at a specific commit will make the link evergreen. But if that file is changed to the point that it no longer reflects what the doc is describing, shouldn't the doc be updated? After all, if the methodology in the repository is updated it's probably for a good enough reason that the documentation should _also_ be updated.

source/_docs/terminus/scripting.md
ataylorme: I added that line as an example of how to set the variable as folks might not know how to set a bash variable. I think it is still worth noting in the doc since if line 7 isn't updated the script will fail.

source/_docs/terminus/scripting.md
ataylorme: > I see how linking to that file at a specific commit will make the link evergreen

Yes, we should make the link evergreen.

> if that file is changed to the point that it no longer reflects what the doc is describing, shouldn't the doc be updated? After all, if the methodology in the repository is updated it's probably for a good enough reason that the documentation should also be updated.

Yes but I think it is better to have an evergreen link and make a PR to update the doc with a new link if the repo changes rather than having a potentially broken/confusing link in the doc.

source/_docs/terminus/scripting.md
```diff
@@ -0,0 +1,81 @@
...
+ - The [Pantheon Example Terminus Auto Update Script](https://github.com/pantheon-systems/example-terminus-auto-update-script){.external} demonstrates how you can use Terminus to automate plugin and theme updating for multiple sites.
+
+
+ - The [Example WordPress Composer](https://github.com/pantheon-systems/example-wordpress-composer){.external} repository (used by our [Build Tools](/docs/guides/build-tools/) guide) uses Terminus to [deploy staged changes](https://github.com/pantheon-systems/example-wordpress-composer/blob/46ff34e2b9f421a1c0eae72ade80376e8dd42f31/.circleci/deploy-to-pantheon.sh){.external} to multidev environments.
+
+
+
+## Bash Variables
+
+One of the ways Terminus can be used in script is the generation of environment variables. In the example below, we use the output of `terminus multidev:list` to create an environment variable with all our multidev environments:
```
ataylorme: > One of the ways Terminus can be used in script is the generation of environment variables. In the example below, we use the output of terminus multidev:list to create an environment variable with all our multidev environments

"can be used in script" doesn't sound quite right. Perhaps "used in scripts" (plural) or "used in scripting"

Also, it should be just "variables", not "environment variables". Environment variables are variables set by the environment (like CircleCI itself), not by the user.

alexfornuto: Resolved in f1c1067

mcdwayne:
Reviewed the changes just now and think this is a great jumping off point.  I keep thinking there might be an easier example script to get people encouraged but am not sure what that might be.  
The example at the end is solid though and a real use case people will find useful for their non-prod sites. 

tl:dr - 👍

source/_docs/terminus/scripting.md
```diff
@@ -0,0 +1,81 @@
...
+Here's a more complete example. This script goes through every site in an organization, skips any that are frozen, and takes a backup of the Live environment:
+
+```bash
+#!/bin/bash
+
+# Exit on error
+set -e
+
+# Stash org UUID
+ORG_UUID="722c5f3b-8ba8-423a-ae44-d72daefd6c8a"
```
ataylorme: @alexfornuto can we put a placeholder here rather than a real org UUID?

PR #4214, opened 2018-10-31
Update cookies.md

mathetos:
## Effect
- Adds a sentence about cookie naming prefix for WordPress websites.

## Remaining Work
- [ ] Technical Review
- [ ] Copy Review 

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/cookies.md
+++ b/source/_docs/cookies.md
@@ -6,7 +6,7 @@ categories: []
 ---
 
 ## Cookie Naming Requirements
-For cookies to function on cached pages they must match the pattern: `STYXKEY[a-zA-Z0-9_-]`.
+For cookies to function on cached pages they must match the pattern: `STYXKEY[a-zA-Z0-9_-]`. Note that on WordPress sites, if you want session cookies to be excluded from caching, they must be named with `wp-` as their prefix.
 
 ## Disable Caching for Specific Pages
 You can use regular expression(s) to determine if the current request (`$_SERVER['REQUEST_URI']`) should be excluded from cache. If the request matches, bypass cache by setting the `NO_CACHE` cookie in the response.


alexfornuto:
Thanks for the submission! I've added back the items from the PR template, which are there for the docs team's benefit and should not be removed.

source/_docs/cookies.md
```diff
@@ -6,7 +6,7 @@ categories: []
 ---
 
 ## Cookie Naming Requirements
-For cookies to function on cached pages they must match the pattern: `STYXKEY[a-zA-Z0-9_-]`.
+For cookies to function on cached pages they must match the pattern: `STYXKEY[a-zA-Z0-9_-]`. Note that on WordPress sites, if you want session cookies to be excluded from caching, they must be named with `wp-` as their prefix.
```
danielbachhuber: Is this a recent infrastructure change? I know that WordPress core cookies have a `wp-` prefix but I thought Varnish whitelisted them explicitly.

mathetos: For context, we had a Give customer report a problem with sessions on Pantheon. I confirmed the problem in a new Pantheon environment. We had a Pantheon support technician make the suggestion that Give using `wp_` as the prefix was the problem, I tested that hypothesis by changing our prefix and it resolved the problem both on my new environment and for the customer. So, in my experience between all of those points of contact it's definitely the case. But I can't speak to internal documented changes that Pantheon might have. Thanks for reviewing!

mathetos: Here's our Give issues in this regard:
* https://github.com/impress-org/give/issues/3819
* https://github.com/impress-org/give/issues/3705#issuecomment-423485726

sarahg: I think two types of cookies are being conflated here:

- STYXKEY cookies are [cache-varying](https://pantheon.io/docs/cookies/#cache-varying-cookies).
- Cookies with the `wp-[A-Za-z0-9_]+` prefix are [cache-busting](https://pantheon.io/docs/caching-advanced-topics/#cookie-handling). There are several other cookie prefixes that bust cache.

There are Varnish rules for both types of these cookies, but they accomplish different things.

I'd suggest:
- Remove or clarify this "Cookie Naming Requirements" section since it only refers to STYXKEY-style cookies.
- Link or copy the cache-busting cookie naming patterns from the Advanced Caching doc ([this](https://pantheon.io/docs/caching-advanced-topics/#cookie-handling)) over to the Cookies doc.

source/_docs/cookies.md
```diff
@@ -6,7 +6,7 @@ categories: []
 ---
 
 ## Cookie Naming Requirements
-For cookies to function on cached pages they must match the pattern: `STYXKEY[a-zA-Z0-9_-]`.
```
greg-1-anderson: The pattern should actually be `STYXKEY[a-zA-Z0-9_-]+` (n.b. `+` on end)

source/_docs/cookies.md
greg-1-anderson: Yes, 👍 on @sarahg's suggestion; the cache-busting cookies should not be described in the same section as the cache-varying cookies. (At the moment only STYXKEY is cache-varying, but there may be more of these in the future). Deleting the section called "Cookie Naming Requirements" makes sense, as this information is duplicated further down on the page. Referencing the Advanced Caching page as suggested is also a good idea.

alexfornuto:
@mathetos your commit was merged as part of #4265, but not reflected in this PR due to a rebase. Thanks for your contribution!

mathetos:
Awesome, happy to help!

PR #4218, opened 2018-11-02
Put limit in the large file recommended plugins in wp

### LINKED ISSUE #4217: Platform Considerations Doc Update recommendation for files larger than 100mb
Re: [Platform Considerations](https://pantheon.io/docs/platform-considerations/)

Priority: Medium

## Issue Description:

We have recommended WP plugins when using large files in the platform 
https://pantheon.io/docs/platform-considerations/#large-files but it seems it is limited to 100mb only

## Suggested Resolution

- Add that those plugins will still respect the max file upload limit of 100mb
- Maybe add additional recommendations on how to handle files larger than 100mb

carl-alberto:
Closes #4217

## Effect
- Wp recommended plugins will still respect the max file upload limit of 100mb

## Remaining Work
- [ ] Technical Review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/platform-considerations.md
+++ b/source/_docs/platform-considerations.md
@@ -110,6 +110,12 @@ It is also possible to deliver smaller media files from Pantheon using [progress
 
 Due to the configuration of the [Pantheon Filesystem](/docs/files/), Pantheon's file serving infrastructure is not optimized to store and deliver very large files. Files over 100MB cannot be uploaded through WordPress or Drupal, and must be added by [SFTP or rsync](/docs/rsync-and-sftp/). Files over 256MB will fail no matter how they are uploaded. Transfers with files over 50MB will experience noticeable degradation in performance.
 
+| File Size       | Platform Compatibility            | Notes                               |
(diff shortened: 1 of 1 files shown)


alexfornuto:
> In my understanding there are 3 sets of file size range with different solutions needs to be clarified in this doc:
> 
> 1. less than 100mb > can be done through CMS using the plugins suggested
> 2. 100mb - 256mb > needs to be ftped/rsynced then optionally use this plugin https://wordpress.org/plugins/add-from-server/ for WP sites if they want to manage the files from the CMS
> 3. 256mb > cannot be done using the first 2 methods and they need to upload directly to a 3rd party service like s3 and manually use the urls generated from them, this method also needs to be clarified that they cannot do the file management via the CMS
> 
> Not sure on how to properly clarify these points in the existing docs so customers can plan beforehand on which method they will use

Sounds like a Table might be a good solution here.

alexfornuto:
image

@carl-alberto how's this grab ya?

PR #4220, opened 2018-11-02
Use email instead of user ID in the password reset example

sarahg:
## Effect
- Adjust the example WP-CLI password reset command to use an email instead of a user ID. This also works and folks are more likely to know their email than their WP user ID.

## Remaining Work
- none

## Post Launch
- [ ] Update Status Report


alexfornuto:
Works for me, thanks @sarahg!

PR #4224, opened 2018-11-05
Update redis.md

ccjjmartin:
Closes NA

## Effect
Redis as of 3.13 supports compressing the cache using the `Redis_CacheCompressed` class.  This saves a whopping 80% of memory on Redis for a small tradeoff on CPU useage.  For large sites this is huge, especially when trying to fight high Redis turnover rates.  I implemented this on one of our large sites on Pantheon and noticed an actual 75% memory reduction.  I hope to increase awareness of this great feature.

Redis Docs: https://www.drupal.org/project/redis/releases/7.x-3.13

## Remaining Work
- [ ] Technical Review
- [ ] Copy Review (if Technical Edits make any copy changes, else ignore)

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/redis.md
+++ b/source/_docs/redis.md
@@ -194,7 +194,7 @@ All plans except for a Basic plan can use Redis. Redis is available to Sandbox s
       // If you've installed your plugin in a contrib directory, use this line instead:
       // $conf['cache_backends'][] = 'sites/all/modules/contrib/redis/redis.autoload.inc';
 
-      $conf['cache_default_class'] = 'Redis_Cache';
+      $conf['cache_default_class'] = 'Redis_CacheCompressed';
       $conf['cache_prefix'] = array('default' => 'pantheon-redis');
 
       // Do not use Redis for cache_form (no performance difference).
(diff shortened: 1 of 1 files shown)


alexfornuto:
@ccjjmartin thanks for the contribution! To clarify, were you running the Redis provided by Pantheon, or using an externally hosted Redis?

alexfornuto:
Related: `UNIPAN-18`

ccjjmartin:
@alexfornuto I was running the Redis that is Pantheon hosted.

alexfornuto:
Thanks @ccjjmartin. Is looks like the version `3.13` you're referring to is the version of the Drupal module, not the Redis server itself. That clears up some confusion on my end.

ccjjmartin:
@sarahg Let me know if you think we should include a note somewhere here explaining which Redis class to use based on Redis module version.

I was also thinking that this PR may open up some technical review on the backend / advertising to other Pantheon clients since this great feature would reduce the load on the Redis servers a lot.

sarahg:
This is very cool, thanks for the suggestion @ccjjmartin! I'm a +1 for recommending this setting for D7. The [D8 version](https://www.drupal.org/project/redis/issues/2991121) is still in-progress. 

I think we should note that you need the Redis module version 7.x-3.13 or newer for it work, yes, and maybe add a "Troubleshooting" note for the "class not found" errors you'd see if using this with an older version.

We may also want to recommend setting `cache_compression_size_threshold` when using this, though, since we've seen some cases where using this for all objects can be a performance hit.

Let's also try to get a review from someone in engineering before merging this one. @davidstrauss maybe?

alexfornuto:
Thanks @ccjjmartin, and thanks @sarahg for reviewing!

PR #4231, opened 2018-11-07
Simplify language around sandboxes

alexfornuto:


alexfornuto:
+1 in Slack from @allenfear

PR #4248, opened 2018-11-15
Reference Drush and WP-CLI under script execution time

### LINKED ISSUE #4243: Timeouts on Pantheon Doc Update - Add a timeout for Drush/WP-cli commands
Re: [Timeouts on Pantheon](https://pantheon.io/docs/timeouts/)

Priority: Low

## Issue Description:

In our timeouts doc, it is unclear what would be the timeout for Drush/WP-cli commands. I asked from #cse-training channel and Alex D. responded with the following:

"_drush/wp-cli has a maximum execution time of 120 seconds_"

https://pantheon.slack.com/archives/C0ACN422E/p1542017693215300

## Suggested Resolution

I think we should also put an entry under "_Timeouts That are not Configurable_" section for Drush/WP-cli commands(120 secs)

Related chat: 19501645526

alexfornuto:
Closes #4243 

## Effect
- Explicitly states that Drush and WP-CLI will hit the max script execution limit.

## Remaining Work
- [x] Review from @ruthiehallarsis to make sure that resolves her concerns.
- [x] Technical review (@alexdicianu)
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/timeouts.md
+++ b/source/_docs/timeouts.md
@@ -21,7 +21,7 @@ Rules are for the good of the group, and timeouts are no exception. We've config
   <tr>
     <td><a href="https://secure.php.net/manual/en/info.configuration.php#ini.max-execution-time">PHP max_execution_time</a></td>
     <td>120 seconds</td>
-    <td>Maximum time a script can run before being terminated by the parser. Helps prevent poorly written scripts from tying up your application container's PHP workers.
+    <td>Maximum time a script can run before being terminated by the parser. This included Drush & WP-CLI commands. Helps prevent poorly written scripts from tying up your application container's PHP workers.
     <br/><br/>
      You can reduce this timeout to be less than 120 seconds via <code>settings.php</code> or <code>wp-config.php</code>, but you cannot increase it beyond 120 seconds.</td>
   </tr>
--- a/source/_docs/timeouts.md
+++ b/source/_docs/timeouts.md
@@ -21,9 +21,9 @@ Rules are for the good of the group, and timeouts are no exception. We've config
   <tr>
     <td><a href="https://secure.php.net/manual/en/info.configuration.php#ini.max-execution-time">PHP max_execution_time</a></td>
     <td>120 seconds</td>
-    <td>Maximum time a script can run before being terminated by the parser. This included Drush & WP-CLI commands. Helps prevent poorly written scripts from tying up your application container's PHP workers.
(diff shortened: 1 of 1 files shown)


ruthiehallarsis:
Hey @alexfornuto this looks good to me. We can proceed with the merging.

alexfornuto:
Thanks @ruthiehallarsis. I'd actually like to get @alexdicianu's input, to say if we should instead include these tools under `PHP set_time_limit` instead of `PHP max_execution_time`.

source/_docs/timeouts.md
```diff
@@ -21,7 +21,7 @@ Rules are for the good of the group, and timeouts are no exception. We've config
   <tr>
     <td><a href="https://secure.php.net/manual/en/info.configuration.php#ini.max-execution-time">PHP max_execution_time</a></td>
     <td>120 seconds</td>
-    <td>Maximum time a script can run before being terminated by the parser. Helps prevent poorly written scripts from tying up your application container's PHP workers.
+    <td>Maximum time a script can run before being terminated by the parser. This included Drush & WP-CLI commands. Helps prevent poorly written scripts from tying up your application container's PHP workers.
```
alexdicianu: > This included Drush & WP-CLI commands.
I believe this is correct and the max execution time does include Drush & WP-CLI commands. This is just a default, as in the case of cli commands (drush/wp-cli) it can easily be overwritten via `ini_set` to accommodate longer running jobs which are expected from time to time. If we're going to mention drush/wp-cli I think it would be good to say that the defaults can be overwritten if needed.

alexdicianu:
Actually, this section under the `User-Configurable Timeouts` is a bit confusing.
```
Maximum time a script can run before being terminated by the parser. This included Drush & WP-CLI commands. Helps prevent poorly written scripts from tying up your application container's PHP workers.

You can reduce this timeout to be less than 120 seconds via settings.php or wp-config.php, but you cannot increase it beyond 120 seconds.
```
This assumes the user is visiting over http(s) in which case this is true. The connection is going to get killed much sooner actually due to the GCDN 60-second timeout. For php in CLI mode (i.e. drush/wp-cli) this does not hold as the timeout can be increased. We should probably differentiate http(s) visits from cli ones and make it clear that drush can run for a longer period of time if configured to do so.

alexfornuto:
@alexdicianu I think I've addressed your comment.

alexdicianu:
LGTM 👍

PR #4272, opened 2018-11-22
Update private-paths.md

tinefin:
Only one of the 3 private path options were mentioned, which causes confusion:
https://www.drupal.org/project/ubercart/issues/2842813
 And - internal ticket id: 119859

Closes #

## Effect
- Make sure both options for private path is mentioned, sites/default/files/private too:
 Previously we only mentioned the version controlled path, which is rarely used for encryption purposes as then the encryption key will be committed to git (security issue). 
The version controlled private path is likely more relevant for other private path use cases, but. People should choose themselves.
- Simplify the "between environments" part. This is really just normal deployment process on Pantheon.
- Remind folks to use relative paths as otherwise things break on container architecture (container id is part of the absolute path, relative path will work anywhere)
- Make it obvious that this is normal drupal config which can also happen in Drupal admin UI
- Mention other steps needed in order to enable the private paths: set variable in DB + create folder + upload key.
We regularly have tickets where folks expected the private directory to be created for them for instance.
- Remind folks to do this on all environments.
- Make the difference between the 2 private directories clear. Version controlled and not version controlled.

## Remaining Work
- [ ] If needed

## Post Launch
- [ ] Update Status Report


source/_docs/private-paths.md
```diff
@@ -65,22 +65,27 @@ If you have a private code library that needs to have a specific sub-directory e
...
 $: terminus drush <site>.<env> -- vget uc_credit_encryption_path
-uc_credit_encryption_path: 'private'
+uc_credit_encryption_path: <check this matches your setting>
 ```
 
+In the above, `<my_private_path>` can be set to either of these non-web accessible private directories:
+`'sites/default/files/private'` (preferred)
+`'private'` (version controlled)
+More about private paths here:
+https://pantheon.io/docs/private-paths/
```
alexfornuto: This is a link to the current doc. Did you mean to send them somewhere else?

source/_docs/private-paths.md
```diff
@@ -1,87 +1,99 @@
...
+    - `'private'` (version controlled)
+
+
+2. Verify that `uc_credit_encryption_path` is set correctly:
+
+   ```bash
+   terminus drush <site>.<env> -- vget uc_credit_encryption_path
+   ```
+
+3. Check that the output of `uc_credit_encryption_path:` matches your settings.
```
alexfornuto: is `uc_credit_encryption_path:` a shell command?

source/_docs/private-paths.md
tinefin: You are absolutely right! :)
Line 84 and 85 can be removed.

source/_docs/private-paths.md
tinefin: `uc_credit_encryption_path` is a code level variable, set in the database.

The drush vset command sets the variable:
`terminus drush <site>.<env> -- vset uc_credit_encryption_path`

The drush vget command checks what the variable is currently set to:
`terminus drush <site>.<env> -- vget uc_credit_encryption_path`

PR #4298, opened 2018-12-05
allow to enable redis without module enabled

heddn:
## Effect
- documents method to add redis without the module enabled.

## Remaining Work
- [ ] Technical review
- [ ] Copy review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/redis.md
+++ b/source/_docs/redis.md
@@ -135,10 +135,6 @@ All plans except for a Basic plan can use Redis. Redis is available to Sandbox s
 1. Enable the Redis cache server from your Pantheon Site Dashboard by going to **Settings** > **Add Ons** > **Add**. It may take a couple minutes for the Redis server to come online.
 2. Install and activate the [Redis](https://www.drupal.org/project/redis){.external} module from Drupal.org.
 
-    <div class="alert alert-info">
-    <h4 class="info">Note</h4><p markdown="1">You **must** activate the module before proceeding.
-    </p></div>
-
     You can install and enable the module from the command line using [Terminus](/docs/terminus):
 
         terminus remote:drush <site>.<env> -- en redis -y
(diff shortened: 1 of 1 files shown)


sarahg:
Thanks for the PR, @heddn! What is the use-case for loading Redis module functions like this instead of via Drupal (by enabling the module)?

heddn:
@sarahg if you wire everything up, and push your code, even if you have redis enabled in core.extensions.yml, your site will die. You have to first push/enable redis. Then in a later push you can push the config. This let's you run site install (un-tested but in theory possible) with redis enabled. Even before any modules are wire up.

heddn:
BTW, these examples are copied from the README.txt in memcache from https://www.drupal.org/project/memcache/issues/2934916

alexfornuto:
Assigning @sarahg for technical review.

greg-1-anderson:
This is useful, causes no real harm (at the momen), and other people are doing it. 

I still have some reservations insofar as the custom autoloader isn't necessary after the module is enabled. I suppose we could conditionalize adding the autoloader based on whether the module is enabled, but that would add even more complexity to the example. I guess the rub here is that at this point we haven't gotten far enough in the bootstrap to know that the module is enabled or not. If we could do that, I'd rather go the other way and only add `$settings['container_yamls'][]` when the module is already enabled, rather than monkey-patch to make the services.yml work even when the module is not enabled. This works today, but I'm not sure that this is guaranteed to work e.g. in future versions of the redis module.

Maybe we should wait until this documentation is in the redis module too before putting it in the Pantheon documentation.

See: https://cgit.drupalcode.org/redis/tree/README.md

Since the advice here is not recommended in the Redis README, I think I am slightly -1 on this PR.

sarahg:
Agreed with Greg. This is definitely useful, but I suggest we stick with the recommendations from the module README.

alexfornuto:
I'm closing this issue based on the input above, but thank you for contributing nonetheless! Please feel free to submit new PRs/Issues, or continue the discussion here if you'd like.

PR #4320, opened 2018-12-13
add note about scope of support

### LINKED ISSUE #4316: Point users of the Build Tools process to Power Users for support questions
Re: https://pantheon.io/docs/guides/build-tools/

Priority: Low

## Issue Description

Support receives questions from customers on a regular basis asking for help with the Build Tools process. However, because the vast majority of these questions have to do with some interplay between Pantheon and 3rd party applications (GitHub, CircleCI, or both) and these 3rd party applications are outside our Scope of Support, it is left to CSEs in chat and tickets to explain how this is outside of scope, and then suggest posting the question/s to the Power User's Slack channel or email group.

While most seem OK with that - especially after describing who the Power Users are (experienced users who use the tools all the time, some Pantheon engineers, etc.) - it would still be better to properly set support expectations prior to their opening a chat or ticket.

It would also (likely) prevent a small number of low csat ratings we occasionally get when the customer feels we're pushing them off to some other group, or those who simply don't understand why this is not in Scope / a valid issue for Support to handle.

## Suggested Resolution

Add a clear note setting support expectations for anyone using the Build Tools process where anyone who uses it would see it, suggesting the Power Users as a great source for answers and describing why contacting Support - unless it is a Pantheon-only issue - is not likely be be helpful.

If it is possible to do, I'd add something to each of the pages of the Build Tools guide (make it big and obvious on whatever page is viewed most often, and smaller / lower in the page on the others) as well as the Scope of Support page.

This could be made more general if it does not refer to Build Tools explicitly, but instead describes how documented workflow examples & tools (like Build Tools) that rely on 3rd party apps cannot be supported via Chat or Ticket.

alexfornuto:
Closes #4316 

## Effect
- Adds a note about the scope of Pantheon Support, and link to Power Users

## Remaining Work
- [x] Review from @xq42 or others in @pantheon-systems/cse 
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/guides/build-tools/01-introduction.md
+++ b/source/_docs/guides/build-tools/01-introduction.md
@@ -142,14 +142,13 @@ Composer is used to fetch dependencies declared by the project as part of a Circ
 
 9. [Authorize CircleCI on GitHub](https://github.com/login/oauth/authorize?client_id=78a2ba87f071c28e65bb){.external}.
 
-    <div class="alert alert-info" markdown="1">
-    #### Note {.info}
     If you are redirected to the CircleCI homepage, you have already authorized the service for your GitHub account. Nice! Way to be ahead of the game.
-    </div>
 
 <div class="alert alert-info" markdown="1">
 #### Note {.info}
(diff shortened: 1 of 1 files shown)


alexfornuto:
>Will that be available in any other docs, or just in this one?

I think saying it once in the beginning is enough. This isn't the sort of guide that we expect people to jump into partway through.

PR #4321, opened 2018-12-13
Note the Terminus ACME plugin

### LINKED ISSUE #4315: Add information about pre-provisioning ssl using dns TXT records
Re: https://pantheon.io/docs/https/

Priority: Medium

## Issue Description

Pre-provisioning ssl certs for domains already using them on another host and being moved to Pantheon is currently possible using the default acme-challenge .txt file downloaded from the dashboard, or with a CSE-generated dns TXT record.

However the dns TXT record is not mentioned anywhere (that I could find) in our docs. 

This was brought up by El more than a year ago here:

https://github.com/pantheon-systems/documentation/issues/2945

but Ari mentioned it was not yet ready. It is now...

The number of sites that have troubles with the default .txt file method seems to be relatively low, but for those who cannot use it, their only documented alternative is to simply change the dns records to point to Pantheon and wait for however long it takes dns to propagate and Let's Encrypt to provision the cert.

A platinum support customer requested information on how to request dns TXT records be generated get added to the docs, as he's had to launch 2 sites recently and then wait for ssl since the default .txt file method wouldn't work. He's familiar with our docs and would have found the instructions to contact Support for help if he'd have known it was an option.

Ticket: `121559`

## Suggested Resolution

Add information on how to request dns TXT records from Support.

Ideally phrased in such a way to discourage the request until they've tried the default .txt challenge file, but since we've been doing this for customers for months now, it seems it would be good for it to be available for customers to request as well.

alexfornuto:
Closes #4315 

## Effect
- Adds a note to the launch guide linking to the ACME challenge plugin for Terminus

## Remaining Work
- [ ] Review from @xq42 (optional)
- [x] Copy review

## Post Launch
- [x] Update Status Report

### DIFF
--- a/source/_docs/guides/launch/03-domains.md
+++ b/source/_docs/guides/launch/03-domains.md
@@ -54,6 +54,8 @@ To pre-provision HTTPS, CAA records must either:
     The validation file to pre-provision HTTPS must be accessible over HTTP, not just HTTPS. A redirect from HTTP to HTTPS will work, but if a request over HTTP returns a 404, for example, the validation will fail.
 
     We recommend leaving the challenge file in place until the migration is complete. That way, if DNS is not updated within the 7 day window provided by Let's Encrypt, the certificate can be re-issued.
+
+    If you're unable to host the challenge file, consider using the [Terminus ACME Plugin](https://github.com/pantheon-systems/terminus-acme-plugin){.external} to generate DNS TXT records to validate domain ownership.
     </div>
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
@xq42 Customers can run this themselves, and this plugin may soon be integrated into Terminus, at which point we'll update this note. That's also why it's not listed in the plugin directory. @greg-1-anderson do you think we should include it there until it's merged into Terminus?

PR #4322, opened 2018-12-13
add clarity to the filemount path option

### LINKED ISSUE #3771: Pantheon YAML Configuration Files Doc Update
Re: [Pantheon YAML Configuration Files](https://pantheon.io/docs/pantheon-yml/)
https://pantheon.io/docs/pantheon-yml/#filemount-path

Priority (Low‚ Medium‚ High): Medium?

## Issue Description:  I'd like maybe get some advanced use cases for altering this file because it seems that most users should not edit this location.  
1. When should this setting be used?
2. Does this automatically create a symlink?
3. What kinds of errors could happen if the location is set improperly and how can CSE check to verify if the setting made in the yml file is problematic?

alexfornuto:
Closes #3771 

## Effect
- Notes what the filemount option does (create a symlink)
- Notes as valid only for custom upstreams
- Links to Non-Standard Files Locations

## Remaining Work
- [x] Technical review (@ttrowell and/or @greg-1-anderson)
- [x] Copy Review

## Post Launch
- [x] Update Status Report

### DIFF
--- a/source/_docs/pantheon-yml.md
+++ b/source/_docs/pantheon-yml.md
@@ -77,7 +77,7 @@ Pantheon provides a [cloud-based filesystem](/docs/files/) to store user-generat
 filemount: /files
 ```
 
-This creates a new symlink to the filesystem at the specified location. Note that this setting is only useful when creating a custom upstream that requires it. Otherwise, consider using a symlink as described in [Non-Standard Files Locations](/docs/non-standard-file-paths/).
+This creates a new symlink to the filesystem at the specified location. Note that this setting is only recommended to use when creating a custom upstream. Instead, consider using a symlink as described in [Non-Standard Files Locations](/docs/non-standard-file-paths/).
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Hm, we are getting into the area of philosophy a bit here.

> This creates a new symlink to the filesystem at the specified location. 

This is correct. +1

> Note that this setting is only useful when creating a custom upstream 

This feature can also be useful for individual sites; however, there are complications with that. Therefore, I'd say **recommended** instead of **useful** here.

> that requires it.

I don't think that this feature is ever "required". It is an optimization that allows the removal of the `sites/default` directory if it is not needed. I think I'd just delete this bit.

> Otherwise, consider using a symlink as described in [Non-Standard Files Locations](/docs/non-standard-file-paths/).

Maybe say "Instead," instead of "Otherwise,".  Optional.

PR #4325, opened 2018-12-14
Fix SimpleSAML Composer heading

sarahg:
## Effect
- Fix the Composer heading so these aren't nested under the WP directions. This section is not WordPress-specific.

## Remaining Work
none

## Post Launch
- [ ] Update Status Report


alexfornuto:
Thanks @sarahg!

PR #4334, opened 2018-12-18
explain why CNAME records are preferred

### LINKED ISSUE #3798: Why does Pantheon recommend using CNAME records?
Priority:

## Title
Why does Pantheon recommend using CNAME records?

## Description
Advantage of using CNAME over A record and why we recommend to use CNAME.

## Expected Audience
Site Owner/Developer

alexfornuto:
Closes #3798 

## Effect
- Creates a partial file explaining why CNAME records are preferred.
- Adds the partial to several docs covering DNS records.

## Remaining Work
- [ ] Technical Review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/dns.md
+++ b/source/_docs/dns.md
@@ -173,6 +173,8 @@ Address: 2620:12a:8000::1
 
 In the example above, we can create an `A` record for `www` set to `23.185.0.1` to substitute the `CNAME` record.
 
+{% include("content/cname-advantage.html")%}
+
 ## See Also
 
  - [Launch Essentials](/docs/guides/launch/)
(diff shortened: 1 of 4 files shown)


greg-1-anderson:
This is a decent explanation; however, I wonder if we should even continue recommending CNAMEs at this point.

- Most customers want their bare domain in addition to the www subdomain to route to their site. This means that they will be using the A record for at least one entry in any event.
- With our global CDN, all customers (except for legacy customers not yet on GCDN) are using one of just four IP addresses.
- We are contemplating platform changes that would give us even more control about shifting customer sites around to different regions in our platform without requiring an IP address change.

Although the future isn't here yet, the fact that A records are so widely used already more-or-less negates our ability to do anything practical with the CNAMEs over a wide portion of our sites, and I'm not convinced that the utility of the CNAME is really there under the circumstances. I therefore think this explanation isn't fully accurate, and furthermore, it is not needed, as the user interface already tells the user to use an A record or a CNAME record as needed.

alexfornuto:
Based on @greg-1-anderson's feedback and internal discussion, I'mma just close this out.

PR #4335, opened 2018-12-18
Removing unused settings.php step in Drupal 8 command line guide

stevector:
This pull request updates a guide page that walks through the set up of a new Drupal 8 site using the command line. https://pantheon.io/docs/guides/drupal8-commandline/

As of a recent Drupal 8 release (8.5.0 or 8.6.0) the Drupal profile chosen during installation is no longer written to the settings.php file. That means that file is not changed and does not need to be committed when making a new site.

The now-unnecessary steps are removed.
 
## Remaining Work
- [x] Peer review (I think @davidneedham,  @greg-1-anderson, and @ari-gold have been interested in this detail)

## Post Launch
- [ ] Update Status Report
- [ ] Archive from Done in Waffle

### DIFF
--- a/source/_docs/guides/drupal8-commandline.md
+++ b/source/_docs/guides/drupal8-commandline.md
@@ -67,35 +67,15 @@ If you see your Pantheon sites, then it was installed and authenticated successf
 
         terminus drush steve-site-d8.dev -- site-install -y
 
-
     Use the password included in the output of that command to sign into the site with your browser, or use this command to get a one-time login link:
 
         terminus drush  steve-site-d8.dev  -- user-login
 
-    Installing Drupal results in a one-line change to `settings.php`, which we can then review and commit. As a reminder, Drush is the command line utility for Drupal itself.	Terminus is simply passing through the Drush commands to the site on Pantheon. To get a full list of Drush commands run:
-
-        terminus drush steve-site-d8.dev -- help
-
-    The `--` signifies the end of the Terminus options, anything after `--` gets passed straight to Drush.
-
-4. Review the file changes:
-
-  ```
-  terminus env:diffstat steve-site-d8.dev
-  ```
-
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Cool. Maybe we can remove the "can't install Drupal 8 in git mode" check now too.

davidneedham:
Thanks for being proactive about this Steve, but can you confirm that it's actually something that has been updated with Drupal core? As I mentioned in my update, this doesn't happen every time. And there were other oddities that lead me to believe this was an error. The first time this happened I thought it was a change to core, but I think it turned out to be an error.

I'll see if I can find my notes or any changelog from core to confirm a change.

greg-1-anderson:
I haven't looked at this recently, but the old patch (before it was merged) only skipped writing the installation profile if there was a single distribution in the codebase. Thought maybe they improved it. Should confirm for sure.

greg-1-anderson:
I'll give it a spin and a look when I have a chance

stevector:
> Still, I'd feel more confident if we had something to point to at d.o confirming this change.

Same. I'll keep looking.

I did confirm the behavior by running this string of commands 15 times with different integers.

```
INT=2 && terminus site:create settings-check-$INT "My Site D8" "Drupal 8"  && terminus drush settings-check-$INT.dev -- site-install -y && terminus env:diffstat settings-check-$INT.dev
```

alexfornuto:
@stevector @greg-1-anderson @davidneedham just checking in, is this still under "technical review" from one/some/all of you?

greg-1-anderson:
@stevector I was thinking of a different issue. I think the one that you found is the correct fix. I am satisfied that we can publish this documentation update.

alexfornuto:
Thanks to everyone who contributed to this!

PR #4342, opened 2018-12-20
Correcting greater than/less than signs.

sparklingrobots:
I believe the greater than/less than signs were used in the wrong order (e.g. it read 'greater than or equal to 100MB', 'greater than 100mb-256mb' and 'less than 256mb' in the chart. I swapped the direction of the 1st and 3rd, and removed the middle one since it's really between them. I'd normally just merge but want to make sure I'm not doing this incorrectly.

Closes n/a

## Effect
- Swaps the greater than/less than signs
- Removes one greater than sign 

## Remaining Work
- [ ] none

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/platform-considerations.md
+++ b/source/_docs/platform-considerations.md
@@ -112,9 +112,9 @@ Due to the configuration of the [Pantheon Filesystem](/docs/files/), Pantheon's 
 
 | File Size       | Platform Compatibility            | Notes                               |
 |:--------------- | --------------------------------- |------------------------------------ |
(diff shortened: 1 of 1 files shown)


alexfornuto:
@sparklingrobots I believe the usage is correct. What makes it less than or greater than is what side the number is on. At least, that's what my 4th grade math teacher taught me, that the open end of the `>` should point to the larger number, and the point "points" to the smaller number.

alexfornuto:
@sparklingrobots nevermind, a fresh pot of coffee has ended last night's stupification. Thanks for the correction!

PR #4371, opened 2019-01-03
Remove the "Analyze the PHP FPM Error Log" section.

sarahg:
## Effect

Removes the "Analyze the PHP FPM Error Log" section. I think the example given is incorrect on a few counts: 

- It's normal behavior to see a php child process exit — we limit the number of requests a process can handle using the `pm.max_requests` directive in `php-fpm.conf`. After a process fields that many requests, it's re-spawned and shown as exiting in the php-fpm log (and you'll see a new one start right after). It's not indicative of a problem.

- The example shows a PHP process that exits 11 hours after the slow trace it was associated with. The process exit and the slowness are not related in a meaningful way. Matching worker IDs like this is only potentially useful information if the timestamps match up and if there's an actual error (not a routine exit).

## Remaining Work
- [x] technical review
- [ ] maybe a follow-up issue to replace this section with a different explanation of how to use the php-fpm log

## Post Launch
- [ ] Update Status Report


alexfornuto:
I'm not sure who to ask for a technical review of this, maybe @greg-1-anderson?

greg-1-anderson:
So, I am ambivalent about this. The example isn't really claiming that php processes exiting is a problem, it is merely attempting to point out a technique that might be used to correlate items in the php error log with items in the php slow log.

As observed, though, in this example, the entry in the php error log is reported to have run for about 5 minutes, and the docs then line it up with some process that ran 11 hours ago. This is erroneous.

Since I am uncertain how to make the example better in a way that would be meaningful and productive for our users, though, I guess I am 👍 on just removing it as suggested.

alexfornuto:
Thanks for your input @greg-1-anderson, and thanks @sarahg for PR!

PR #4383, opened 2019-01-10
Update sculpin_site.yml

Javi-er:
Added contributor.

Closes #

## Effect
PR includes the following changes:
-
-
-

## Remaining Work
- [ ] List any outstanding todos
- [ ] If needed

## Post Launch
To be completed by the docs team upon merge: 
- [ ] Redirect `/docs/old-path/` => `/docs/new-path/` (if applicable)
- [ ] Include/exclude pages ^ respectively within docs search service provider (if applicable)
- [ ] Update Status Report
- [ ] Archive from **Done** in Waffle

PR #4382, opened 2019-01-10
Create guide for implementing SimpleSAMLPHP on D8

Javi-er:
Step by step guide about how to implement SimpleSAMLphp as an indentity provider on a Drupal 8 site hosted at Pantheon.

Closes #

## Effect
PR includes the following changes:
Wrote a guide for implementing SimpleSAMLphp on D8 as an Idp.

## Post Launch
To be completed by the docs team upon merge: 
- [ ] Redirect `/docs/old-path/` => `/docs/new-path/` (if applicable)
- [ ] Include/exclude pages ^ respectively within docs search service provider (if applicable)
- [ ] Update Status Report
- [ ] Archive from **Done** in Waffle

PR #4384, opened 2019-01-10
Simplesamlidp

alexfornuto:
Continues #4382 
Continues #4383 

## Effect
- New guide for implementing SimpleSAMLphp on D8 as an Idp.
- Continued from the external fork here for CI testing and collaboration.

## Remaining Work
- [ ] Technical Review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/app/config/sculpin_site.yml
+++ b/app/config/sculpin_site.yml
@@ -335,7 +335,7 @@ contributors:
     github: //github.com/edwardangert
     avatar: //avatars2.githubusercontent.com/u/17991901
     linkedin: //www.linkedin.com/in/edward-angert/
-   Javi-er:
+  Javi-er:
     name: Javier Reartes
     bio: Drupal developer at Taoti
     github: //github.com/Javi-er
--- a/scripts/deploy-multidev.sh
+++ b/scripts/deploy-multidev.sh
@@ -90,7 +90,7 @@ if [ "$CIRCLE_BRANCH_SLUG" != "master" ] && [ "$CIRCLE_BRANCH_SLUG" != "dev" ] &
   touch ./multidev-log.txt
   while true
   do
-    if ! rsync -vvvv --checksum --delete-after -rtlz --ipv4 --log-file=multidev-log.txt -e 'ssh -p 2222 -oStrictHostKeyChecking=no' output_prod/docs/ --temp-dir=../../tmp/ $normalize_branch.$STATIC_DOCS_UUID@appserver.$normalize_branch.$STATIC_DOCS_UUID.drush.in:files/docs/; then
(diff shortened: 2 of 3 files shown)


sparklingrobots:
I haven't yet tested the process shown in this doc, but wanted to share a few initial thoughts after a read-through:  

1. We don't use apache, so that section will need to be rewritten.  I'm not sure at this point if it can just be removed, though I suspect it can.
2. The contrib module in question is still an alpha release, is marked as minimally maintained and hasn't been touched since March 2018.  The doc references a fork of the contrib module, presumably because the maintainer isn't moving forward on the patch submitted by Javier. 

Given the semi-abandoned nature of the module & the relatively few (if any) requests for IdP set-up documentation, I wanted to get further input from @alexfornuto before proceeding.  Generally speaking, it seems like a helpful guide and I'm happy to do further technical validation and testing if this is something we'd like to include in our docs.

Javi-er:
Thanks for looking at this, the Apache section is actually for setting up SimpleSAMLphp on the developer machine, it's not needed for setting up on Pantheon of course. That section could be removed if you prefer it and instead adding a note linking to SimpleSAMLphp documentation section where it shows that.

Regarding the saml_idp module, the module maintainer never replied to to the patch we submitted, the module doesn't seems to be maintained at all at the moment. 
If the module owner doesn't reply back in some weeks, we'll submit a request to become maintainers and then merge this patch, after this I can update this guide removing that comment since it will not longer be relevant.

Reviewing the guide here [/docs/guides/drupal-8-simplesamlphp/](https://simplesamli-static-docs.pantheonsite.io/docs/guides/drupal-8-simplesamlphp/) I already noticed some text formatting to fix and other small things, but I'll wait on making edits on this until you decide if you want to move forward with it.

alexfornuto:
@Javi-er thanks for the updates, and thanks @sparklingrobots for looking at this. I agree that we should:

1. Wait to see if the patch is merged or stewardship of the plugin is transferred
2. If the instructions on setting up SimplaSAMLphp on the developer machine with Apache are pulled directly from existing documentation on their site, then we should link to it instead of duplicating information. That way when the information is updated, our content doesn't become state.
3. @Javi-er let me worry about formatting fixes. If you made additional commits on your fork I can merge them into this branch, but we run the risk of merge conflicts. I *hate* merge conflicts.

alexfornuto:
I'm going to merge this PR with @Javi-er's original one, now that I'm more comfortable with the concept of pushing to branches on remote forks.

alexfornuto:
Closing in favor of #4382 - let's follow-up there with conversation on the status of the plugin patch.

PR #4392, opened 2019-01-14
downgrading caveat to custom certificates

EdwardAngert:
## Effect
- adds information relevant to sites using custom certs regarding switching to let's encrypt before attempting to downgrade

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/custom-certificates.md
+++ b/source/_docs/custom-certificates.md
@@ -142,6 +142,10 @@ If a Let's Encrypt certificate was deployed to the Global CDN before adding CAA 
 ### CA limitations
 Your CA must accept the CSR Pantheon provides. If your CA fails to accept our CSR, you will not be able to use it to generate a certificate. The CA Globalsign does not currently meet this requirement. The workaround is to simply use another CA.
 
+### Downgrading a Site that uses a Custom Certificate
+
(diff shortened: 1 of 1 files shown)


source/_docs/custom-certificates.md
```diff
@@ -142,6 +142,10 @@ If a Let's Encrypt certificate was deployed to the Global CDN before adding CAA
 ### CA limitations
 Your CA must accept the CSR Pantheon provides. If your CA fails to accept our CSR, you will not be able to use it to generate a certificate. The CA Globalsign does not currently meet this requirement. The workaround is to simply use another CA.
 
+### Downgrading a Site that uses a Custom Certificate
```
ari-gold: @EdwardAngert  @alexfornuto  Just looking at this after reading the weekly docs report. What is the scenario we're capturing here? Downgrading to a free site? Or switching from custom cert to automated HTTPS with Let's Encrypt?

EdwardAngert: @ari-gold, we added this after a customer had trouble downgrading their site (with a custom cert) to a plan that doesn't feature custom certs. We realized it was because the Platform requires a cert in place and dropped in this paragraph on the Custom Cert doc as it's potentially the only scenario someone would encounter it.

anything I should add to it for clarity?

PR #4420, opened 2019-01-28
update SSO documentation

tmountjr:
Ran into a situation with a client where the `user_id` attribute was not being set in their SAML request and had to be passed manually.

Closes (n/a)

## Effect
- Adds an optional consideration when configuring the SP

## Remaining Work
n/a

## Post Launch
- [ ] Update Status Report


alexfornuto:
Looks good, thanks @tmountjr!

PR #4440, opened 2019-02-01
Adding a description of the migration process - first draft

obicke:
Closes #

## Effect
- Add details about the migration process. This version is more fitting for a documentation page, than the information provided about this service on the marketing site. 

## Remaining Work
- [ ] Get reviews from @lauren-kelly and @bentekwork 
- [ ] Additional wordsmithing would probably be A Good Thing.
- [ ] Add links
- [ ] Review/improve formatting

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/professional-services.md
+++ b/source/_docs/professional-services.md
@@ -18,7 +18,20 @@ If you need advice on architecting a solution, planning for the future, preparin
 
 
 ## Migrations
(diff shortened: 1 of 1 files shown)


obicke:
Based on conversations, we'll want to provide more re-assurance around the fact that PS will expressly want to avoid competing with Pantheon partners, and in fact that the goal is to provide complimentary services. There may also be a little Too Much Information there. I'll review shortly.

obicke:
After further internal discussion, we're going to take a slightly different approach. Closing this PR.

PR #4488, opened 2019-02-13
Add password reset options

### LINKED ISSUE #4432: Resetting Passwords Doc Update
Re: [Resetting Passwords](https://pantheon.io/docs/resetting-passwords/)

Priority: Low

## Issue Description:
We provide Drush/WP-CLI commands for resetting Drupal/WP account passwords. These are the best option when users can use Terminus, but Windows users are often unable to run Terminus (and this is [unsupported](https://pantheon.io/docs/platform-considerations/#terminus-support)). Support cannot reset Drupal/WP passwords.

## Suggested Resolution
Provide the MySQL equivalent of the Drush/WP-CLI commands as an alternative to using Terminus.

sarahg:
Closes #4432

## Effect
- Adds headings to clarify the 3 password reset options for each CMS
- Adds a new PW reset method, doing this directly in the database. This is not the easiest option, but it may be necessary for users who can't use Terminus (e.g Windows users).

## Remaining Work
- [x] add the MySQL password reset options
- [ ] maybe split Drupal/WP into separate tabs?

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/resetting-passwords.md
+++ b/source/_docs/resetting-passwords.md
@@ -39,31 +39,8 @@ Or you can reset any user's password from the command line by running the [`user
 $ terminus drush <site>.<env> -- user-password user_name --password='Astr0nGP455w0rD'
 ```
 
-#### Option C: Reset your password directly in the database (advanced)
-Since Drupal does not store passwords in plain text, you need to generate a hashed version of the password before updating it directly in your database. Drupal core provides a script to do this:
-
-* Drupal 7: `scripts/password-hash.sh`
-* Drupal 8: `core/scripts/password-hash.sh`
-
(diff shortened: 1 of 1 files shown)


sarahg:
@alexfornuto This one is ready for a copy review. I reverted the step-by-step directions and am linking off to Drupal/WP docs instead. Now that we have 3 short sections for each CMS, maybe we want to format this in tabs?

alexfornuto:
Thanks @sarahg! I think this doc is short enough that we can do without the tabs, in the interest of making it easy to link directly to any section. For that same reason, I'm removing the second H2 and decreasing each header level after, for easier TOC / Linking.

PR #4516, opened 2019-02-22
update git mv grep filtering

### LINKED ISSUE #4494: Serving Sites from the Web Subdirectory Doc Update (Invalid Order)
Re: [Serving Sites from the Web Subdirectory](https://pantheon.io/docs/nested-docroot/)

Priority: High

## Issue Description:

The command mentioned in the documentation to move WordPress files is incorrect:
`git mv -k $(find . -type f -maxdepth 1  | grep -v pantheon.yml) wp-includes wp-content wp-admin ./*.php web`

It throws the following error in Ubuntu 18.10:

`find: warning: you have specified the -maxdepth option after a non-option argument -type, but options are not positional (-maxdepth affects tests specified before it as well as those specified after it).  Please specify options before other arguments.`

## Suggested Resolution

Replace `find . -type f -maxdepth 1` with `find . -maxdepth 1 -type f`. The `maxdepth` should come before the `type`.

### LINKED ISSUE #4495: Serving Sites from the Web Subdirectory Doc Update (Too Greedy)
Re: [Serving Sites from the Web Subdirectory](https://pantheon.io/docs/nested-docroot/)

Priority: High

## Issue Description:

The command suggested to move WordPress files is too greedy:
`git mv -k $(find . -type f -maxdepth 1  | grep -v pantheon.yml) wp-includes wp-content wp-admin ./*.php web`

It will move not only file, required by WordPress, but it will move everything in the root directory, which is almost always is a bad idea. Because it will move `pantheon.upstream.yml`, `.gitignore`, `README.md` and others.

## Suggested Resolution

The much better idea would be to copy only required files, using a less greedy approach: `git mv -k ./wp-includes ./wp-content ./wp-admin ./*.php ./license.txt ./readme.html ./web`.

alexfornuto:
Closes #4495 
Closes #4494 

## Effect
- Fixes the invalid order of `find` flags.
- Adds to our `grep` exclusionary filter additional files we know we don't want to move into a nested docroot.
- Replaces `grep` with `egrep` for fun and profit.

## Remaining Work
- [x] Technical Review
- [x] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/nested-docroot.md
+++ b/source/_docs/nested-docroot.md
@@ -82,7 +82,7 @@ You'll need to move the CMS code into the `web` subdirectory, either manually or
 
   ```bash
   mkdir web
-  git mv -k $(find . -type f -maxdepth 1 | grep -v pantheon.yml) includes/ misc/ modules/ profiles/ scripts/ sites/ themes/ index.php web
+  git mv -k $(find . -maxdepth 1 -type f | egrep -v 'pantheon.yml|.gitignore|README.md|pantheon-upstream.yml') includes/ misc/ modules/ profiles/ scripts/ sites/ themes/ index.php web
   ```
   These commands create the `web` subdirectory, then use Git to move required files into the nested docroot.
 
@@ -111,7 +111,7 @@ You'll need to move the CMS code into the `web` subdirectory, either manually or
 
   ```bash
   mkdir web
-  git mv -k $(find . -type f -maxdepth 1 | grep -v pantheon.yml) core drush modules profiles sites themes vendor index.php web
+  git mv -k $(find . -maxdepth 1 -type f | egrep -v 'pantheon.yml|.gitignore|README.md|pantheon-upstream.yml') core drush modules profiles sites themes vendor index.php web
   ```
   These commands create the `web` subdirectory, then use Git to move required files into the nested docroot.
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
@slavafomin feel free to review.

slavafomin:
Hello! Thank you for the PR, however, I still believe, that it's a bad idea to use exclusion filter, because you would never know what files end developers are storing in the project's root. I can easily give you another examples of files, which could be stored there, e.g.: `.editorconfig`, `.idea`, `docker-compose.yml`.

As I stated in #4495 the only viable strategy here is to use inclusion filter and specify exact files used by WordPress. This rarely changes and the list of files is fixed, so it could be easily maintained over time.

alexfornuto:
@davidstrauss request for comment, since you suggest being more exclusionary instead of switching to inclusionary.

davidstrauss:
The intent for documentation on moving to a nested docroot should be (1) to preserve the functionality of the site and (2) not introduce security issues. Since we're starting with content that was already in the docroot, I think the latter is a low risk.

So, we need to look at how we can preserve functionality. A whitelist of files and paths to move to the nested docroot will introduce problems for sites with content that needs to be in the docroot but won't be on our whitelist. I've seen many sites drop things like PDFs and images right into their existing docroot, even through it's not a best practice.

For some of the file types mentioned by @slavafomin, it's unclear what the right choice is. For example, a `.editorconfig` file might be more appropriate in the docroot if it concerns the CMS and uses paths that expect the CMS directories to be immediate children. It would potentially break the configuration to hold it outside the nested docroot while moving the other files into one.

So, I would support expanding the blacklist of patterns to exclude from nested docroot migration, but I'd be reluctant to adopt a whitelist approach for the reasons above. If we haven't already, we should also encourage developers to inspect the split of files and potentially make the modifications of their choice.

alexfornuto:
Thanks for the additional input! I thinking the best way to move forward is to add a note describing how to add items to the exclusion list, and using the `-n` flag for `git mv` to perform dry runs first.

PR #4238, opened 2018-11-09
Update continuous-integration.md

ccjjmartin:
SimpleTests are not supported on Pantheon.  I have an open ticket with Pantheon support discussing this: https://dashboard.pantheon.io/sites/025e433e-4745-4e86-bd6a-9fb252581f65#support/ticket/119011

Closes #

## Effect
PR includes the following changes:
-
-
-

## Remaining Work
- [ ] List any outstanding todos
- [ ] If needed

## Post Launch
To be completed by the docs team upon merge: 
- [ ] Redirect `/docs/old-path/` => `/docs/new-path/` (if applicable)
- [ ] Include/exclude pages ^ respectively within docs search service provider (if applicable)
- [ ] Update Status Report
- [ ] Archive from **Done** in Waffle

PR #4524, opened 2019-02-25
SimpleTests

alexfornuto:
Continues #4238 

## Effect
- Adds steps to run SimpleTests on Pantheon.

## Remaining Work
- [x] Technical review from @ccjjmartin and/or @sarahg, since I made such heavy edits.
- [x] Confirm & Update which versions of Drupal these instructions apply to.
- [x] Copy Review

## Post Launch
- [ ] Update Status Report


source/_docs/continuous-integration.md
```diff
@@ -17,18 +17,69 @@ Continuous Integration (CI) is a method of running automated unit and integratio
...
+          command: |
+            if [ "${CIRCLE_BRANCH}" != "master" ]; then
+
+              terminus drush $SITE_NAME.$ENV_NAME -- en site_test -y
+
+              # If you don't clear the cache immediately before running tests
+              # we get the html gibberish instead of a passing test.
+              terminus drush $SITE_NAME.$ENV_NAME -- cc all
+
+              # NOTE: It is a requirement to use Terminus <=1.9
```
ccjjmartin: @alexfornuto This is my bad.  I should not have listed a specific version of terminus here as the new version required is 2.0.  This should read it is a requirement to use the latest version of terminus to exclude notices about needing to upgrade.  (there may be a terminus option to exclude notices I am not aware of we could put in the command but for now documentation of using the latest version will cover the error)

source/_docs/continuous-integration.md
```diff
@@ -17,18 +17,69 @@ Continuous Integration (CI) is a method of running automated unit and integratio
...
+              # in order to exclude the notice in shell output of the
+              # embedded command to find the absolute path.
+              terminus drush $SITE_NAME.$ENV_NAME -- exec php `terminus drush $SITE_NAME.$ENV_NAME -- eval "echo DRUPAL_ROOT" 2>/dev/null`/scripts/run-tests.sh --url http://$ENV_NAME-$SITE_NAME.pantheonsite.io ExampleTestGroupName
```
ccjjmartin: @alexfornuto To be consistent lest change `ExampleTestGroupName` to `OptionalTestGroup`.  Maybe they should both read `OptionalTestGroupName`?

source/_docs/continuous-integration.md
alexfornuto: @ccjjmartin do the warnings break the test?

ccjjmartin: @alexfornuto Yes, if the warnings are displayed they will break the test since we have to build a path to the test files using the output of `echo DRUPAL_ROOT` the `[WARNING] Upgrade Terminus` string gets appended to our file path and that is how it breaks.

alexfornuto:
@ccjjmartin I believe I've made all the required edits based on your feedback, with one remaining question about Terminus version warnings.

source/_docs/continuous-integration.md
alexfornuto: Resolved by https://github.com/pantheon-systems/documentation/pull/4524/commits/68085c0872b0a730060406607284f7a18f0e8bf8

source/_docs/continuous-integration.md
```diff
@@ -17,18 +17,69 @@ Continuous Integration (CI) is a method of running automated unit and integratio
...
+
+To run tests on Pantheon:
+
+1. Enable `site_test`:
+
+   ```bash
+   terminus drush $SITE_NAME.$ENV_NAME -- en site_test -y
+   ```
+
+1. Clear the cache immediately before running test to avoid strange failures:
```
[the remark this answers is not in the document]
alexfornuto: Resolved by https://github.com/pantheon-systems/documentation/pull/4524/commits/68085c0872b0a730060406607284f7a18f0e8bf8

source/_docs/continuous-integration.md
```diff
@@ -17,18 +17,69 @@ Continuous Integration (CI) is a method of running automated unit and integratio
...
+#### Note {.info}
+The drush test-run command was dropped in Drush 7 and 8. See [this GitHub issue](https://github.com/drush-ops/drush/issues/1362){.external} for more details.
+</div>
+
+To run tests on Pantheon:
+
+1. Enable `site_test`:
+
+   ```bash
+   terminus drush $SITE_NAME.$ENV_NAME -- en site_test -y
```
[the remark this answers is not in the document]
alexfornuto: Resolved by https://github.com/pantheon-systems/documentation/pull/4524/commits/68085c0872b0a730060406607284f7a18f0e8bf8

source/_docs/continuous-integration.md
alexfornuto: Resolved by https://github.com/pantheon-systems/documentation/pull/4524/commits/d493ff6235c638b187eb11b0c180418b39f8ad1e

alexfornuto:
@ccjjmartin can you confirm, should this be updated to reflect that it's only for D7, and/or tested for D8 compatibility?

source/_docs/continuous-integration.md
```diff
@@ -17,18 +17,73 @@ Continuous Integration (CI) is a method of running automated unit and integratio
...
+   ```bash
+   terminus drush $SITE_NAME.$ENV_NAME -- eval "echo DRUPAL_ROOT"
+   ```
+
+  In the command above, you may need to strip out warnings by ending the command with `2>/dev/null`.
+
+The full command will look something like this:
+
+```bash
+terminus drush $SITE_NAME.$ENV_NAME -- exec php `terminus drush $SITE_NAME.$ENV_NAME -- eval "echo DRUPAL_ROOT" 2>/dev/null`/scripts/run-tests.sh --url http://$ENV_NAME-$SITE_NAME.pantheonsite.io OptionalTestGroup
```
ccjjmartin:
```suggestion
export TERMINUS_HIDE_UPDATE_MESSAGE=1
terminus drush $SITE_NAME.$ENV_NAME -- exec php `terminus drush $SITE_NAME.$ENV_NAME -- eval "echo DRUPAL_ROOT" 2>/dev/null`/scripts/run-tests.sh --url http://$ENV_NAME-$SITE_NAME.pantheonsite.io OptionalTestGroup
```

source/_docs/continuous-integration.md
```diff
@@ -17,18 +17,73 @@ Continuous Integration (CI) is a method of running automated unit and integratio
...
+
+A full CircleCI command might look similar to this:
+
+```
+      - run:
+          name: Test simpletest
+          command: |
+            if [ "${CIRCLE_BRANCH}" != "master" ]; then
+
+              terminus drush $SITE_NAME.$ENV_NAME -- en site_test -y
```
ccjjmartin:
```suggestion
              export TERMINUS_HIDE_UPDATE_MESSAGE=1

              terminus drush $SITE_NAME.$ENV_NAME -- en site_test -y
```

We could add the export line in these two locations and it would help make this more stable.

source/_docs/continuous-integration.md
```diff
@@ -17,18 +17,76 @@ Continuous Integration (CI) is a method of running automated unit and integratio
...
-
-[SimpleTest](https://drupal.org/project/simpletest) is a testing framework based on the [SimpleTest PHP library](https://github.com/simpletest/simpletest) that is included with Drupal core. If you are creating a custom web application, you should consider including SimpleTests of your module functionality.
-
-After enabling the SimpleTest module, you can use Drush to remotely execute SimpleTest on your Pantheon site. For example, if you want to execute the UserSaveTestCase test and generate XML output into a writeable directory, use the following command:
-
-    SITE_NAME=yoursitename
-    ENV=dev
-    drush @pantheon.$SITE_NAME.$ENV test-run -l https://$ENV-$SITE_NAME.pantheon.io/ UserSaveTestCase --xml='sites/default/files'
-
-The end result is written to https://dev-yoursitename.pantheon.io/sites/default/files/UserSaveTestCase.xml
```
alexfornuto: @ccjjmartin Is this line no longer true? Where are the results?

alexfornuto: P.S. I'm pretty sure this is the _last_ question I have about this PR.

ccjjmartin: @alexfornuto The way that we run the command, the results are output directly to the terminal (screen) when run.  I haven't tested the command with a file output option `--xml='sites/default/files' (but it is probably still there).  In my opinion it works better seeing the results in the case of an active developer or from CIs (which normally save the logs already for us).  So yeah, we should be good to remove that.

app/config/sculpin_site.yml
```diff
@@ -335,3 +335,7 @@ contributors:
     github: //github.com/edwardangert
     avatar: //avatars2.githubusercontent.com/u/17991901
     linkedin: //www.linkedin.com/in/edward-angert/
+  ccjjmartin:
```
alexfornuto: @ccjjmartin feel free to add a bio line here if you'd like, either in this PR or a new one if this is already merged when you see this.

alexfornuto: P.S. Or any of the other available fields, of course!

PR #4551, opened 2019-03-12
Ps documentation page update v2

obicke:
Closes #

## Effect
- Frames as "working with partners"
- Restructures to highlight some services over others
- Fleshes out a couple sections to provide details and inspire confidence in the services 
- Removes service-level/qualification indicators
- Some wordsmithing and standardizing.

## Remaining Work
- [ ] Get final review from @bentekwork 

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/professional-services.md
+++ b/source/_docs/professional-services.md
@@ -41,21 +41,21 @@ Your Engagement Manager monitors the process and works with you to ensure that u
 Implement additional services like extended WAF, Image Optimization, and Domain Masking using our advanced CDN implementation. [Contact us](https://pantheon.io/professional-services){.external} for details.
 
 ## Pantheon Enterprise Gateway (PEG)
-Provides a static outgoing IP address for use in IP whitelisting connections to external services. [Contact us](https://pantheon.io/professional-services){.external} for details.
+Provides a static outgoing IP address for use in IP whitelisting connections to external services. [Learn more about PEG](https://pantheon.io/docs/pantheon-enterprise-gateway){.external} or [contact us](https://pantheon.io/professional-services) for details.
 
 ## SSO/SAML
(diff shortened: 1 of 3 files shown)


source/_docs/professional-services.md
```diff
@@ -4,54 +4,63 @@ description: Pantheon Professional Services include consulting, migrations, load
 tags: []
 ---
 
-## Consulting and Specialty Services
-If you need advice on architecting a solution, planning for the future, preparing for expansion, tackling a particularly complex use case, or just need problem solving expertise, Pantheon's Professional Services (**PS**) team can help you achieve your goals. Our Professional Services team has the broad experience and deep expertise to guide you through your implementation needs. [Contact us](https://pantheon.io/contact-us){.external} for details.
+Pantheon Professional Services (**PS**) works closely with partners to provide customized solutions for the most complex technical challenges and ensure your success on Pantheon. [Contact us](https://pantheon.io/contact-us){.external} for details.
```
EdwardAngert: @obicke 
`partners` here could be taken to mean that PS only works with Pantheon Partners through the Partner Program or some other form of partner organization. I think you mean to use some friendlier version of "customers who've contracted our service,"  and if so, we can try to find a way to make that more clear.
want me to take a stab at that?

obicke: I don't think that "customers who've contracted our service" is right. I think that the primary idea is to avoid being seen as *in competition* with agencies, but rather that we work with them. There's also the other sense, wherein we works with 3rd-party vendors to provide some of these solutions. Perhaps, "agencies" is a better word. I'm sure Ben will have a solid suggestion here.

obicke: Discussed this with Ben... we are comfortable leaving it as is, even if somewhat ambiguous.

bentekwork: The goal here is to be ambiguous.  If they have questions we would like to talk to them in more detail while being sensitive to their specific perspective.

PR #4563, opened 2019-03-14
use deploy tokens for Gitlab custom upstreams

### LINKED ISSUE #4562: Gitlab upstream docs didn't work; chat rep was able to add using deploy token but not API key
Re: https://pantheon.io/docs/create-custom-upstream/

Related to  #3492

Priority: Low

## Issue Description

When trying to add a Gitlab upstream by the recommended method (create a new user with reporter permission, create an OAuth access token, chat support with credentials), custom support was unable to access repo. We tried a Gitlab deploy token instead, which uses HTTP basic auth to access a private repo and it worked fine.

If there's no reason to use the API, it's much easier (and more secure) for Gitlab maintainers to create deploy tokens to provide Pantheon access to repos.

## Suggested Resolution

PR incoming to update docs.

ahebrank:
Closes  #4562

## Effect
- Update Gitlab custom upstream docs to suggest creating a deploy token for a repo rather than a new user + API token

## Remaining Work
- [x] Technical review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report


alexfornuto:
CC @leematos

leematos:
This looks reasonable to me. It should work for GitLab.com and any self hosted instance. 👍

alexfornuto:
Thanks @ahebrank! We appreciate you helping keep our docs up to date.

alexfornuto:
@ahebrank #4581 suggests that this method doesn't work, in that new commits aren't showing up on site dashboards. Can you confirm that you've received custom upstream updates on sites where the custom upstream uses this token?

ahebrank:
#4581 is correct, and I'm going to add some additional info once I hear back from Pantheon support and confirm that there is a way that works. But this should probably be reverted as a start.

PR #4575, opened 2019-03-19
Adjust example Drush commands

sarahg:
https://pantheon.io/docs/drush/#terminus-drush-and-local-drush

## Effect
- Adjusts example Drush commands shown where we explain that Terminus Drush and Drush aliases can do the same thing. This is clearer if the commands are the same ("drush cc" and "drush cc drush" are different commands).

## Remaining Work
- [x] technical review

## Post Launch
- [ ] Update Status Report


alexfornuto:
Thanks @sarahg!

PR #4578, opened 2019-03-19
expand instructions for expiring cached assets

### LINKED ISSUE #4557: Suggest a specific solution for static asset caching issues of css and js files
Re: Caching: Advanced Topics -- https://pantheon.io/docs/caching-advanced-topics/

Specifically the section on Static Assets:
https://pantheon.io/docs/caching-advanced-topics/#manually-expiring-cache-for-static-assets-eg-css-js-images

Priority: Medium

(probably closer to low, but this does come up fairly frequently so adding this to docs may deflect some chats / tickets, or at least make them quicker to resolve...)

## Issue Description

Add additional information on how to remedy the 1 year caching issue for static css & js files by aggregating the site's css / js.

## Suggested Resolution

For Drupal sites, it is standard practice - with an option built into the CMS - to aggregate css & js files for Live environments. This combines all the css for the site into a single file stored in sites/default/files/css, and all js in sites/default/files/js. Each time the cache is cleared, the css & js are re-aggregated resulting in a new file name for each being generated, which effectively eliminates the static caching issue for Drupal sites using aggregation.

It is possible for WordPress sites to do the same thing, though the option is only available for those sites that have installed, activated, and configured a plugin that will handle this aggregation process.

In our Frontend Performance doc (https://pantheon.io/docs/guides/frontend-performance/) there is a section that refers to using the Autoptimize plugin to aggregate css & js:

https://pantheon.io/docs/guides/frontend-performance/#use-as-few-requests-as-possible

A google search turned up other plugins that look like they'd achieve the same end result, but since we're already documenting the use of Autoptimize to aggregate css & js elsewhere, it seems logical to add a reference to that doc in the Static Assets section of the "Caching: Advanced Topics" doc as well.

I wasn't sure how best to rearrange that section so I opted to create this issue rather than propose an edit directly in the doc, but were I to suggest a change, it would be to separate the solution for css & js from the existing solution (indexing file names) for images, pdfs, and other static assets.

I'll be happy to take a stab at it if you like that way of presenting it, and I'm also happy to have the pros handle it as well :)

Questions, let me know!

alexfornuto:
Closes #4557

## Effect
- Add Drupal and WP options to expire static assets.

## Remaining Work
- [x] :+1: from @xq42
- [ ] Technical review
- [x] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/caching-advanced-topics.md
+++ b/source/_docs/caching-advanced-topics.md
@@ -34,18 +34,18 @@ For more information, see [PANTHEON_STRIPPED GET Parameter Values](/docs/pantheo
 
 If your site or application requires Facebook authentication, we have added exceptions for this to allow users to register and log in. In the event you are having problems with another external authentication service, please contact us and let us know what service you are having issues with.
 
-## Manually Expiring Cache for Static Assets (e.g. CSS, JS, Images)
+## Manually Expiring Cache for Static Assets (e.g. CSS, JS, Images, PDFs)
 
-Pantheon sets a cache lifetime of 1 year for static assets per industry standard best practices. Drupal and WordPress each offer different methods to update cached assets:
+Pantheon sets a cache lifetime of 1 year for static assets per industry standard best practices. Either of the following options should ensure a client's browser receives a new version of any static asset after clearing a site's cache:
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
Assigning @xq42 who's going to add his edits.

PR #4597, opened 2019-03-22
Pantheon API module workaround

### LINKED ISSUE #4497: Pantheon API - Image Path Flush
Re: [Pantheon Modules](https://pantheon.io/docs/modules/)

Priority: Medium

## Issue Description:
Ticket: 115619

Related to Drupal 7, Pantheon API module, a site may be using multiple image styles and activated filefield_paths module. This can cause Drupal to call image_path_flush function 300-700 times just co clear the number of styles against the number of images attached to a node. During node edit, this can cause time-outs for the transaction.

## Suggested Resolution

As a workaround, it is possible to add a workaround to add the image path files clearing using the Drupal Queue API and be executed during cron run. An example change done by one customer is found here:

``` // Image.module
function image_path_flush($path) {
  // Following pantheon function isn't even used and subverts the fix.
  // Pantheon optimization: one bulk operation is better.
  // if (function_exists('pantheon_api_image_path_flush') && function_exists('pantheon_bulk_file_delete')) {
  //  pantheon_api_image_path_flush($path);
  //  return;
  //}
  $styles = image_styles();
  foreach ($styles as $style) {
    $image_path = image_style_path($style['name'], $path);
    if (defined('PANTHEON_ENVIRONMENT')) {
      file_unmanaged_delete($image_path);
      $data = array('image_path' => $image_path);
      $queue_name = 'pantheon_api_performance_file_workaround';
      $result = db_query('SELECT item_id FROM {queue} 
                                        WHERE name = :name 
                                        AND expire = :expire AND data = :data', 
                                        array(':name' => $queue_name, ':expire' => 0, ':data' => serialize($data)));
      if ($result->rowCount() == 0) {
        $queue = DrupalQueue::get($queue_name);
        $queue->createQueue();
        $queue->createItem($data);
      }
    }
    else {
      if (file_exists($image_path)) {
        file_unmanaged_delete($image_path);
      }
    }
  }
}
```

```
/** Implements hook_cron_queue_info(). */

function tlr_system_cron_queue_info() {
    $queues['pantheon_api_performance_file_workaround'] = array(
        'worker callback' => 'pantheon_api_process_image_style_flush',
        'time' => 30
        );
    return $queues;
}

/** Queue callback 
Processes a single image style path flush per logic in Drupal core image_path_flush() function. 
*/

function pantheon_api_process_image_style_flush($args) {

  if (!$args['image_path']) {
    return;
  }
  
  if (file_exists($args['image_path'])) {
	  file_unmanaged_delete($args['image_path']);
  }
}
```

eabquina:
Pantheon API module workaround

Closes #4497 

## Effect
- An additional section for the workaround
-

## Remaining Work
- [ ] Technical review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/_docs/modules.md
+++ b/source/_docs/modules.md
@@ -41,7 +41,9 @@ Note: Currently, there is no API module available for Drupal 8.
 
 Issue: For sites with Filefield Paths module enabled and a lot of image styles, some transactions may hit some performance issues as image paths are flushed due to its volume. 
 
-Workaround: Add the image path files clearing execution using the Drupal Queue API and be executed during cron run
+Workaround: One possible workaround for this is to Add the image path files clearing execution using the Drupal Queue API and be executed during cron run. There may be other ways of doing this depending on a site and one should determine how to implement this better on their specific site before copying the snippet below:
+
+[Line 13 of pantheon_api.module](https://github.com/pantheon-systems/drops-7/blob/default/modules/pantheon/pantheon_api/pantheon_api.module#L13)
 
 ```
 function image_path_flush($path) {
@@ -54,7 +56,9 @@ function image_path_flush($path) {
   $styles = image_styles();
   foreach ($styles as $style) {
     $image_path = image_style_path($style['name'], $path);
-    if (defined('PANTHEON_ENVIRONMENT')) {
+    // Apply these only when in the Pantheon Environment
+    // Can exclude specific environments, or local like Lando
+    if (defined('PANTHEON_ENVIRONMENT' && PANTHEON_ENVIRONMENT != 'lando')) {
       file_unmanaged_delete($image_path);
       $data = array('image_path' => $image_path);
       $queue_name = 'pantheon_api_performance_file_workaround';
--- a/source/_docs/modules.md
+++ b/source/_docs/modules.md
@@ -39,6 +39,71 @@ This module provides general methods your site needs to access aspects of the in
 
 Note: Currently, there is no API module available for Drupal 8.
 
+Issue: For sites with Filefield Paths module enabled and a lot of image styles, some transactions may hit some performance issues as image paths are flushed due to its volume. 
+
+Workaround: Add the image path files clearing execution using the Drupal Queue API and be executed during cron run
+
+```
+function image_path_flush($path) {
+  // Following pantheon function isn't even used and subverts the fix.
+  // Pantheon optimization: one bulk operation is better.
+  // if (function_exists('pantheon_api_image_path_flush') && function_exists('pantheon_bulk_file_delete')) {
+  //  pantheon_api_image_path_flush($path);
+  //  return;
+  //}
+  $styles = image_styles();
+  foreach ($styles as $style) {
+    $image_path = image_style_path($style['name'], $path);
+    if (defined('PANTHEON_ENVIRONMENT')) {
+      file_unmanaged_delete($image_path);
+      $data = array('image_path' => $image_path);
+      $queue_name = 'pantheon_api_performance_file_workaround';
+      $result = db_query('SELECT item_id FROM {queue} 
+                                        WHERE name = :name 
+                                        AND expire = :expire AND data = :data', 
+                                        array(':name' => $queue_name, ':expire' => 0, ':data' => serialize($data)));
+      if ($result->rowCount() == 0) {
+        $queue = DrupalQueue::get($queue_name);
+        $queue->createQueue();
+        $queue->createItem($data);
+      }
+    }
+    else {
+      if (file_exists($image_path)) {
+        file_unmanaged_delete($image_path);
+      }
+    }
+  }
+}
+```
+
+```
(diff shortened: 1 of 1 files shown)


source/_docs/modules.md
```diff
@@ -39,6 +39,71 @@ This module provides general methods your site needs to access aspects of the in
 
 Note: Currently, there is no API module available for Drupal 8.
 
+Issue: For sites with Filefield Paths module enabled and a lot of image styles, some transactions may hit some performance issues as image paths are flushed due to its volume. 
+
+Workaround: Add the image path files clearing execution using the Drupal Queue API and be executed during cron run
+
+```
```
alexfornuto: @eabquina Where is this (what file), and it is an addition or a modification?

source/_docs/modules.md
```diff
@@ -39,6 +39,71 @@ This module provides general methods your site needs to access aspects of the in
...
+  // Following pantheon function isn't even used and subverts the fix.
+  // Pantheon optimization: one bulk operation is better.
+  // if (function_exists('pantheon_api_image_path_flush') && function_exists('pantheon_bulk_file_delete')) {
+  //  pantheon_api_image_path_flush($path);
+  //  return;
+  //}
+  $styles = image_styles();
+  foreach ($styles as $style) {
+    $image_path = image_style_path($style['name'], $path);
+    if (defined('PANTHEON_ENVIRONMENT')) {
```
jdleonard: I think it makes sense to exclude LocalDev/Lando environments here (i.e.  `&& PANTHEON_ENVIRONMENT != 'lando'`). Assuming that the performance problem is specific to Pantheon's filesystem, then it doesn't make sense to change the behavior on local environments that don't share that problem. Also, developers often don't have cron set up locally the same way it is configured in the cloud so, as written, this workaround could result in the queue rarely being processed locally.

source/_docs/modules.md
```diff
@@ -39,6 +39,71 @@ This module provides general methods your site needs to access aspects of the in
...
+      file_unmanaged_delete($image_path);
+      $data = array('image_path' => $image_path);
+      $queue_name = 'pantheon_api_performance_file_workaround';
+      $result = db_query('SELECT item_id FROM {queue} 
+                                        WHERE name = :name 
+                                        AND expire = :expire AND data = :data', 
+                                        array(':name' => $queue_name, ':expire' => 0, ':data' => serialize($data)));
+      if ($result->rowCount() == 0) {
+        $queue = DrupalQueue::get($queue_name);
```
jdleonard: For performance, shouldn't getting and creating the queue be outside of the for loop? While we're at it, the $queue_name and possibly query template could be too.

source/_docs/modules.md
```diff
@@ -39,6 +39,71 @@ This module provides general methods your site needs to access aspects of the in
...
+  // Pantheon optimization: one bulk operation is better.
+  // if (function_exists('pantheon_api_image_path_flush') && function_exists('pantheon_bulk_file_delete')) {
+  //  pantheon_api_image_path_flush($path);
+  //  return;
+  //}
+  $styles = image_styles();
+  foreach ($styles as $style) {
+    $image_path = image_style_path($style['name'], $path);
+    if (defined('PANTHEON_ENVIRONMENT')) {
+      file_unmanaged_delete($image_path);
```
jdleonard: This workaround moves execution of file_unmanaged_delete() to a cron queue callback and should not also be here.

source/_docs/modules.md
eabquina: @alexfornuto  This is for the file in the Pantheon API module

https://github.com/pantheon-systems/drops-7/blob/default/modules/pantheon/pantheon_api/pantheon_api.module

eabquina:
@alexfornuto  added a new commit to specify the file to be changed. 
This snippet is actually from a specific customer, and what appears to be working for their use case. I've modified the statement that this can be a workaround that may be tweaked a bit by anyone who would use this depending on their issues or scenario

cc @jdleonard -- feel free to recommend another version of this. The intention here is just to show an example for a specific use case

alexfornuto:
@jdleonard Would you care to take a pass as the lines you highlighted for adjustment?

source/_docs/modules.md
alexfornuto: @eabquina ?

alexfornuto:
@greg-1-anderson can you take a look at this? I'm out of my depth on this one and I'm not sure if we should document this as is, or at all.

alexfornuto:
@eabquina I applied your fix on a test site and got the following error:

```
Fatal error: Cannot redeclare image_path_flush() (previously declared in /srv/bindings/UUID/code/modules/image/image.module:548) in /srv/bindings/UUID/code/modules/pantheon/pantheon_api/pantheon_api.module on line 13

alexfornuto:
As discussed in #4497, This issue is being tracked as `BUGS-2661` to be fixed upstream, which is preferable to us documenting a patch to the module here.

PR #4651, opened 2019-04-06
Clean core update via sftp

### LINKED ISSUE #4650: WordPress and Drupal Core Updates Doc Update
Re: [WordPress and Drupal Core Updates](https://pantheon.io/docs/core-updates/)

Priority: Low

## Issue Description:
There are users having issues updating their WP core via dashboard due to merge conflicts and the same time having issues in using Git, last option is to use the SFTP way in overwriting WP core 

## Suggested Resolution
Add a section in the doc to do a clean WP core update via SFTP

carl-alberto:
Closes #4650

## Effect
- Added a section in the doc to do a clean WP core update via SFTP

## Remaining Work
- [x] Technical review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report


alexfornuto:
@carl-alberto this seems like a very drastic step to be documenting. I'd like to hear from others in @pantheon-systems/cse to see if there's a consensus that we should document this method.

eabquina:
Sometimes, this is the last resort that we suggest to customers. (Both WP and Drupal). 
For both scenarios, we can specifically mention to delete core files, and retain their custom code or content.

WP = /wp-content/*
D7 = /sites/*
D8 = /modules, /themes, /sites/*

newtoid:
@alexfornuto I agree that this is a drastic measure, but it is also a real option, and simpler for the less git savvy. I would also agree with @eabquina's comments. This is a method I have used in the past when our standard method using -xtheirs fails to remedy issues.

lauren-kelly:
another :+1:  on this being needed and this is a great place for it. 
@sarahg @newtoid @carl-alberto do you think it is worth tossing in a note/warning box to leave the wp-config file alone?

alexfornuto:
Re-questing re-review from @lauren-kelly, @sarahg, and/or @newtoid.

lauren-kelly:
Beautiful. :+1:

Do we need a similar one for Drupal 7 / Drupal 8?

alexfornuto:
@lauren-kelly I wouldn't say no to an update with d7/d8 instructions, but for the sake of keeping this PR moving I'd say that should go in a new PR.

PR #4689, opened 2019-04-24
Adding a note that users can skip the htaccess configuration in the Multisite setup

carl-alberto:
## Effect
- Adding a note that users can skip the `.htaccess` configuration in the Multisite setup

## Remaining Work
- [ ] Technical review
- [x] Copy Review

## Post Launch
- [ ] Update Status Report


alexfornuto:
@carl-alberto thanks for the note! I see you've added it in between two steps that modify the same file. What's your reasoning for the placement? When in the process does the user actually see this warning?

carl-alberto:
@alexfornuto That `htaccess` modification message appears usually appears when `define( 'MULTISITE', true );` this is defined in the wp-config.php

But upon revisiting the documentation, I guess that note might be better to be put in after step 7 because that is where we instruct them to save the wp-config file

PR #4822, opened 2019-06-20
Added troubleshooting section on how to determine which is breaking the cache

### LINKED ISSUE #4821: WordPress and PHP Sessions Doc Update add troubleshooting caching and sessions issue
Re: [WordPress and PHP Sessions](https://pantheon.io/docs/wordpress-sessions/)

Priority: Low

## Issue Description:
There are lot of issues where SESSION cookie is breaking the cache

## Suggested Resolution
It would be helpful for users if the common steps in troubleshooting the issue, documented suggested steps for them to determine which plugin/thems is causing the issue

carl-alberto:
Closes #4821

## Effect
- Added troubleshooting section on how to determine which is breaking the cache

## Remaining Work
- [x] Technical review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/content/wordpress-sessions.md
+++ b/source/content/wordpress-sessions.md
@@ -21,6 +21,27 @@ Warning: session_start(): user session functions not defined
 ```
 Plugins with session-using code are relying on PHP's default session manager, which is temporary files on local disk. Pantheon does not support this because it will not work properly in our distributed environment.
 
+## Varnish or caching is not working when a specific plugin or theme that uses `$_SESSIONS` is enabled
+
+Symptoms of this issue shows when the header is inspected, you will see that the session cookie is always set on every page load eg. 
+
+```
+Set-Cookie: SESS1234XXXXXXXXXXXXXX path=/; domain=.example.pantheonsite.io; HttpOnly
+```
+
+The best way to determine which plugin or theme is not caching your site is by using this [tool](https://varnishcheck.pantheon.io/) with each of the steps until you determine which component is breaking the cache:
+1) To check if it is the theme that is breaking the cache, use your default theme like twentynineteen and check with the tool.
+2) To check if it is any of the plugin is breaking the cache, disable all the plugins and turn it on one by one while checking with the tool if it breaks the cache.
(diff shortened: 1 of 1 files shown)


jfussion:
@carl-alberto I suggest for a reverse approach. Disabling all the plugins first will allow the page to be cached. Varnish checker will still get a cached version of the page until that cache expires.

The reverse approach is to disable plugin one by one then check the page if it starts to be cached. The last plugin disabled is probably the culprit

alexfornuto:
I've assigned this to both of you to discuss and figure out the best solution. don't worry about the merge conflicts, I'll handle that later.

carl-alberto:
Thanks @alexfornuto 

Please review the revision @jfussion 

Ty!

source/_docs/wordpress-sessions.md
```diff
@@ -24,25 +24,20 @@ Plugins with session-using code are relying on PHP's default session manager, wh
...
 Symptoms of this issue shows when the header is inspected, you will see that the session cookie is always set on every page load eg. 
 
 ```
 Set-Cookie: SESS1234XXXXXXXXXXXXXX path=/; domain=.example.pantheonsite.io; HttpOnly
 ```
 
-The best way to determine which plugin or theme is not caching your site is by using this [tool](https://varnishcheck.pantheon.io/) with each of the steps until you determine which component is breaking the cache:
+The best way to determine which plugin or theme is not caching your site is by using this [tool](https://varnishcheck.pantheon.io/) or by inspecting the headers using `curl -sI example.com` with each of the steps until you determine which component is breaking the cache:
 1) To check if it is the theme that is breaking the cache, use your default theme like twentynineteen and check with the tool.
 2) To check if it is any of the plugin is breaking the cache, disable the plugins and turn it on one by one while checking with the tool if it breaks the cache. Do not forget to clear the cache from the Pantheon dashboard everytime that you disable a plugin.
```
jfussion: @carl-alberto Can we reverse the approach here? Disable the plugin one by one while checking with the tool if the page starts to cache again. The last plugin that was disabled is probably the culprit why the page is not being cached.

carl-alberto: Sounds good @jfussion, revised the TS as per your suggestion, please review. Thanks!

alexfornuto:
@carl-alberto please review my copy edits.

alexfornuto:
Moving on...

PR #4857, opened 2019-07-02
note interaction bypass when scripting Terminus

### LINKED ISSUE #4667: New Doc Proposal

## Title
Scripting Terminus page suggestion

## Description
I would highly recommend adding an example that takes some sort of action which, in an unscripted environment, requires some sort of actionable interactive prompt. Examples could be database imports or drush commands. These typically require interactive prompts and showing the use of the optional arguments for "Do not ask any interactive question" and "Answer all confirmations with "yes"" would be very helpful and avoid the need to dig through the help commands.

## Expected Audience
Developers

## Path
https://pantheon.io/docs/terminus/scripting/

alexfornuto:
Closes #4667

## Effect
- New section on Terminus scripting, reminding user to bypass interaction prompts with flags. 

## Remaining Work
- [x] Optional: feedback from @jeffpignataro
- [ ] Technical review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/content/terminus/scripting.md
+++ b/source/content/terminus/scripting.md
@@ -41,6 +41,9 @@ PANTHEON_MULTIDEV_LIST="$(terminus multidev:list -n ${TERMINUS_SITE} --format=li
 
 This example assumes the variable `TERMINUS_SITE` is already set. Now you can iterate through `$PANTHEON_MULTIDEV_LIST` using something like a `while read` loop to perform tasks on each multidev environment.
 
+## Interactive Prompts
+Keep in mind that any commands which normally require interaction from the user will need to be bypassed with the appropriate flag. For most Terminus commands, the flags `-y` or `--yes` will bypass requests to confirm actions.
+
 ## Example Bash Scripts
 
 ### Take a backup of the live environment of all sites in an organization


jeffpignataro:
Nice addition. Thank you for including it. I'm not sure if it's necessary, but it feel like something that would be sensible to place in the Note html block.

`<div class="alert alert-info"><h4 class="note">Note</h4><p markdown="1">Keep in mind that any commands which normally require interaction from the user will need to be bypassed with the appropriate flag. For most Terminus commands, the flags `-y` or `--yes` will bypass requests to confirm actions.</p></h4></div>`

Come to think of it, I guess I could make this update myself assuming you all are open to public pull requests.

alexfornuto:
@jeffpignataro you _could_, but since I'm making this PR against the Gatsby branch (in-progress replacement), we'd want to use an MDX component instead of a div. I did it this way to that it was its own section to be linked to. Under what section would you add that note?

jeffpignataro:
@alexfornuto Ok, that makes sense. I'm totally fine with it that way. I thought the note format made more sense since it was tangential information relating to Bash Scripts as opposed to a section of it's own. That being said, I think having it be a section that can be linked to from other places - as it's possible it could be referenced in a variety of places (not the least of which could be StackOverflow) - is a great idea.

alexfornuto:
@jeffpignataro you make a great point as well, that this is pretty specific to the use case of bash scripting. If I make it a sub-header of that section if will satisfy both points, being part of the bash section and also directly linkable. I'll stage it like that shortly.

jeffpignataro:
@alexfornuto Great idea. Yes, I think that very cleanly accomplishes both sides. Appreciate the work you are doing.

alexfornuto:
@jeffpignataro sorry, I removed my comment thinking you couldn't _possibly_ have already seen it. Upon re-visiting this page I realized that you were probably referring to the "Example Bash Scripts" section. By the time the reader gets there, we should be done with introducing new concepts to the reader and only demonstrating what we've already discussed. With that in mind, I don't see any other section that would be appropriate to nest this under.

alexfornuto:
@EdwardAngert This is "ready" for copy review, in that I've asked the Terminus team to provide technical feedback, but don't expect it or consider it a blocker.

PR #4876, opened 2019-07-05
Section to lock LIVE if they really want to restore LIVE first

### LINKED ISSUE #4466: Restore Backups Section
Re: [Backups Tool](https://pantheon.io/docs/backups/)

Priority: Low

## Issue Description:
Large sites that have more than 100GB files can take too long to restore and sometimes may fail. We can provide additional tips on how to restore better.

Also, other sites who exceed the 100GB archive size fails their back-up run. We can provide comments on what to expect.

## Suggested Resolution

One tip comes to mind is Lock the environment temporarily to avoid any activities during the restoration period.

eabquina:
New Section

Closes #4466 

## Effect
- New section to highlight impact and tips for LIVE restore
- FAQ about how long it will take to restore

## Remaining Work
- [ ] Technical review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/content/restore-environment-backup.md
+++ b/source/content/restore-environment-backup.md
@@ -87,3 +87,37 @@ This is a destructive process. If you're not comfortable with this technique, us
 5. Deploy the change from Dev up to Test and Live.
 
 For more information, see [git reset](https://git-scm.com/docs/git-reset).
+
+## Restoring the LIVE environment
+The restore process removes any recent content or changes applied to your site, since the date of the backup file you will use. Doing so in LIVE means you will lose them forever with no way to recover.
+
+If there is still a need to do this in the LIVE environment, we recommend doing the following steps:
+- [Lock the environment first](/docs/security/)
+- [Run a backup](/docs/backups/) of LIVE, in order to have a copy of files and the database for later use
+- Run the restore while the site is still locked
+
+Doing these steps will allow you to later recreate any new content manually. It also resticts access while the restore process is still running, which is a good practice. 
+
+## Restoring Large Sites
+Large sites that have more than 100GB files can take too long to restore and are likely to fail. We recommend running the steps above for a safer process, or do use the separate [backup archive files as mentioned above](/docs/restore-environment-backup/#restore-an-environment-from-another-environments-backup).
+
+## Frequently Asked Questions
+
+### How long does the restore process will take?
+When the Restore button is pressed, three (3) separate workflow process are triggered in the dashboard. Each one represent code, database and asset media files like images or other attachments). One workflow may complete ahead of the others.
+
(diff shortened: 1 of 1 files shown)


source/content/restore-environment-backup.md
```diff
@@ -87,3 +87,37 @@ This is a destructive process. If you're not comfortable with this technique, us
 5. Deploy the change from Dev up to Test and Live.
 
 For more information, see [git reset](https://git-scm.com/docs/git-reset).
+
+## Restoring the LIVE environment
```
alexfornuto: This section seems to directly conflict with [this warning](https://github.com/pantheon-systems/documentation/pull/4876/files#diff-25d90242c348ba248897ea37c6708856R12) above. Why are we contradicting it?

source/content/restore-environment-backup.md
```diff
@@ -87,3 +87,37 @@ This is a destructive process. If you're not comfortable with this technique, us
...
+The restore process removes any recent content or changes applied to your site, since the date of the backup file you will use. Doing so in LIVE means you will lose them forever with no way to recover.
+
+If there is still a need to do this in the LIVE environment, we recommend doing the following steps:
+- [Lock the environment first](/docs/security/)
+- [Run a backup](/docs/backups/) of LIVE, in order to have a copy of files and the database for later use
```
alexfornuto: I assume you would only do this to get any content not including in older backups? If so we can mention that on L99.

source/content/restore-environment-backup.md
```diff
@@ -87,3 +87,37 @@ This is a destructive process. If you're not comfortable with this technique, us
...
+- Run the restore while the site is still locked
+
+Doing these steps will allow you to later recreate any new content manually. It also resticts access while the restore process is still running, which is a good practice. 
+
+## Restoring Large Sites
+Large sites that have more than 100GB files can take too long to restore and are likely to fail. We recommend running the steps above for a safer process, or do use the separate [backup archive files as mentioned above](/docs/restore-environment-backup/#restore-an-environment-from-another-environments-backup).
```
alexfornuto: Which steps above? Aren't these all steps to restore?

source/content/restore-environment-backup.md
eabquina: In reality, customers might still need to do this (i.e., They backups of TEST is way behind LIVE, or restoring in TEST will also not work etc.)

source/content/restore-environment-backup.md
eabquina: Ah right, I agree

source/content/restore-environment-backup.md
eabquina: above points to the steps that involve locking LIVE temporarily etc.

PR #4936, opened 2019-07-22
Fix terminus site:create codeblock

rachelwhitton:
## Effect
- Adjust spacing in the preceding markdown table so that code fence renders as a nested entity of the ordered list


rachelwhitton:
nvm the table inside the ordered list is throwing things off - closing/abandoning effort

PR #4944, opened 2019-07-25
Update composer-convert.md

### LINKED ISSUE #4940: Convert a Standard Drupal 8 Site to a Composer Managed Site Doc Update - Getting Proper Module Versions
Re: [Convert a Standard Drupal 8 Site to a Composer Managed Site](https://pantheon.io/docs/guides/composer-convert/)

Priority: Low

## Issue Description:
The docs talk about using `composer require` to download all of the contrib modules that the site is using, but the docs don't describe how you ensure that the version of those modules you are getting are compatible with the version you previously had.

For example, on a D8 site we're playing around with right now, we use version 2.x of the Conflict module, but running `composer require "drupal/conflict"` pulls down version 1. I believe this is because of the stability policy in place in the `composer.json` file but I'm not 100% sure.

## Suggested Resolution
The docs should probably include two notes:
- If the site depends on any modules that are unstable (e.g. dev, alpha, beta), then it may be necessary to modify the `minimum-stability` setting in the `composer.json` file.
- It's a good idea to determine what version of each module is being used by inspecting its `.info.yml` file, then requiring at least that version by specifying the constraint when running `composer require`. For example: `composer require "drupal/conflict:^2.0-alpha1"`

GuyPaddock:
## Effect
- Adds a note about `minimum-stability` adjustments which may be required for sites that are looser with what modules they depend on (closes #4940).
- Adds guidance about how to determine what versions of modules are installed, and to use those versions when pulling those modules in via Composer instead of deploying whatever the resolved version ends up being (closes #4940).
- Adds the necessary steps for omitting dev modules from what gets pushed to Pantheon when not using a CI workflow (this does NOT appear to be handled by `prepare-for-pantheon`). With the previous guide, these extra steps help to avoid the message `warning: adding embedded git repository: vendor/drupal/coder` from GIT that was signaling that part of the dependencies were not getting committed.
- Makes minor formatting corrections.

## Remaining Work
- [ ] Technical review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/content/guides/composer-convert.md
+++ b/source/content/guides/composer-convert.md
@@ -16,10 +16,7 @@ Converting to a Composer managed site *removes* the ability to [apply updates vi
 </Alert>
 
 ## Before You Begin
-
-<Alert title="Note" type="info">
-
-As packages pulled by Composer are updated (along with their dependencies), version compatibility issues can pop up. Sometimes you may need to manually alter the version constraints on a given package within the `require` or `require-dev` section of `composer.json` in order to update packages. See the [updating dependencies](https://getcomposer.org/doc/01-basic-usage.md#updating-dependencies-to-their-latest-versions) section of Composer's documentation for more information.
+{% include("content/composer-updating.html")%}
 
 As a first troubleshooting step, try running `composer update` to bring `composer.lock` up to date with the latest available packages (as constrained by the version requirements in `composer.json`).
 
@@ -57,8 +54,7 @@ You're about to make some massive changes to the codebase. We recommend you to d
   You can replace `composify` with a branch name of your choosing, but all following steps assume this name.
 
 ## Setup a Multidev (optional)
-
-If your Pantheon account has access to [multidev](/multidev/), create a multidev to push your new code to:
+If your Pantheon account has access to [multidev](/docs/multidev/), create a multidev to push your new code to:
 
 ```bash
 git push origin composify && terminus env:create $site.dev composify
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This looks very good, but the process is getting a bit complicated. n.b. the documented process produces the wrong result for some sites, so there is cause to be more robust.

We might suggest using https://github.com/grasmash/composerize-drupal though.

carolyn-shannon:
take a peek at conflicts

alexfornuto:
@carolyn-shannon conflicts resolved. ready for SME review.

source/content/guides/composer-convert.md
```diff
@@ -161,11 +176,12 @@ From the `$site-composer` directory, run the following:
...
 ```bash
 composer prepare-for-pantheon
+composer install --no-dev
 ```
 
-The output `> DrupalProject\composer\ScriptHandler::prepareForPantheon` means the command was successful. This should modify the `.gitignore` file and cleanup any errant `.git` directories in the codebase, to prepare your new code for direct deployment to Pantheon.
+This should modify the `.gitignore` file and cleanup any errant `.git` directories in the codebase, to prepare your new code for direct deployment to Pantheon.
 
-### Commit
+## Commit
```
alexfornuto: While I agree that an H2 makes more sense in context of the other headers, I'd like to see if we can't collapse this ToC overall to use H3s more effectively.

carolyn-shannon: this makes sense

PR #4980, opened 2019-08-06
Add missing cache-busting cookie pattern

jfussion:
## Effect
- Adds `Drupal[a-zA-Z0-9_\.-]+` to cache-busting cookie.

## Post Launch
- [ ] Update Status Report


jfussion:
Checked vcl again and this is not a cache-busting cookie

PR #5004, opened 2019-08-09
Implement Caching in CircleCI builds

alexfornuto:
## Effect
- Adjusts the CircleCI configuration file to cache content for faster builds and deploys

## Remaining Work
- [x] Confirm caching works
- [x] Confirm caching doesn't break anything
- [ ] Determine what else we might want to cache.
- [x] Technical review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,11 +8,9 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - npm-v1-{{ checksum "gatsby/package-lock.json" }}
-            - composer-v1-{{ checksum "composer.lock" }}
-      - restore_cache:
-          keys:
-            - always-
+            - v1-always-
+            - npm-v2-{{ checksum "package-lock.json" }}
+            - composer-v2-{{ checksum "composer.lock" }}
       - run:
           name: Prep Environment
           command: |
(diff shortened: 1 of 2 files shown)


alexfornuto:
The [latest build](https://circleci.com/gh/pantheon-systems/documentation/14130) shows that when plugins are updated Gatsby deletes the cached `.cache` directory but still uses the cached `public` directory, and build time is still greatly reduced.

.circleci/config.yml
```diff
@@ -71,14 +72,19 @@ jobs:
...
-          key: composer-v1-{{ checksum "composer.lock" }}
+          key: npm-v2-{{ checksum "gatsby/package-lock.json" }}
+          paths:
+            - ~/build/gatsby/node_modules
+      - save_cache:
+          key: composer-v2-{{ checksum "composer.lock" }}
           paths:
             - ~/.composer/cache
+      - save_cache:
+          key: v1-always-{{ epoch }}
```
stevector: @alexfornuto I'm not sure what the benefit is of saving this key

> myapp-{{ epoch }} - Every build will generate separate cache keys.

From https://circleci.com/docs/2.0/caching/

If a cache is never used, why make it?

DSchau: ^ I think Steve is correct here. You want a less random cache key, otherwise every time you trigger a re-build it's going to generate a new cache.

alexfornuto: My understanding here is that the restore line [here](https://github.com/pantheon-systems/documentation/pull/5004/files#diff-1d37e48f9ceff6d8030570cd36286a61R11) will pull any cache that matches the format `v1-always-`

@felicianotech would you care to weigh in?

alexfornuto: You can see [here](https://circleci.com/gh/pantheon-systems/documentation/14130) that the cache *is* being used.

DSchau: So in theory it'll just grab the newest cache always? I suppose that makes sense!

alexfornuto: To clarify: It will definitely create a new cache, but it will use the new cache each time on the next build - I figured this would keep the cache updated with new content.

alexfornuto: P.S. I borrowed the configuration from [here[(https://github.com/gatsbyjs/gatsby/issues/7366#issuecomment-469433003)

felicianotech: @alexfornuto Sorry, late to the party.

Yes, when it comes to cache loading, we do support partial matching on the key. I don't know your codebase well but not every package manager supports partial caches well. For some, a partial cache means it saves some of the work, which is great. For others, a partial cache is as good as no cache in which case you shouldn't do it.

If you're seeing cache hits here, and it's reducing the build time, then that sounds right to me.

PR #5070, opened 2019-08-28
Update install.md

guylyons:
Make bash snippet more copy/paste friendly.

A newline can be used after "&&" if desirable, but really doesn't add that much to readability.

## Effect
- remove newline after "curl -O"
- remove newline after "...installer.phar"


alexfornuto:
Nice catch, thanks @guylyons!

PR #5088, opened 2019-09-04
[DOC-5] guide width

digitalhydra:
## Effect
- fixes pages with missing container

### DIFF
--- a/gatsby/src/pages/contributors.js
+++ b/gatsby/src/pages/contributors.js
@@ -21,7 +21,7 @@ class Contributors extends React.Component {
 							<h1 className="title">Contributors</h1>
 						</div>
 						<div className="row mb-70">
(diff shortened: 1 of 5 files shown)


alexfornuto:
LGTM! Does @EdwardAngert, @allenfear, or @rachelwhitton want to weigh in?

rachelwhitton:
Spacing looks good, nice! 

If it's reasonable, would be preferred to have the gray background pulled over all the way to left so that there's no whitespace on that side of the nav element.

jmolivas:
@rachelwhitton @alexfornuto After checking this with @digitalhydra we got this

image

alexfornuto:
Sounds like an extra spiffiness that just isn't possible right now. Thanks for checking.

PR #5113, opened 2019-09-09
Fix UTM redirection example

sarahg:
https://pantheon.io/docs/pantheon_stripped

# Effect
- Change example URL from HTTPS -> HTTP. We're illustrating that the redirect from HTTP -> HTTPS does not strip the UTM parameters, so we need to curl the HTTP address to show this.

## Remaining Work
- [x] Copy Review

## Post Launch
- [ ] Update Status Report


alexfornuto:
Nice catch @sarahg!

PR #5131, opened 2019-09-13
Update redis.md

twooten:
Added a note to let customers know that adding Redis config to settings.pantheon.php will likely cause Git conflicts if/when upstream changes are made to settings.pantheon.php.

## Remaining Work
- [x] Wordsmithing 

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/content/redis.md
+++ b/source/content/redis.md
@@ -160,6 +160,12 @@ All plans except for the Basic plan can use Redis. Sandbox site plans can enable
       $settings['cache']['bins']['form']      = 'cache.backend.database';
     }
     ```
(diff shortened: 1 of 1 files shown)


alexfornuto:
Thanks @twooten! Do you think we should include this on the D7 tab as well?

alexfornuto:
I added that as a separate commit, in case you agree.

twooten:
@alexfornuto there is no settings.pantheon.php in our upstream, so not needed.

alexfornuto:
removed.

PR #5154, opened 2019-09-26
[fix] Subtopics alignment

digitalhydra:
## Effect
- subtopic alignment when dividing links in two groups

### DIFF
--- a/gatsby/src/layout/subtopic-group/style.css
+++ b/gatsby/src/layout/subtopic-group/style.css
@@ -20,6 +20,7 @@
   font-weight: 300;
   text-align: left;
   margin: 0px;
+  min-height: 21px;
 }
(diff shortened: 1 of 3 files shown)


digitalhydra:
Screen Shot 2019-09-25 at 8 14 45 PM

carolyn-shannon:
I'm looking at this in FF, Safari, and Chrome (macOS) and the alignment of the second column of items is higher in all three than your screenshot. My concern is that it doesn't visually associate col 2 with the content in col1. What browser were you using?
column-alignment-pr-5154

alexfornuto:
@digitalhydra it looks like this isn't working across the board:

image

alexfornuto:
And if I move the separator up by one: 
image

PR #5195, opened 2019-10-08
Ci test then deploy

alexfornuto:
## Effect
- Sets up deploy as a separate build step, dependent on testing.
- Adds the Terminus installer script to the project, to not depend on external repo.

## Remaining Work
- [x] Technical review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/scripts/terminus-installer.php
+++ b/scripts/terminus-installer.php
@@ -1,0 +1,106 @@
+<?php
+
+/* Terminus Installer Script
+Created by Alex Fornuto - alex@fornuto.com
+*/
+
+// Define environment variables.
+$pathUpdated = false;
+$paths = explode(":", getenv('PATH'));             // Creates an array with all paths in $PATH
+$installdir = ("/usr/local/bin");                  // Creates a string with the desired installation path
+$rcfiles = array(                                  // Array of common .rc files to look for.
+  ".bashrc",
+  ".zshrc",
+  ".config/fish/config.fish",
+  ".profile",
+);
+$package = "terminus";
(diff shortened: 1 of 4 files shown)


scripts/terminus-installer.php
```diff
@@ -0,0 +1,106 @@
...
+    $version  = $releases[0]->tag_name;
+    $url      = $releases[0]->assets[0]->browser_download_url;
+    // Do the needful
+    echo("\nDownloading Terminus " . $version . " from " . $url . " to /tmp \n");
+    $couldDownload = file_put_contents("/tmp/" . $package . ".phar", file_get_contents($url));
+    echo("Moving to " . $installdir . "...\n");
+    if(!rename("/tmp/" . $package . ".phar", $installdir . "/" . $package)){
+        echo("\n" . $installdir . " requires admin rights to write to...\n");
+        exec("sudo mv /tmp/" . $package . ".phar " . $installdir . "/" . $package);
+        echo("\n");
```
carolyn-shannon: Is a rename failure also possible if the directory already exists?

alexfornuto: I don't understand the question. This renames the terminus binary we downloaded into tmp, and then moves into the install directory. No directories are being renamed, and `mv` will overwrite the binary if it already exists in the target directory.

carolyn-shannon: It sounds like mv overwrites the binary and won't fail if there's something w/that name already in the directory. OK, good to go then - thanks.

PR #5208, opened 2019-10-10
Add WPML Key to modules-plugins-known-issues.md

jahzlariosa:
~This code snippet will set the site key to be registered for each or a single environment.~

## Effect
- ~Adding a code snippet to register the site key for each environment.~
- Document how to avoid an error when registering the WPML plugin.

## Remaining Work
- [x] Technical review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report


source/content/modules-plugins-known-issues.md
```diff
@@ -857,6 +857,29 @@ if (isset($_ENV['PANTHEON_ENVIRONMENT'])) {
 
 <hr />
 
+### [WPML - Automatic WPML registration using PHP for easy moves between development, test, and live environment](https://wpml.org/)
+**Issue**: Accessing /wp-admin/plugin-install.php?tab=commercial will throw `Sorry, you are not allowed to access this page`.
+
+**Solution**: Define WPML Site Key in wp-config.php 
+```php
+define( 'OTGS_INSTALLER_SITE_KEY_WPML', 'your-site-key' );
```
alexfornuto: @jahzlariosa Two questions:

1) Where does the user get the site key?
2) Is the site key something that's OK to have exposed in `wp-config.php`?

jahzlariosa: 1. Where does the user get the site key?

 - The site key is generated by the user from the plugin's website 
    https://wpml.org/account/sites/

2. Is the site key something that's OK to have exposed in wp-config.php?
- Yes it's ok :)

alexfornuto:
Technical testing will require a copy of this paid plugin, let me see if anyone in @pantheon-systems/documentation has one...

alexfornuto:
Blocked pending approval to purchase WPML license.

alexfornuto:
@jahzlariosa So I think I found an easier solution than adding the key in code manually.

image

You need to use the **Configure WPML** button instead of the link to "Purchase a subscription or enter an existing key.

Did you encounter a situation where this doesn't work?

jahzlariosa:
> @jahzlariosa So I think I found an easier solution than adding the key in code manually.
> 
> image
> 
> You need to use the **Configure WPML** button instead of the link to "Purchase a subscription or enter an existing key.
> 
> Did you encounter a situation where this doesn't work?

Yes Alex,  I had a chat where the customer was not able to use that feature and had to add the keys manually for each environment. 

I think using that feature will only set the key for the current environment.

PR #5246, opened 2019-10-21
Getting host ip to replace hostname

### LINKED ISSUE #5245: FileZilla on Pantheon Doc Update troubleshooting in servers with multiple appserver
Re: [FileZilla on Pantheon](https://pantheon.io/docs/filezilla)

Priority: Low

## Issue Description:
Previously can access test and live but now getting `Error: Directory /srv/bindings/c1f103b56xxxxxxxxxxxx/files: no such file or directory` error but can connect in dev or multidevs with single appserver

## Suggested Resolution
Add troubleshooting

carl-alberto:
Closes #5245

## Effect
- Getting host ip to replace hostname using the dig command
-

## Remaining Work
- [x] Technical review
- [ ] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/content/filezilla.md
+++ b/source/content/filezilla.md
@@ -91,7 +91,16 @@ Error:            ssh_init: nodename nor servname provided, or not known
 Error:            Could not connect to server
 ```
 
-Double check settings and resolve typos to fix this issue.
+### Cannot sftp to environment with multiple appservers
+Previously can access sftp in test and live but now getting `Error: Directory /srv/bindings/c1f103b56xxxxxxxxxxxx/files: no such file or directory` error. Symptoms include users able to connect in dev or multidevs with single appserver.
+
+Getting any of the equivalent IP of the host appserver by using a `dig` command to replace the hostname lets you connect again by:
+
+```bash
+dig +short appserver.live.120330a1-xxxxxxxxxxxxxxxxx.drush.
+35.194.x.x
+35.222.x.x 
+```
 
 ### Site Manager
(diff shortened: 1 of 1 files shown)


alexfornuto:
@carl-alberto Is this solving the same issue discussed [here](https://pantheon.io/docs/sftp#dns-connection-issues)

If so, I like the existing solution better as it will persist through the _next_ IP change, and suggest a cross-link to resolve.

carl-alberto:
@alexfornuto don't have much luck with this solution https://pantheon.io/docs/sftp#dns-connection-issues and that error it seems more specific on FileZilla users that tries to upload on a server that has multiple appservers like in test and live

alexfornuto:
@carl-alberto 10-4 good buddy. What about this [recent addition](https://pantheon.io/docs/filezilla#file-transfer-failed-on-production-environments) to the doc to limit simultaneous connections, to avoid issues when there are multiple appservers?

alexfornuto:
Pinged @carl-alberto for feedback.

carl-alberto:
Thanks for pointing this out @alexfornuto as they exhibit same symptoms, maybe adding my proposed solution as a secondary solution to use dig +short to get IP of the hostname if the existing solution wont work

source/content/filezilla.md
```diff
@@ -80,6 +82,15 @@ Resolve instances of transfer failures by reconfiguring FileZilla to limit the n
 
 See also, [this FAQ on our related SFTP doc](/sftp#i-am-receiving-errors-connecting-to-my-server-with-an-sftp-client).
 
+#### Fallback solution when file transfer fails
+If all else fails, you can connect directly to an application container via the IP address. Use Dig to find the IP address:
+
+```bash{outputLines:2-3}
+dig +short appserver.live.120330a1-xxxxxxxxxxxxxxxxx.drush.
+35.194.x.x
+35.222.x.x
```
EdwardAngert: @alexfornuto should these be changed to example IPs or are the results always going to be in that IP range? If the numbers here are consistent, then I'm 👍

alexfornuto: @carl-alberto do all Pantheon IPs fall into these two class Bs?

alexfornuto:
Thanks @carl-alberto!

PR #5225, opened 2019-10-16
Optimize image sizes

EdwardAngert:
**Note:** Please fill out the PR Template to ensure proper processing.

## Effect
PR includes the following changes:
- all images optimized using [ImageOptim](https://imageoptim.com)
- CI increment

## Remaining Work
- [ ] Make sure things aren't broken

## Post Launch
**Do not remove** - To be completed by the docs team upon merge:
- [ ] Update Status Report
- [ ] Archive from **Done** in [Main Project](https://github.com/pantheon-systems/documentation/projects/14)

PR #5249, opened 2019-10-21
Revert "Optimize image sizes"

alexfornuto:
Reverts pantheon-systems/documentation#5225

Just in case, close in a few days if no issues pop up from #5225


alexfornuto:
I think we're in the clear by now.

PR #5308, opened 2019-11-08
Best Practices for handling XML RPC on WP Sites

### LINKED ISSUE #5211: WordPress Best Practices Doc Update
Re: [WordPress Best Practices](https://pantheon.io/docs/wordpress-best-practices)

Priority: Low/Medium/High (choose one, remove the other options)

## Issue Description:
Missing content re: a common attack vector for WP sites

## Suggested Resolution
Work with SMEs to document recommendation to disable XML-RPC 

cc @ksenour 

WIP -- Kaylan and Rachel to follow up w/stronger details soon.

rachelwhitton:

Closes #5211 

## Effect
- New section to describe XML-RPC attacks for WP sites
 - Rough first pass

## Remaining Work
- [x] Review pass from @ksenour 
- [x] Review pass from @sarahg (por favor) 
- [x] Technical review
- [x] Copy Review

## Post Launch
- [ ] Update Status Report

### DIFF
--- a/source/content/wordpress-best-practices.md
+++ b/source/content/wordpress-best-practices.md
@@ -53,7 +53,7 @@ This article provides suggestions, tips, and best practices for developing and m
 * Follow our [Frontend Performance](/guides/frontend-performance/) guide to tune your WordPress site.
 
 ## Avoid XML-RPC Attacks
-There is a common attack vector for WordPress sites which involves bad actors attempting to brute force the `/xmlrpc.php` and `/wp-login.php` paths.
+The `/xml-rpc.php` is a potential security risk for WordPress sites. It can be used by bad actors to brute force administrative usernames and passwords. All calls made to this file must supply a plain text username and password. If a valid combination is found,  a success message is returned letting the bad actor know they can log in via the standard `/wp-login.php` method. 
 
 This can be surfaced by reviewing your site's `nginx-access.log` for the Live environment. If you leverage [GoAccess](/nginx-access-log/), you might see something similar to the following:
 
--- a/source/content/wordpress-best-practices.md
+++ b/source/content/wordpress-best-practices.md
@@ -87,7 +87,7 @@ This method has the advantage of being toggleable without deploying code, by act
 
 1. [Set the connection mode to SFTP](/sftp) for the Dev or target Multidev environment via the Pantheon Dashboard or with [Terminus](/terminus):
 
-  ```bash{pomptUser: user}
+  ```bash{promptUser: user}
   terminus connection:set <site>.<env> sftp
   ```
 
@@ -95,7 +95,7 @@ This method has the advantage of being toggleable without deploying code, by act
 
   In the following example, replace `my-site` with your Pantheon site name, and `disable-xmlrpc` with your preferred name for this new plugin:
 
-  ```bash{pomptUser: user}
+  ```bash{promptUser: user}
   terminus wp my-site.dev -- scaffold plugin disable-xmlrpc
   ```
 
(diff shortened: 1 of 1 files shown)


alexfornuto:
Thanks to everyone involved!

source/content/wordpress-best-practices.md
```diff
@@ -51,3 +51,44 @@ This article provides suggestions, tips, and best practices for developing and m
...
+    - /xmlrpc.php
+  ```
+
+This method is more performant than disabling via PHP since this won't involve bootstrapping the application.
+
+### Disable XML-RPC via PHP
+
+This method has the advantage of being toggleable without deploying code, by activating or deactivating a custom MU Plugin.
+
+1. If you have not already created an **MU Plugin** (Must Use Plugin) to maintain custom code, review [Create a WordPress MU Plugin for Actions and Filters](/mu-plugin/).
```
rachelwhitton: @alexfornuto MU Plugins can't be disabled, so as is the advantage of toggling does not apply

Suggest we advise they use this WP-CLI command instead (align with @sarahg's feedback)
https://developer.wordpress.org/cli/commands/scaffold/plugin/

sarahg: Would not recommend a MU plugin for this, just a regular one.

alexfornuto: @sarahg @rachelwhitton I updated this section as suggested, but I know goose-egg about making custom plugins, and don't know if this needs any other information besides the what-do code to disable XML.

alexfornuto:
@rachelwhitton I like the expanded explanation, as it tells me, the uninformed reader, how the issue works and why I should be concerned. And I'm not just "playing the part" on this one.

source/content/wordpress-best-practices.md
rachelwhitton: @alexfornuto you can assign back to me and I'll push a commit for review tomorrow

rachelwhitton:
Both methods have been tested and verified - currently, my test site (`xmlrpc-test`) for this task has the custom plugin implemented. 

```
╰─➤ $ curl -I https://dev-xmlrpc-test.pantheonsite.io/xmlrpc.php
HTTP/2 405
```

Ready for your review @alexfornuto - Thanks for your help

sarahg:
@rachelwhitton xmlrpc.php will always return a 405 on a GET request (even without a custom plugin to disable it) — xmlrpc.php only accepts POST requests. To verify a plugin is really blocking this stuff, you'd want to send a POST request with a valid method call, something like:

`curl -d '<?xml version="1.0" encoding="utf-8"?><methodCall><methodName>system.listMethods</methodName><params></params></methodCall>' https://dev-xmlrpc-test.pantheonsite.io/xmlrpc.php`

Looks like our PHP version still allows requests and returns all the available methods. I think this needs another look before merging.

sarahg:
Made a change in 4d09a1d to address my last comment here. The issue with our original `xmlrpc_enabled` filter is that it only disables xmlrpc methods which require authentication ([docs](https://developer.wordpress.org/reference/hooks/xmlrpc_enabled/)), not the whole thing. 

Disabling the authenticated methods is a fine thing to do, but since there are plenty of exploits that don't involve authentication (e.g, DoS via a pingback avalanche), we should ideally disable *all* of WP's xmlrpc methods.

Comparison of using `xmlprc_enabled` vs `xmlrpc_methods`: https://gist.github.com/sarahg/65ec209a78e054b2dad6801f81ff27e9

On that note, I think a49439d could use another pass — if we're going to dive into the "why" in this doc, this is not quite right.

PR #5460, opened 2020-01-15
Docs: removed reference to 'Elite Starter' plan

ccharlton:
I assume 'Elite Starter' was a legacy plan?


alexfornuto:
Confirmed that there are still enough sites on this plan to warrant keeping this reference. But still a good issue to have brought up @ccharlton!

ccharlton:
Do we still have lingering Elite Starter Plan customers/sites?

ccharlton:
Do we still have lingering Elite Starter Plan customers/sites?

PR #5599, opened 2020-03-25
Update services.yml doc with information about directory permissions

dutton242:

Closes: #

## Summary

**[Creating a services.yml File for Drupal 8](https://pantheon.io/docs/services-yml)** - Update services.yml documentation with information about changing sites/default directory permissions

## Effect

The following changes are already commited:

- Add information about possibly needing to change directory permissions
- Add external link to site describing how to change directory permissions

## Remaining Work

- [ ] No remaining work

## Post Launch

- [ ] Update Status Report

### DIFF
--- a/source/content/services-yml.md
+++ b/source/content/services-yml.md
@@ -23,7 +23,10 @@ The table below illustrates which services file is used in which Pantheon enviro
 
 ## Create and Modify services.yml
 1.  Before you begin, make sure that you have updated your Drupal site to version 8.2.0 or greater.
-2.  From within the `sites/default` directory, create a new `services` file. Name it based on which environment you wish to configure settings for:
+2.  If Drupal has already been installed, permissions on the `sites/default` directory may need to be temporarily changed to allow write access.
+      * [`How to change directory permissions`](https://www.pluralsight.com/blog/it-ops/linux-file-permissions)
+      * Once the .yml has been created and edited, __change permissions__ back to lock down the `sites/default` directory.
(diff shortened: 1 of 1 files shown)


alexfornuto:
Ping @carolyn-shannon as our resident Drupal wizard.

alexfornuto:
@greg-1-anderson Is there a case on Pantheon where a customer would need / want to do this? It seems counter-intuitive to our files vs code webops workflow, but I'd like your opinion please.

stevector:
I'm not sure of the exact circumstances that trigger the need to change permissions. On my local clone of a Drupal 8 site I was able to create a new `services.yml` file because the `default` directory was already set to `755`.

dutton242:
This can occur if you are doing on-server SFTP development and need to add a settings.php file, I've found that sometimes I can't create the file unless the `default` directory is set to world writable first. Then I have to switch it back. This could be Drupal 7 only? I can double check and reply back with my results.

dutton242:
OK, verified that the default permissions for that folder are 755 for Drupal 8 and I also tested a Drupal 7 site and it was set that way too and in both instances I was able to create a file without changing directory permissions. I'm not sure why one of our developers had an issue with this, but I have seen it in the past. We can close this. Thanks!

alexfornuto:
Thanks for the issue @dutton242, even if it ended up closed! I wonder if your developer having issues was trying to work over SFTP while in Git mode?

Reference: https://pantheon.io/docs/sftp#sftp-mode

Issue #4972, opened 2019-08-03
Temporary File Management Doc Update 


jfussion:
Re: [Temporary File Management](https://pantheon.io/docs/tmp/#default-temporary-path)

Priority: Medium

## Issue Description:
Changing file_temporary_path is not possible on Drupal 7 since it is being overwritten by PRESSFLOW_SETTINGS here: https://github.com/pantheon-systems/drops-7/blob/default/includes/bootstrap.inc#L759-L787

## Suggested Resolution
Remove recommendation to replace temp path.

jfussion:
@greg-1-anderson Is this a bug? Or we don't really allow changing the `file_temporary_path`?

alexfornuto:
Reached out to @greg-1-anderson for input.

carolyn-shannon:
@greg-1-anderson do you have any input on this?

alexfornuto:
@greg-1-anderson 
![image](https://user-images.githubusercontent.com/2349184/70463897-6bcf3100-1a83-11ea-9b16-7ac704060221.png)

carolyn-shannon:
Check to see if this is also true for the Drupal 8 version.

jfussion:
In Drupal 8, the configurations are in settings.pantheon.php so I believe it is possible to override the temp path via settings.php for D8.

For D7, a possible temporary workaround would be to modify the file_temporary_path value inside the $_SERVER['PRESSFLOW_SETTINGS'] which might not be a good idea?

jfussion:
To set D7's temp path, we need to modify the PRESSFLOW_SETTINGS. Ticket reference #461683

```PHP
if (isset($_ENV['PANTHEON_ENVIRONMENT'])) {

  // Pressflow will override whatever we set here unless we override Pressflow first.

  $pressflow = json_decode($_SERVER['PRESSFLOW_SETTINGS'], TRUE);

  $pressflow['conf']['file_temporary_path'] = $pressflow['conf']['file_directory_temp'] = 'sites/default/files/private/tmp';

  $_SERVER['PRESSFLOW_SETTINGS'] = json_encode($pressflow);

}
```

greg-1-anderson:
Sorry, didn't see this notification earlier. I commented on #5732

alexfornuto:
Closing without merge per discussion in #5732 5732

PR #5732, opened 2020-05-20
D7 Temp paths

alexfornuto:
Closes: #4972 

## Summary

**[Temporary File Management Doc Update](https://pantheon.io/docs/tmp)** - Adjusted example code to overwrite D7 temp paths to account for Pressflow settings.

## Post Launch

- [ ] Update Status Report

### DIFF
--- a/source/content/tmp.md
+++ b/source/content/tmp.md
@@ -17,6 +17,7 @@ export env=dev
 </Alert>
 
 ## Default Temporary Path
+
 Pantheon configures an appropriate temporary path for [WordPress](https://github.com/pantheon-systems/WordPress/blob/4.9.6/wp-config.php#L83-L86) and [Drupal 8](https://github.com/pantheon-systems/drops-8/blob/8.5.3/sites/default/settings.pantheon.php#L146-L154). Drupal 7 sites can achieve the same configuration by adding the following to `settings.php`:
 
 ```php:title=settings.php
@@ -40,6 +41,7 @@ if (isset($_ENV['PANTHEON_ENVIRONMENT'])) {
 Change `sites/default/files/private/tmp` to your desired temporary file path.
 
 ## Fix Unsupported Temporary Path
+
 Errors caused by an unsupported temporary path typically surface as permission errors for `.tmp` files and can be replicated on any environment.
 
 <TabList>
@@ -48,7 +50,7 @@ Errors caused by an unsupported temporary path typically surface as permission e
 
 Correct an unsupported temporary path set by a plugin or theme in `wp-config.php`. Replace `SOME_TMP_SETTING` with the conflicting plugin or theme option:
 
-```php
+```php:title=wp-config.php
 /**
 * WordPress
 * Fix unsupported temporary path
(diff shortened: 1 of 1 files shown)


pantheondocs:
Lighthouse Audit Results:

| Description | Live | pr-5732 | Difference |
| --- | --- | --- | --- |
| Performance Score | `43` | `47` | `4` |
| Best Practice Score | `86` | `86` | `0` |
| Accessibility Score | `96` | `96` | `0` |

Lighthouse Audit Reports:

- [`pr-5732` HTML Report](https://circleci.com/gh/pantheon-systems/documentation/21151/artifacts/0//lighthouse_data/lighthouse-audit-dev.html)
- [`Pantheon Docs` HTML Report](https://circleci.com/gh/pantheon-systems/documentation/21151/artifacts/0//lighthouse_data/lighthouse-audit-reference.html)

greg-1-anderson:
Setting the temporary path to /tmp is fine on COS, but it should remain /srv/bindings/BDIR/tmp on heirloom. The way to find the temporary path on Pantheon would be "$HOME/tmp", but the file temporary path should already be set to this location, and setting it to any other directory would be undesirable.

Is therefore should never be necessary to override file_temporary_path on Pantheon. If customers want to set the file_temporary_path on their local environment, not, then perhaps these examples should be wrapped in conditionals so that they do not run on Pantheon. If we're not running on Pantheon, though, then there won't be any pressflow settings to change.

What is the problem this documentation update is trying to solve?

alexfornuto:
> What is the problem this documentation update is trying to solve?

@jfussion can you provide more context for this?

jfussion:
@greg-1-anderson for D7 site, it is not possible to change the temp path via settings.php and requires to modify the pressflow config as mentioned in #4972. Changing the temp dir is necessary when trying to achieve persistent temp dir https://pantheon.io/docs/tmp#persistent-temporary-path-workaround

source/content/tmp.md
```diff
@@ -16,18 +17,31 @@ export env=dev
...
+  $pressflow = json_decode($_SERVER['PRESSFLOW_SETTINGS'], TRUE);
+
+  $pressflow['conf']['file_temporary_path'] = $pressflow['conf']['file_directory_temp'] = '/tmp';
+
+  $_SERVER['PRESSFLOW_SETTINGS'] = json_encode($pressflow);
+
 }
 ```
+
+Change `sites/default/files/private/tmp` to your desired temporary file path.
```
greg-1-anderson, line 41: I didn't notice the recommendation here to put the private files directory in Valhalla; note that the code snippet above is inconsistent with this advice, recommending `/tmp` instead.

alexfornuto:
@jfussion does setting the temp directory for the specific module work for the instances that spawned this request?

@greg-1-anderson what do you think about the [current first section](https://pantheon.io/docs/tmp#default-temporary-path) of this doc? It sounds like Joseph is saying in the issue that it doesn't work, and you're saying you shouldn't change it anyway.

jfussion:
@alexfornuto yes, tmp dir for specific module/plugin works but there are some modules and plugins that doesn't have any config for tmp dir and therefore lots they end up changing the global tmp dir.

> The persistent temporary path workaround recommends setting the temporary directory for just one plugin / module, using a variable specific to that plugin or module. I think it's not a good practice to put all temporary files in valhalla. I am -1 on recommending this in documentation. I am also -1 on recommending that folks rewrite Pressflow Settings.
> 
> If you wish to move forward with this PR, please get approval from product. I recommend closing without merge.

This completely make sense and if we go for this, we may just remove the first section, Default Temporary Path, while correct, it gives the impression to the customer that it is ok to have the tmp dir in valhalla.

alexfornuto:
OK. Sounds like we're in agreement on closing this, and possible different changes to this page. Regarding the latter, I'd like to see a new issue and/or PR for that, as this thread is a bit too convoluted at this point to keep track of what anyone is actually suggesting.

Issue #5748, opened 2020-05-28
Needs COE update - Symlinks and Assumed Write Access Doc Update 


joshualieb:
Re: [Symlinks and Assumed Write Access](https://pantheon.io/docs/symlinks-assumed-write-access)

Priority: Medium

## Issue Description:
Some details need to be updated with COE rolling out, specifically around the 'mv ./wp-content/path/plugin-expects-write-to ~/backups/' instructions. 

See slack/archives/C012XNFB35Y/p1590691663064400 for context.

## Suggested Resolution

* Update the ~/backups instructions

PR #5808, opened 2020-06-18
adjust symlinked backup to writable dir

EdwardAngert:

Closes: #5748 

## Summary

**[Symlinks and Assumed Write Access](https://pantheon.io/docs/symlinks-assumed-write-access)** - Change `~/backups` to a writable directory, and add note about impersistent directories. [PR](#)

## Effect

- change ~/backups to writable dir in symlinks doc
- add note about directory bindings

## Remaining Work
(Remove if not needed)

- [ ] copy edit

## Post Launch

- [ ] Update Status Report

### DIFF
--- a/source/content/symlinks-assumed-write-access.md
+++ b/source/content/symlinks-assumed-write-access.md
@@ -43,10 +43,10 @@ The following is for Mac and Linux only. Windows users may refer to Microsoft do
 1. Move the directory you want to replace with a symlink. This serves to both back up any data that may otherwise be lost, and to prevent the symlink from being nested inside the existing directory:
 
     ```bash{promptUser: user}
-    mv ./wp-content/path/plugin-expects-write-to ~/backups/
+    mv ./wp-content/path/plugin-expects-write-to $_ENV['HOME']/tmp/backups
     ```
 
-    The command above moves the directory to a folder named `backups` in your home directory, `~/`. Replace this with your preferred backup location.
+    The command above moves the directory to the environment's home directory `/tmp/backups`. Replace this with your preferred backup location. Keep in mind that due to Pantheon's [container infrastructure](/application-containers), bindings may change and the only persistent directories that can be writen to are `/files` and `/code`.
 
 1. `cd` to the location where you want to place the symlink. The symlink command (`ln`) is sensitive to the **working directory**, the folder your command line prompt is currently in. Working from the location of the symlink allows for correct relative paths.
 
@@ -80,7 +80,7 @@ The following is for Mac and Linux only. Windows users may refer to Microsoft do
     ```
 
  Your commit can be seen in the Dev environment's commit history. Once this commit is synced to all environments, the plugin will successfully write files within any environment, even when the Dev environment's connection mode is set to Git.
- 
+
  You should not see the newly created files in the Dashboard as "ready to commit," as files are not version controlled. Only the symlink to the new path is in the codebase.
 
   <Alert title="Note" type="info">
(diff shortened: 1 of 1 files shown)


source/content/symlinks-assumed-write-access.md
```diff
@@ -43,10 +43,10 @@ The following is for Mac and Linux only. Windows users may refer to Microsoft do
 1. Move the directory you want to replace with a symlink. This serves to both back up any data that may otherwise be lost, and to prevent the symlink from being nested inside the existing directory:
 
     ```bash{promptUser: user}
-    mv ./wp-content/path/plugin-expects-write-to ~/backups/
+    mv ./wp-content/path/plugin-expects-write-to $_ENV['HOME']/tmp/backups
```
alexfornuto, line 46: How are these two versions more or less writable than each other?

alexfornuto: FYI This is the reader's local computer, so I'm not sure what @joshualieb was getting at tin the original issue.

source/content/symlinks-assumed-write-access.md
```diff
@@ -43,10 +43,10 @@ The following is for Mac and Linux only. Windows users may refer to Microsoft do
...
     ```
 
-    The command above moves the directory to a folder named `backups` in your home directory, `~/`. Replace this with your preferred backup location.
+    The command above moves the directory to the environment's home directory `/tmp/backups`. Replace this with your preferred backup location. Keep in mind that due to Pantheon's [container infrastructure](/application-containers), bindings may change and the only persistent directories that can be writen to are `/files` and `/code`.
```
alexfornuto, line 49: Again, this is a local backup, not one on the platform.

EdwardAngert: I'm reexamining what you're pointing out in the first comment, to find a way to rework that to reflect a method that works locally and will continue working once the code is pushed.
For this one though, I think the statement is true- that is what the command above is doing, and if their `backups` directory of choice is within on directory that doesn't persist, it's worth noting (and I think was the source of the issue)

alexfornuto: @EdwardAngert disagree. See [L35-l40](https://github.com/pantheon-systems/documentation/pull/5808/files#diff-f31d3c5f73c5c66b40d1eb11b3624597R35-R40), and note that this whole section is scoped to a local clone of the code base on the user's computer. So saying "the environment's home directory" is false. Saying `/tmp/backups` is wrong for all cases that aren't on the platform, unless the reader has a highly unusual desktop configuration where their home directory is set to `/` (and if that were the case, they'd have bigger problems). 

While the information on persistent directories in the platform is correct, it's irrelevant to the task at hand, which is making a _local_ backup of the content being replaced in code by a symlink.

source/content/symlinks-assumed-write-access.md
```diff
@@ -43,12 +43,12 @@ The following is for Mac and Linux only. Windows users may refer to Microsoft do
 1. Move the directory you want to replace with a symlink. This serves to both back up any data that may otherwise be lost, and to prevent the symlink from being nested inside the existing directory:
 
     ```bash{promptUser: user}
-    mv ./wp-content/path/plugin-expects-write-to ~/backups/
+    mv ./wp-content/path/plugin-expects-write-to ~/backups
     ```
 
-    The command above moves the directory to a folder named `backups` in your home directory, `~/`. Replace this with your preferred backup location.
+    The command above moves the directory to a local `backups` directory in your home folder. Replace this with your preferred backup location. Note that this backup is now outside and separate from your site's codebase, and is only a safety measure to prevent data loss. Once you've confirmed that the symlink works across all environments and no data has been lost, you can remove this backup.
```
EdwardAngert, line 49: this makes more sense
let's do it

PR #6090, opened 2020-10-29
Updates wp-config-php.md

clobee:
Re: [Doc wp-config.php](https://pantheon.io/docs/wp-config-php)

Priority: Low

## Issue Description
The document currently has a PHP code example missing a closing `)`

## Suggested Resolution
The fix is to add `)` at the example


alexfornuto (approved):
I'm no PHP expert, but I concur that closing parens are generally considered good and right, by both people and the PHP binary.

Thanks for the patch @clobee!

Issue #6117, opened 2020-11-13
Convert an existing Drupal 8 site to use Integrated Composer


greg-1-anderson:
Re: Converting an existing Drupal 8 site to use Integrated Composer

Priority: Medium

## Issue Description

Existing sites that wish to upgrade to Drupal 9 and convert to Integrated Composer, or just convert to Integrated Composer and defer the Drupal 9 upgrade will need to go through some manual steps. This conversion process may be automated in the future, but initially we will need to document the manual process.

## Suggested Resolution

The steps outlined below should for the most part work fairly well. Some small adjustments to the steps may still be needed; validation is required.

FYI: In the instructions below, note that `git checkout <branch> <file|dir>` copies a file or directory from another branch to your current branch without changing branches. This is a useful technique when migrating or converting code on a new branch.

- Add a new remote for the ic upstream and put it on a new branch, “comoserify”

```
git remote add ic git@github.com:pantheon-upstreams/drupal-project.git
git fetch ic
git checkout -b composerify ic/master
```

- Copy your configuration from the default branch

```
git checkout master sites/default/config
git mv sites/default/config/* config
git commit -m 'Pull in configuration from default branch'

```

- Compare your pantheon.yml with the new pantheon.upstream.yml and check for conflicts

```
git diff master:pantheon.yml pantheon.upstream.yml
```

- If you need your pantheon.yml, copy it over (be sure to edit if needed):

```
git checkout master pantheon.yml
# edit as needed
git add pantheon.yml
git commit -m 'Copy my pantheon.yml'
```

- OPTIONAL: Return to Drupal 8 if you’re not ready to update to Drupal 9 yet. (Advisable: test the Composer conversion before testing the Drupal 9 upgrade)

```
composer require drupal/core-recommended:^8
git add composer.*
git commit -m 'Revert to Drupal 8'
```

- Survey your enabled modules:
```
terminus drush site.dev -- pm:projectinfo --fields=name,version --format=table
```
- For each module in the list, use `composer require` to add it to your new site

  - For entries similar to “Address (address) 8.x-1.8”: `composer require drupal/address:^1.8`
  - For entries similar to “Devel (devel) 4.0.1”: `composer require drupal/devel:^4.0.1`
  - Add a ^ in front of the version to indicate “this version or any later version”
  - You may put all of the modules on a single line, if you prefer. This is faster.
  - Skip the “drupal” line. Never install drupal/drupal via Composer.
[...]

PR #6145, opened 2020-11-30
Convert a Drupal 8 site to a Drupal 9 site managed with Integrated Composer

EdwardAngert:

Closes: #6117

## Summary

**[Convert a Standard Drupal 8 Site to a Composer Managed Site](https://pantheon.io/docs/guides/composer-convert)** - Convert a Drupal 8 site to a Drupal 9 site managed with Integrated Composer. This process converts the existing D8 site to a Composer-managed site, creates a new D9 site, and moves the existing D8 to the D9 site.

## Remaining Work
(Remove if not needed)

- [x] runthrough
- [x] tech edit
- [ ] copy edit

## Post Launch


pantheondocs:
Lighthouse Audit Results:

| Description | [Live](https://pantheon.io/docs) | [pr-6145](https://pr-6145--pantheon-docs.my.pantheonfrontend.website/docs) | Difference |
| --- | --- | --- | --- |
| Performance Score | `59` | `54` | `-5` |
| Best Practice Score | `86` | `86` | `0` |
| Accessibility Score | `88` | `88` | `0` |

Lighthouse Audit Reports:

- [`pr-6145` HTML Report](https://circleci.com/gh/pantheon-systems/documentation/25023/artifacts/0//lighthouse_data/lighthouse-audit-dev.html)
- [`Pantheon Docs` HTML Report](https://circleci.com/gh/pantheon-systems/documentation/25023/artifacts/0//lighthouse_data/lighthouse-audit-reference.html)


[Preview Environment](https://pr-6145--pantheon-docs.my.pantheonfrontend.website/docs)

source/content/guides/composer-convert.md
```diff
@@ -1,26 +1,21 @@
 ---
 title: Convert a Standard Drupal 8 Site to a Composer Managed Site
-description: Drupal 8 sites often require the usage of Composer to manage site dependencies. The need to begin using Composer for a site build can often surface after a site is in development, necessitating a divergence from the Pantheon managed upstream.
+description: Use Composer to manage Drupal 8 sites often require the use of Composer to manage site dependencies. The need to begin using Composer for a site build can often surface after a site is in development, necessitating a divergence from the Pantheon managed Upstream.
```
mbaynton, line 3: Since e.g. a Drupal 8 to 9 upgrade wouldn't necessitate a divergence from our D9 managed upstream
```suggestion
description: Drupal 8 and 9 sites often require the use of Composer to manage site dependencies. The need to begin using Composer for a site build can often surface after a site is in development, necessitating a conversion of the site's codebase.
```

source/content/guides/composer-convert.md
```diff
@@ -1,26 +1,21 @@
...
 cms: "Drupal 8"
 categories: [develop]
 tags: [composer, site, workflow]
-contributors: [dustinleblanc]
+contributors: [dustinleblanc, greg-1-anderson]
+reviewed: "2020-12-01"
 ---
 
-<Alert title="Note" type="info">
+Drupal 8 sites often require the use of Composer to manage site dependencies. The need to begin using Composer for a site build can often surface after a site is in development, necessitating a divergence from the Pantheon managed Upstream.
```
mbaynton, line 13:
```suggestion
Drupal sites often require the use of Composer to manage site dependencies. The need to begin using Composer for a site build can often surface after a site is in development, necessitating a conversion of the site's codebase.
```

source/content/guides/composer-convert.md
```diff
@@ -35,72 +30,82 @@ export site=yoursitename
 
 </Alert>
 
-## Checkout a New Branch
-You're about to make some massive changes to the codebase. We recommend you to do this work on a new branch, as it might take you some time to complete and rolling back can be complicated:
+## Steps we'll take
+
+1. Convert to Composer
```
mbaynton, line 35: Just some extra context about what this guide applies to
```suggestion
1. Convert a Drupal 8 (drops-8) site to Composer
```

source/content/guides/composer-convert.md
```diff
@@ -35,72 +30,82 @@ export site=yoursitename
...
 
-  You can replace `composify` with a branch name of your choosing, but all following steps assume this name.
-
-## Setup a Multidev (optional)
-If your Pantheon account has access to [multidev](/multidev), create a multidev to push your new code to:
-
-```bash{promptUser:user}
-git push origin composify && terminus env:create $site.dev composify
-```
+  You can replace `composerify` with another branch name. If you do, remember to adjust the other examples in this doc to match.
```
mbaynton, line 55:
```suggestion
  This upstream is a useful starting point for converting Drupal 8 sites to Composer, but it is also a Drupal 9 upstream. The only changes for Drupal 8 are a few version numbers, which we'll edit later.

  You can replace `composerify` with another branch name. If you do, remember to adjust the other examples in this doc to match.
```

source/content/guides/composer-convert.md
```diff
@@ -35,72 +30,82 @@ export site=yoursitename
...
 
-This will setup the multidev environment to receive and demo our changed code.
+1. Copy your configuration from the default branch:
 
-## Create a New Composer Project
+  ```bash{promptUser:user}
+  git checkout master sites/default/config
+  git mv sites/default/config/* config
```
mbaynton, line 61:
```suggestion
  git mv sites/default/config/* config
  git rm -f sites/default/config/.htaccess
```
The `git mv` glob doesn't move dotfiles, but there's already a config/.htaccess. We end up with an empty `sites/default/config` containing only a .htaccess which I think is unnecessary

source/content/guides/composer-convert.md
```diff
@@ -35,72 +30,82 @@ export site=yoursitename
...
+  git commit -m "Pull in configuration from default branch"
+  ```
 
-1. In your local terminal, from the repository root of your Pantheon site, move a directory up:
+1. Compare your `pantheon.yml` with the new `pantheon.upstream.yml` and check for conflicts:
```
mbaynton, line 65: I didn't understand at first what this step was about, this context about what the goal is could help people to successfully interpret relevant things in the diff they get
```suggestion
1. Check for `pantheon.yml` settings you need to preserve by comparing your old codebase's `pantheon.yml` to the new `pantheon.upstream.yml`:
```

source/content/guides/composer-convert.md
```diff
@@ -35,72 +30,82 @@ export site=yoursitename
...
 
   ```bash{promptUser:user}
-  cd ..
+  git diff master:pantheon.yml pantheon.upstream.yml
   ```
 
-2. Use Composer to create a new project, using the [Pantheon Drupal 8 Composer](https://github.com/pantheon-systems/example-drops-8-composer) repository:
+1. If you need your `pantheon.yml`, copy it over (be sure to edit if needed):
```
mbaynton, line 71:
```suggestion
1. If there are lines from `pantheon.yml` (shown with a `-` in the diff output), consider copying over your old pantheon.yml to preserve these settings:
```

source/content/guides/composer-convert.md
```diff
@@ -35,72 +30,82 @@ export site=yoursitename
...
 
-    ```bash{promptUser:user}
-    composer create-project pantheon-systems/example-drops-8-composer $site-composer
-    cd $site-composer
-    ```
+  ```bash{promptUser:user}
+  git checkout master pantheon.yml
+  git add pantheon.yml
+  git commit -m 'Copy my pantheon.yml'
+  ```
```
mbaynton, line 78:
```suggestion
Note that a few settings are required to be set as they are in the `pantheon.upstream.yml` for sites using Integrated Composer. These are:
* `web_docroot`
* `build_step`
```
expounding a little on the prior "be sure to edit if needed"

source/content/guides/composer-convert.md
```diff
@@ -35,72 +30,82 @@ export site=yoursitename
...
 
-This will create a new directory based on the example project [pantheon-systems/example-drops-8-composer](https://github.com/pantheon-systems/example-drops-8-composer) in the `$site-composer` directory.
+### (Optional) Return to Drupal 8
 
-### Copy pantheon.upstream.yml
-Since the drops-8 upstream has a `pantheon.upstream.yml` and the example-drops-8-composer upstream does not, copy over our old file for the platform to properly load the site. From the `$site-composer` directory, run:
+If you’re not ready to update to Drupal 9 yet. (Advisable: test the Composer conversion before testing the Drupal 9 upgrade)
+Note that you might want to hand-edit your composer.json and remove the drupal/core entry altogether. The empty drupal/core entry from the steps above will cause drupal/core to be erased and redownloaded frequently
```
mbaynton, line 82: This sentence confuses me. I don't see `drupal/core` included in the root or upstream-configuration's composer.json, and I am wondering if there's more information available about the problem that was experienced. Composer shouldn't reinstall a dependency unless a newer release is available.

source/content/guides/composer-convert.md
```diff
@@ -35,72 +30,82 @@ export site=yoursitename
...
 
 ```bash{promptUser:user}
-cp ../$site/pantheon.upstream.yml .
+cd upstream-configuration/
+composer remove zaporylie/composer-drupal-optimizations --no-update
```
mbaynton, line 86:
```suggestion
```
No longer needs to be removed because a Composer-2 compatible release was created

source/content/guides/composer-convert.md
```diff
@@ -35,72 +30,82 @@ export site=yoursitename
...
+composer require drupal/core-recommended:^8 --no-update
+cd -
+composer config --json "extra.patches.drupal/core" '{}'
```
mbaynton, line 89:
```suggestion
```
No longer needed because the configuration this targets in the upstream was removed in https://github.com/pantheon-upstreams/drupal-project/commit/5ebb134bf1024320b805f4ac2a10d4a7e4908211

source/content/guides/composer-convert.md
```diff
@@ -147,13 +156,22 @@ cp -r /themes/custom/great_theme ../$site-composer/web/themes/custom
...
 Your existing site may have customizations to `settings.php` or any other config files. Review these carefully and extract relevant changes from these files to copy over. Always review any file paths referenced in the code, as these paths may change in the transition to Composer.
 
-It is not wise to completely overwrite the  `settings.php` file with the old one, as there are customizations for moving the configuration directory you don't want to overwrite, as well as platform specific customizations.
+It is not wise to completely overwrite the `settings.php` file with the old one, as there are customizations for moving the configuration directory you don't want to overwrite, as well as platform specific customizations.
+
+```bash{promptUser:user}
+git checkout master sites/default/settings.php
+mv -f sites/default/settings.php web/sites/default/
+git rm -rf sites
+git diff
```
mbaynton, line 168: The irony here is that the below commands do what the above suggestion says is not wise. Here's an alternative incantation that's more in keeping with "extract relevant changes from these files"
```suggestion
# Ensure working tree is clean
git status
git checkout master
git checkout -b old-settings-php
git mv sites/default/settings.php web/sites/default
git commit -m 'Relocate settings.php'
git checkout composerify
# Interactively extract relevant hunks from old settings.php
git checkout -p old-settings-php web/sites/default/settings.php
```

source/content/guides/composer-convert.md
```diff
@@ -167,14 +185,44 @@ If you are using an exported config, you will need to move the configuration fil
...
+This will create a new directory based on the example project [pantheon-systems/example-drops-8-composer](https://github.com/pantheon-systems/example-drops-8-composer) in the `$site-composer` directory.
+
+
 ## Update to Latest Drupal Core
+
 The Pantheon Drupal 8 Composer repository may not depend on the latest secure release of Drupal Core. To keep your site safe, it's a good idea to pull the latest version in now -- before moving on -- by running:
 
 ```bash{promptUser:user}
 composer update drupal/core --with-dependencies
 ```
```
mbaynton, line 222: These instructions are out of date -- suggest deleting this section.

source/content/guides/composer-convert.md
```diff
@@ -167,14 +185,44 @@ If you are using an exported config, you will need to move the configuration fil
...
 
 ## Prepare to Deploy
+
 At this point, your new project directory should contain all of the unique code from your existing Drupal 8 site, plus all of the code required to make a Composer driven project work. Since Pantheon requires all runtime code to be present when deploying to the platform, if no CI solution is a part of your workflow, you must now modify the project to be deployable straight to Pantheon.
```
mbaynton, line 226: Heads up to update this part about if no CI solution is available and using Build Tools. I assume we want to direct folks to Integrated Composer as the default now.

source/content/guides/composer-convert.md
```diff
@@ -201,20 +250,23 @@ git commit -m "Convert to Composer based install"
...
 
 ## Change Upstreams
+
 Your Pantheon site is no longer compatible with traditional upstream updates. Avoid confusion by moving your site to an empty upstream:
 
 ```bash{promptUser:user}
 terminus site:upstream:set $site empty
 ```
 
 ## Ongoing Core Updates
```
mbaynton, line 268: Heads up to update this part to recommend use of the dashboard's check for updates / apply updates functionality when using Integrated Composer

source/content/guides/composer-convert.md
```diff
@@ -35,72 +30,82 @@ export site=yoursitename
...
+Note that you might want to hand-edit your composer.json and remove the drupal/core entry altogether. The empty drupal/core entry from the steps above will cause drupal/core to be erased and redownloaded frequently
 
 ```bash{promptUser:user}
-cp ../$site/pantheon.upstream.yml .
+cd upstream-configuration/
+composer remove zaporylie/composer-drupal-optimizations --no-update
+composer require drupal/core-recommended:^8 --no-update
+cd -
+composer config --json "extra.patches.drupal/core" '{}'
+composer update
```
mbaynton, line 90:
```suggestion
# Remove new MariaDB directive from pantheon.upstream.yml
sed -i 's/^\s*version: 10.[0-9]*$/  {}/' pantheon.upstream.yml
composer update
```
There's a database version directive in the `pantheon.upstream.yml` that, at present, must be removed for any existing site regardless of Drupal version. I'm including it in this "Optional return to Drupal 8" section though to future-proof this documentation.

source/content/guides/composer-convert.md
```diff
@@ -201,20 +250,23 @@ git commit -m "Convert to Composer based install"
...
 
-If your plan does not include multidev, you will have to merge to master before deploying, then follow the rest of the steps above. If you have a local development solution, consider testing your `composify` branch locally before merging.
+If your plan does not include Multidev, you will have to merge to master before deploying, then follow the rest of the steps above. If you have a local development solution, consider testing your `composerify` branch locally before merging.
 
 ## Change Upstreams
+
 Your Pantheon site is no longer compatible with traditional upstream updates. Avoid confusion by moving your site to an empty upstream:
 
 ```bash{promptUser:user}
 terminus site:upstream:set $site empty
```
mbaynton, line 265: I think we can now support -- and therefore want -- an end state where the site *is* compatible with traditional upstream updates. Therefore we want to point sites -- both Drupal 8 and 0 -- at the `drupal9` official upstream. That way these sites will benefit from receiving all updates through the normal flow.
```suggestion
terminus site:upstream:set $site drupal9
```

mbaynton:
Related to this comment https://github.com/pantheon-systems/documentation/pull/6145/files#r534661469, I expect that we want to at some point in these instructions have the user force-push the new history based on `drupal-project` to `master`. I did this as the only not-yet-documented step to convert a vanilla drops-8 site to an IC-powered Drupal 8 site that can receive updates through the dashboard at https://admin.dashboard.pantheon.io/sites/4ebaba3f-7fba-49b2-9858-bcbfd0d86d4a.

If that's added, then changing the upstream to drupal9 makes sense IMO; if we don't want to do a force-push to master then nevermind. :)

source/content/guides/composer-convert.md
EdwardAngert: @greg-1-anderson had this in the issue (#6117), and I copied it over - I haven't encountered it either, but I'm just playing with the Umami D8 sample site in my testing.

greg-1-anderson: Remove that part -- I was talking about the PATCH to drupal/core, which is now gone. Sorry for being unclear in my haste.

greg-1-anderson:
It was my intention for the master branch to end up with the complete history of the upstream so that future dashboard updates will work. I did not get far enough to test that. I suspect a force-push is probably necessary at some point, because otherwise drupal-project commits will be a second-parent, and I don't think the dashboard traverses those.

EdwardAngert:
@greg @mbaynton @ari-gold  , my latest iteration seems to bring me closest to working (says D8.9.6), but I'm getting stuck on the last force push with a

```
remote: Contains changes to pantheon.upstream.yml which is reserved for the exclusive use of the upstream.
```

and then fails with:

```
remote: error: hook declined to update refs/heads/master
To ssh://codeserver.dev.redacted.drush.in:2222/~/repository.git
 ! [remote rejected]     master -> master (hook declined)
error: failed to push some refs to 'ssh://codeserver.dev...
```

did either of you encounter this as well, or am I missing something very obvious?
In particular, I'm looking at https://github.com/pantheon-systems/documentation/pull/6145/files#diff-de5e95860ba4c809f9e43c75a3ff0cd70ad67998ae04beec2df43d8d52958415R55 and it says that we'll adjust version numbers later, but then we never do

greg-1-anderson:
Hm pantheon.upstream.yml changes should flag a warning, not an error, so I don't know why the platform is rejecting that push. I don't think I got to the point of a force push; I think I was working exclusively in a multidev.

mbaynton:
@EdwardAngert I haven't encountered that, no. Can you paste the complete output from the remote in one block? Usually there should be another reason besides the note of the `pantheon.upstream.yml` modification which prompts the rejection.

source/content/guides/composer-convert.md
```diff
@@ -25,34 +36,70 @@ export site=yoursitename
...
+
+### (Optional) Return to Drupal 8
+
+If you’re not ready to update to Drupal 9 yet. (Advisable: test the Composer conversion before testing the Drupal 9 upgrade)
+Note that you might want to manually edit `composer.json` and remove the `drupal/core` entry if it exists. The empty drupal/core entry from the steps above will cause drupal/core to be erased and redownloaded frequently
 
 ```bash{promptUser:user}
-git push origin composify && terminus env:create $site.dev composify
+cd upstream-configuration/
+composer require drupal/core-recommended:^8 --no-update
```
mbaynton, line 94: Note 1: this will be inaccurate once https://github.com/pantheon-systems/drupal-project/pull/6 is released

Note 2: due to a new development / release process we're putting into effect, merging that PR is not the same as it being released. I'm making a note to ping @EdwardAngert when it's actually released.

EdwardAngert:
updated with a few tweaks that worked for my tests (seems my previous issue was from either mixing steps in testing or from not adding all the changed files to be committed)

moving to Tech Review, ~but~ and since this is still being actively worked on Product-side and getting better all the time, there may be things that we should update before this goes live ([this comment in particular](https://github.com/pantheon-systems/documentation/pull/6145#discussion_r541211672) from @mbaynton )

source/content/guides/composer-convert.md
```diff
@@ -25,34 +36,69 @@ export site=yoursitename
...
+     git commit -m 'Copy my pantheon.yml'
+     ```
+
+   Note that a few settings must be left as they are in `pantheon.upstream.yml` for sites using Integrated Composer. These are:
 
-If your Pantheon account has access to [Multidev](/multidev), create a Mmultidev to push your new code to:
+     * `web_docroot`
+     * `build_step`
+
+### (Optional) Return to Drupal 8
```
EdwardAngert, line 87: I've tried to move this to different parts of the doc, but none really feel right, and I'm not sure it belongs here either. Feel free to find a better place for this.

source/content/guides/composer-convert.md
```diff
@@ -25,34 +36,70 @@ export site=yoursitename
 
 </Alert>
 
-## Checkout a New Branch
+## Add the Pantheon Integrated Composer Upstream in a New Local Branch
 
-You're about to make some massive changes to the codebase. We recommend you to do this work on a new branch, as it might take you some time to complete and rolling back can be complicated:
+You're about to make massive changes to the codebase. We recommend you to do this work on a new branch, as it might take you some time to complete and rolling back changes can be complicated:
```
greg-1-anderson, line 41: This comment is still true; however, the new instructions are different from the old in now, not only are we making massive changes to the codebase, we are also going to switch our main branch to a completely different git history with no commits in common with the existing repository.

source/content/guides/composer-convert.md
```diff
@@ -1,18 +1,29 @@
...
+Existing sites that wish to upgrade from Drupal 8 to Drupal 9 and convert to Integrated Composer, or just convert to Integrated Composer and defer the Drupal 9 upgrade will need to go through some manual steps.
+
 ## Before You Begin
 
 - Review our documentation on [Git](/git), [Composer](/composer), and [Terminus](/terminus), and have them installed and configured on your local computer.
-- [Clone](/git#clone-your-site-codebase) your _current_ Pantheon site repository to a working directory on your local computer.
+   - Mac users can use [Homebrew](https://brew.sh/) to install both Git and Composer, along with their required dependencies:
+
+     ```bash{promptUser:user}
+     brew install git composer
```
greg-1-anderson, line 23: It might also be a good idea to advise users that they should be on Composer 2. If we do this, then we should probably also link to a document with guidance. If we don't do this, they may run into problems if they push a site that only works with Composer 1 up to the platform, which uses Composer 2.

source/content/guides/composer-convert.md
```diff
@@ -25,34 +36,69 @@ export site=yoursitename
...
+You're about to make massive changes to the codebase. We recommend you to do this work on a new branch, as it might take you some time to complete and rolling back changes can be complicated:
 
 1. In your local terminal, change directories to your site project. For example, if you keep your projects in a folder called `projects` in the home directory:
 
   ```bash{promptUser:user}
   cd ~/projects/$site/
   ```
 
-2. Create the new branch:
+1. Add the Pantheon Drupal Upstream as a new remote, fetch the `ic` branch, and checkout to a new local branch based on it called `composerify`:
```
greg-1-anderson, line 49: `ic` is a remote, not a branch. The commands below add a new remote called ic, fetches its master branch and then creates a new branch called "composerify" from the master branch of the ic remote.

Note that the resulting branch "composerify" shares a commit history with the drupal-project upstream, and has no commits in common with the site's existing master branch.

source/content/guides/composer-convert.md
```diff
@@ -25,34 +36,69 @@ export site=yoursitename
...
+  git remote add ic git@github.com:pantheon-upstreams/drupal-project.git && git fetch ic && git checkout -b composerify ic/master
   ```
 
-  You can replace `composify` with a branch name of your choosing, but all following steps assume this name.
+  This upstream is a useful starting point for converting Drupal 8 sites to Composer, but it is a Drupal 9 upstream. To use it for Drupal 8, we will adjust the version numbers in a later step.
 
-## Set up a Multidev (Optional)
+  You can replace `composerify` with another branch name. If you do, remember to adjust the other examples in this doc to match.
+
+1. Copy your configuration from the default branch:
```
greg-1-anderson, line 59: While it is recommended to store configuration in the database, some sites might not do this. It's not an error if this step does not bring in any files.

source/content/guides/composer-convert.md
```diff
@@ -25,34 +36,69 @@ export site=yoursitename
...
+
+  ```bash{promptUser:user}
+  git checkout master sites/default/config
+  git mv sites/default/config/* config
+  git rm -f sites/default/config/.htaccess
+  git commit -m "Pull in configuration from default branch"
+  ```
+
+1. Check for `pantheon.yml` settings you need to preserve by comparing your old codebase's `pantheon.yml` to the new `pantheon.upstream.yml`:
```
greg-1-anderson, line 68: Users should expect their settings to change for `database` and `build_step` and `web_docroot` but they might want to preserve their current settings for `php_version`, `drush_version` & similar

source/content/guides/composer-convert.md
```diff
@@ -25,34 +36,69 @@ export site=yoursitename
...
+     git checkout master pantheon.yml
+     git add pantheon.yml
+     git commit -m 'Copy my pantheon.yml'
+     ```
+
+   Note that a few settings must be left as they are in `pantheon.upstream.yml` for sites using Integrated Composer. These are:
 
-If your Pantheon account has access to [Multidev](/multidev), create a Mmultidev to push your new code to:
+     * `web_docroot`
+     * `build_step`
```
greg-1-anderson, line 85: Add `database` to this list. Maybe mention that it's best to just remove the setting from pantheon.yml if they want to use the same value as pantheon.upstream.yml.

source/content/guides/composer-convert.md
```diff
@@ -25,34 +36,69 @@ export site=yoursitename
...
+# Remove new MariaDB directive from pantheon.upstream.yml
+sed -i 's/^\s*version: 10.[0-9]*$/  {}/' pantheon.upstream.yml
+composer update
+git add .
+git commit -m 'Revert to Drupal 8'
 ```
 
-This will set up the Multidev environment to receive and demo our changed code.
-
 ## Create a New Composer Project
```
greg-1-anderson, line 102: This section should be removed (everything down to "## Add in the Custom and Contrib Code Needed to Run Your Site"), as this is part of the old documentation. It is not part of the new procedure.

source/content/guides/composer-convert.md
```diff
@@ -92,18 +128,15 @@ A Composer-managed site should be able to include all custom code via Composer.
 
 When reviewing your site, take stock of exactly what versions of modules you depend on. One way to do this is to use a command like the following from within a contributed modules folder (e.g. `/modules`, etc):
 
-```bash{outputLines:2-4}
-find . -maxdepth 2 -name '*.info.yml' \
-  -exec basename '{}' '.info.yml' ';' \
-  -exec grep -e '^version' '{}' ';' \
-  -exec echo ';'
+```bash{promptUser:user}
+terminus drush $site.dev -- pm:projectinfo --fields=name,version --format=table
```
greg-1-anderson, line 132: Note that `pm:projectinfo` was removed in Drush 9, so the users will need to revert back to Drush 8 if they are using a newer Drush. Usually folks stay on Drush 8 until they convert to Composer, so this might not be a common problem.

source/content/guides/composer-convert.md
```diff
@@ -92,18 +128,15 @@ A Composer-managed site should be able to include all custom code via Composer.
...
 ```
 
 This will list each module followed by the version of that module that is installed.
 
 You can add these modules to your new codebase using Composer by running the following for each module in the `$site-composer` directory:
 
-```{promptUser:user}
+```bash{promptUser:user}
 composer require drupal/MODULE_NAME:^VERSION
```
greg-1-anderson, line 140: It would probably be a good idea to tell folks that they should skip the ludwig module if they are using it. This module is used to integrate modules that require Composer into sites that are not composer-managed, so it might be fairly common.

greg-1-anderson: Also, note that Drush will show version strings in two different format.

If Drush says 8.x-1.8, then they should use ^1.8 (common)

If Drush says 4.0.1, then they should use ^4.0.1 (less common)

source/content/guides/composer-convert.md
```diff
@@ -92,18 +128,15 @@ A Composer-managed site should be able to include all custom code via Composer.
 
 When reviewing your site, take stock of exactly what versions of modules you depend on. One way to do this is to use a command like the following from within a contributed modules folder (e.g. `/modules`, etc):
```
greg-1-anderson, line 129: Here you can say "modules and themes", and it's fine to just say "One way to do this is to use a command like the following:" (it doesn't matter which folder you're in when you run drush pm:projectinfo)

source/content/guides/composer-convert.md
greg-1-anderson: Below, the section "#### Themes" may be removed, because it is now covered by the section above. Furthermore, the section "#### Libraries" is now largely unneeded. With Composer-managed sites, for the most part the libraries your modules will need will be included automatically by Composer. This will happen for any module that has been updated for Composer, which by now is the majority of them. There might be some modules remaining that still need external libraries, and do not use Composer to install them. These will need special handling. I'm not sure what to say in the "Libraries" section, though, since I don't have any specific examples where this is still needed.

greg-1-anderson: The section "#### Modules and Themes" should also be adjusted:

> Custom code should be checked out from the master branch, which contains the original site:
> ```
> git checkout master modules/custom
> ```
> Note that the command above does not switch branches; it copies the contents of the `modules/custom` directory from the master branch onto the current branch.
>
> Foolow suit with any other custom code you may need to carry over from other directories.

source/content/guides/composer-convert.md
```diff
@@ -150,7 +183,19 @@ Follow suit with any other custom code you need to carry over.
...
+# Ensure working tree is clean
+git status
+git checkout master
+git checkout -b old-settings-php
+git mv sites/default/settings.php web/sites/default
+git commit -m 'Relocate settings.php'
+git checkout composerify
+# Interactively extract relevant hunks from old settings.php
+git checkout -p old-settings-php web/sites/default/settings.php
+```
```
greg-1-anderson, line 198: It's kind of a lot of steps just to get your settings.php in the right place on a separate branch (you also need a `mkdir -p web/sites/default` in the instructions above, as `web` does not exist on the original branch).

How about this instead:
```
git checkout master sites/default/settings.php
diff -Nup web/sites/default/settings.php sites/default/settings.php 
# Edit settings.php as needed
rm sites/default/settings.php
```
You don't get to use `git checkout -p`, but there are significantly fewer steps.

source/content/guides/composer-convert.md
```diff
@@ -169,19 +214,31 @@ If you are using an exported config, you will need to move the configuration fil
...
 
-## Update to Latest Drupal Core
+## Set up a Multidev (Optional)
 
-The Pantheon Drupal 8 Composer repository may not depend on the latest secure release of Drupal Core. To keep your site safe, it's a good idea to pull the latest version in now -- before moving on -- by running:
+If your Pantheon account has access to [Multidev](/multidev), create a Multidev to push your new code to:
 
 ```bash{promptUser:user}
-composer update drupal/core --with-dependencies
+git push origin composerify && terminus env:create $site.dev composerify
```
greg-1-anderson, line 222: Here's the point where the intrepid adventurer may receive the error:

> Changing database version after environment creation is not currently supported.

If this happens, then remove the `database:` section from pantheon.yml. If the user does this, then they should also downgrade to Drupal 8 and upgrade to Drupal 9 later, after the database upgrade is available.

greg-1-anderson: It would also probably be better to have the user create the 'composerify' multidev before pushing the branch. If you do it the other way, Pantheon won't necessarily pick up changes to the composer.json file.

source/content/guides/composer-convert.md
```diff
@@ -169,19 +214,31 @@ If you are using an exported config, you will need to move the configuration fil
...
 ```
 
-## Prepare to Deploy
+This will set up the Multidev environment to receive and demo our changed code.
+## Commit
```
greg-1-anderson, line 226: This section ("Commit") may be removed.

If the user has been following along with the instructions in this document, then all of the work should have been done on the "composerify" branch, and at this point everything should be committed and pushed.

source/content/guides/composer-convert.md
```diff
@@ -169,19 +214,31 @@ If you are using an exported config, you will need to move the configuration fil
...
+cd ../$site-composer
+git add .
+git commit -m "Convert to Composer based install"
+```
 
-At this point, your new project directory should contain all of the unique code from your existing Drupal 8 site, plus all of the code required to make a Composer driven project work. Since Pantheon requires all runtime code to be present when deploying to the platform, if no CI solution is a part of your workflow, you must now modify the project to be deployable straight to Pantheon.
+You should see a large number of files committed to the new branch created earlier.
 
-If you do plan on using a CI solution, refer to our [Build Tools](/guides/build-tools) guide at this point.
+## Prepare to Deploy
```
greg-1-anderson, line 239: The "Prepare to Deploy" step should no longer be necessary. There is no `prepare-for-pantheon` any longer, and `composer install` should have already ran implicitly as one of the previous steps.

source/content/guides/composer-convert.md
```diff
@@ -190,49 +247,49 @@ composer prepare-for-pantheon
...
 ```
 
-This should modify the `.gitignore` file and cleanup any errant `.git` directories in the codebase, to prepare your new code for direct deployment to Pantheon.
+This modifies the `.gitignore` file and cleans up any errant `.git` directories in the codebase to prepare your new code for direct deployment to Pantheon.
 
-## Commit
+## Deploy
 
-Commit your work to the git repo. From the `$site` directory, run the following:
+You've now committed the code to the local branch. If your site has Multidev, you can deploy that branch directly to a new Multidev and test the site in the browser. If the site doesn't load properly, clear the cache. If there are any issues, utilize your site's logs via `terminus drush $site.composerify -- wd-show` to inspect the watchdog logs, or follow the directions in our documentation on [log collection](/logs).
```
greg-1-anderson, line 254: This paragraph with `drush wd-show` could be moved up to the "Set up a Multidev" section above. The rest of this ("Deploy" and "Deploy to a Multidev") is redundant with what has come before.

source/content/guides/composer-convert.md
```diff
@@ -190,49 +247,49 @@ composer prepare-for-pantheon
...
+
+```bash{promptUser:user}
 git add .
-git commit -m "Convert to Composer based install"
+git commit -m "ran composer prepare-for-pantheon and install"
+git push origin composerify
 ```
 
-You should see a large amount of files committed to the new branch we created earlier.
+Once you have confirmed the site is working, merge `composerify` into `master`, and follow the standard workflow to QA a code change before going live.
```
greg-1-anderson, line 266: Maybe this sentence also should move up to the "Set up a Multidev" section.

source/content/guides/composer-convert.md
```diff
@@ -190,49 +247,49 @@ composer prepare-for-pantheon
...
 
-## Deploy
+### Deploy to Dev
 
-You've now committed the code to a branch. If your site has Multidev, you can deploy that branch directly to a new Multidev and test the site in the browser. If the site doesn't load properly, clear the cache. If there are any issues, utilize your site's logs via `terminus drush $site.composify -- wd-show` to inspect the watchdog logs, or follow the directions on our documentation on [log collection](/logs).
+If you have a [local development](/local-development) solution, consider testing your `composerify` branch locally before merging.
```
greg-1-anderson, line 270: Redundant with "Set up a Multidev" section (although you might add a note up there to skip that step if their plan does not support multidevs)

source/content/guides/composer-convert.md
```diff
@@ -190,49 +247,49 @@ composer prepare-for-pantheon
...
+```
 
 ## Change Upstreams
 
-Your Pantheon site is no longer compatible with traditional upstream updates. Avoid confusion by moving your site to an empty upstream:
+Set the site to use the Drupal 9 Upstream:
 
 ```bash{promptUser:user}
-terminus site:upstream:set $site empty
+terminus site:upstream:set $site drupal9
```
greg-1-anderson, line 285: It should be safe to do this even if the user pinned to Drupal 8.

mbaynton:
@greg-1-anderson @EdwardAngert I'm thinking we should not block the release of this documentation on the platform being able to upgrade a site's MariaDB version. All of the content here should be applicable to customers that only need to convert an existing D8 or D9 site to IC, but do not need to upgrade from D8 to D9, and we should provide a guide for them. Does that make sense to you folks?

greg-1-anderson:
@mbaynton Yeah we could just have a note advising folks to always roll back to Drupal 8.9.x for now instead of making that an optional step, and mention that the Drupal 9 upgrade will be available soon w/ the Mariadb upgrade.

mbaynton:
Ok, I'm starting up the automated scripting now using this doc as a guide so I will by default be re-testing its accuracy. I'll report back if I find anything, although don't feel obliged to block on that.

source/content/guides/composer-convert.md
EdwardAngert: I tried to explain this as succinctly as possible, but might need to come back with more words if it feels like it's missing the "why"

source/content/guides/composer-convert.md
```diff
@@ -25,85 +40,74 @@ export site=yoursitename
...
+     git commit -m 'Copy my pantheon.yml'
+     ```
 
-```bash{promptUser:user}
-cp ../$site/pantheon.upstream.yml .
-```
+   If you prefer to keep the values for any of the following settings from `pantheon.upstream.yml`, remove them from `pantheon.yml`:
 
-`ls` should reveal that the new code repository now has a copy of the `pantheon.upstream.yml`.
+     * `web_docroot`
```
greg-1-anderson, line 86: `web_docroot` should always be inherited from the upstream.

source/content/guides/composer-convert.md
greg-1-anderson: Using `build_step: true` is the main point of this conversion, so that one should not be overridden.

source/content/guides/composer-convert.md
```diff
@@ -25,85 +40,74 @@ export site=yoursitename
...
+     * `build_step`
+     * `database`
```
greg-1-anderson, line 88: So right now the upstream tries to upgrade to 10.4, which isn't going to work for a while. Perhaps we need to instruct folks to set the database version to 10.0 until general availability.

mbaynton: I thought our approach to this was to release the documentation for performing conversions to IC only for now, staying on the major version of Drupal you're already on. The title says "Convert to Drupal 9", but we fundamentally can't do that, so we should change the title.

I thought we had an optional step describing what to do to stay on Drupal 8 and we would essentially just make it a required step. But I do not see it in the rendered documentation. Perhaps this is what I had that post-it I had to ping you was about @EdwardAngert -- the details of how to stay on Drupal 8 did change with the new release of the upstream. I'll take a stab at those steps now.

source/content/guides/composer-convert.md
mbaynton: @EdwardAngert @greg-1-anderson I thought I'd been down this road before. This is how we would want to deal with https://github.com/pantheon-systems/documentation/pull/6145#discussion_r575312499 I believe. Removing the database version will allow us to upgrade the sites as time goes on, so it's better than including the database version and requiring 10.0.

source/content/guides/composer-convert.md
mbaynton: The comments are getting numerous and hard to follow, but it looks like I posted a recommended change for what makes sense to me at https://github.com/pantheon-systems/documentation/pull/6145#discussion_r575484057

greg-1-anderson: The goal here is to put people on the pantheon-upstreams/drupal-project upstream. We could make an 8.x branch of that upstream, or make a different upstream if we want to simplify the instructions here, but if we keep on the current course, pantheon.upstream.yml sets the database version to 10.4.  It is therefore insufficient for us to remove the database version; we must reset it to 10.0 or the user will receive an error.

Should confirm -- I have not tried this for a while. My understanding, though, is that this error will persist until Mariadb / Drupal 9 general availability.

source/content/guides/composer-convert.md
mbaynton: Doh, bad idea to modify pantheon.upstream.yml. I've tested however that if you put this into `pantheon.yml` instead, then you override `pantheon.upstream.yml`'s 10.4, and also maintain the desired effect of the database version not being strictly pinned:
```
database:
  {}
 ```
That's what I think I'll have the terminus conversion tool I'm working on do if the user says they're staying with Drupal 8.

Comment (greg-1-anderson) [reply source/content/guides/composer-convert.md]: I think that this doc is blocked on the database upgrade, so we can just leave this step out.

File: source/content/guides/composer-convert.md [hunk 40]
```diff
@@ -1,18 +1,33 @@
...
 
+Drupal 9 sites on Pantheon have Composer built-in to manage site dependencies.
+
+For a smooth upgrade experience, and to avoid potential conflicts, this doc shows how to migrate a Drupal 8 site to a freshly prepared, new Drupal 9 site.
+
+The goals of this upgrade are to remove dependencies from the old site that Composer will manage from your Git repository, and tell Composer about those dependencies in the new site instead.
+
+Please note, that since you are migrating a site through this process, the new site will no longer maintain your existing commit history.
+
 ## Before You Begin
```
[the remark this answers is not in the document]
EdwardAngert: added https://github.com/pantheon-systems/documentation/pull/6145/files#diff-de5e95860ba4c809f9e43c75a3ff0cd70ad67998ae04beec2df43d8d52958415R32

source/content/guides/composer-convert.md
EdwardAngert: removed from this list and reframed the sentence to show `database` as the only value. Is that right?

source/content/guides/composer-convert.md
EdwardAngert: offered `pm:list` and linked to drush docs

source/content/guides/composer-convert.md
EdwardAngert: can we resolve this conversation?

source/content/guides/composer-convert.md
EdwardAngert: should I just add the line `It should be safe to do this even if the site is pinned to Drupal 8.`? or is there more nuance here that I'm missing?

source/content/guides/composer-convert.md
```diff
@@ -1,156 +1,301 @@
...
-This will list each module followed by the version of that module that is installed.
+Begin by reviewing the existing site's code. Check for contributed modules in `/modules`, `/modules/contrib`, `/sites/all/modules`, and `/sites/all/modules/contrib`.
 
-You can add these modules to your new codebase using Composer by running the following for each module in the `$site-composer` directory:
+1. When reviewing the site, take stock of exactly what versions of modules and themes you depend on. One way to do this is to change to run the `pm:projectinfo` Drush command from within a contributed modules folder (e.g. `/modules`, `/themes`, `/themes/contrib`, `/sites/all/themes`, `/sites/all/themes/contrib`, etc.).
 
-```{promptUser:user}
-composer require drupal/MODULE_NAME:^VERSION
-```
+  This command works on Drush 8. If you're using Drush 9, use `pm:list` or refer to [Drush Commands](https://drushcommands.com/drush-9x/pm/pm:projectinfo/):
```
greg-1-anderson, line 182: Users with Drush 9 are probably on Composer-managed sites, and should be following a different guide. n.b. pm:list shows modules, and pm:projectinfo shows projects. Projects are collections of modules. If the user listed modules, they'd need to figure out the project name from the module name. This is usually the same for projects with one module. Projects with multiple modules often have one with the same name as the project.

source/content/guides/composer-convert.md
```diff
@@ -1,156 +1,301 @@
...
-[InvalidArgumentException]
-Could not find a version of package drupal/MODULE_NAME matching your minimum-stability (stable). Require it with an explicit version constraint allowing its desired stability.
-```
+1. You can add these modules to your new codebase using Composer by running the following for each module in the `$SITE-composer` directory:
+
+  ```bash{promptUser:user}
+  composer require drupal/MODULE_NAME:^VERSION
+  ```
+
+  Where `MODULE_NAME` is the machine name of the module in question, and `VERSION` is the version of that module the site is currently using. Composer may pull in a newer version than what you specify, depending upon what versions are available. You can read more about the caret (`^`) in the [Composer documentation](https://getcomposer.org/doc/articles/versions.md#caret-version-range-).
```
greg-1-anderson, line 196: At one point we had an explanation about versions strings.

Old-style Drupal version strings:
```
Chaos Tools (ctools)  8.x-3.4
```
Drop the `8.x-` to convert this into `^3.4`

Semver version strings:
```
Devel (devel)  4.1.1
```
Use this version directly, e.g. `^4.1.1`

source/content/guides/composer-convert.md
```diff
@@ -1,156 +1,301 @@
...
+
+   - If you get the following error, the module listed in the error (or its dependencies) does not meet compatibility requirements:
+
+   ```none
+   [InvalidArgumentException]
+   Could not find a version of package drupal/MODULE_NAME matching your minimum-stability (stable). Require it with an explicit version constraint allowing its desired stability.
+   ```
+
+   If there is no stable version you can switch to, you may need to adjust the `minimum-stability` setting of `composer.json` to a more relaxed value, such as `beta`, `alpha`, or even `dev`. You can read more about `minimum-stability` in the [Composer documentation](https://getcomposer.org/doc/04-schema.md#minimum-stability).
```
greg-1-anderson, line 205: We should not recommend lowering the minimum stability to `dev`, as this can have odd side effects. We should instead tell people to require the specific version they need via Composer, e.g.
```
composer require drupal/ctools:^3.4@dev
```
If requiring a dev version fails because the dev version of the module requires a dev version of a dependency, they should fix this by also allowlisting the dev dependency at the same time:
```
composer require drupal/some-module:^1@dev org/some-dependency:^2@dev
```

source/content/guides/composer-convert.md
```diff
@@ -1,156 +1,301 @@
...
+
+One of Pantheon's engineers got tired of doing this process by hand, so he trained some robots to identify modules and put them into `composer.json`.
+
+Robots are cool, but they're not perfect, so you should understand the goal of this process as well as the limitations of automating the process.
+
+<Accordion title="A script that can help add modules to composer.json" id="modules-script" icon="cogs">
+
+First, disclaimers:
+
+Pantheon support is not available for this script or anything that results from trying to use it.
```
greg-1-anderson, line 217: I think we need slightly higher levels of support for scripts we publish in official documentation. This statement sounds a little too strong. Do we have precedence for this (support or lack of support for scripts in documentation)?

I think we should publish the docs without the scripts first & then have a discussion about support. e.g. with nominal buy-in from Devrel we could perhaps direct folks to #composer-workflow

source/content/guides/composer-convert.md
greg-1-anderson: Actually we need one more command before the `rm`:
```
git restore --staged sites/default/settings.php 
```
I don't know if there's a better way to do this.

greg-1-anderson: OK, here's a good solution:
```
git show master:sites/default/settings.php > web/sites/default/original-settings.php
diff -Nup --ignore-all-space web/sites/default/settings.php web/sites/default/original-settings.php
# edit web/sites/default/settings.php and commit as needed
rm web/sites/default/original-settings.php
```

source/content/guides/composer-convert.md
greg-1-anderson: Regarding precedence, we have this text in Terminus plugins:

> This repository is provided without warranty or direct support. Issues and questions may be filed in GitHub but their resolution is not guaranteed.

source/content/guides/composer-convert.md
```diff
@@ -167,77 +312,54 @@ If you are using an exported config, you will need to move the configuration fil
...
-You've now committed the code to a branch. If your site has Multidev, you can deploy that branch directly to a new Multidev and test the site in the browser. If the site doesn't load properly, clear the cache. If there are any issues, utilize your site's logs via `terminus drush $site.composify -- wd-show` to inspect the watchdog logs, or follow the directions on our documentation on [log collection](/logs).
+If you have a [local development](/local-development) solution, consider testing your `composerify` branch locally before merging.
 
-Once you have confirmed the site is working, merge `composify` into `master`, and follow the standard workflow to QA a code change before going live.
+To force push the changes from the local branch to the `master` Dev branch:
 
-If your plan does not include Multidev, you will have to merge to master before deploying, then follow the rest of the steps above. If you have a local development solution, consider testing your `composify` branch locally before merging.
+```bash{promptUser:user}
+git add .
+git commit -m "ran composer prepare-for-pantheon and install"
```
greg-1-anderson, line 341: Comment does not match instructions; `composer prepare-for-pantheon` isn't part of this guide (this is from the guild on converting a site to build tools)

source/content/guides/composer-convert.md
```diff
@@ -1,156 +1,308 @@
...
+   - If there are settings from `pantheon.yml` (shown with a `-` in the diff output), consider copying over your old `pantheon.yml` to preserve these settings:
+
+     ```bash{promptUser:user}
+     git checkout master pantheon.yml
+     git add pantheon.yml
+     git commit -m 'Copy my pantheon.yml'
+     ```
 
-`ls` should reveal that the new code repository now has a copy of the `pantheon.upstream.yml`.
+   If you prefer to keep the value for `database` from `pantheon.upstream.yml`, remove it from `pantheon.yml`.
```
greg-1-anderson, line 166: :rotating_light: :rotating_light: :rotating_light:  DANGER :rotating_light: :rotating_light: :rotating_light:
If a user has an existing site that's running Mariadb 10.0 and you follow this guide and end up with a pantheon.yml or pantheon.upstream.yml with a database version of 10.4 before Drupal 9 General Availability, your site might be in danger of database corruption.

source/content/guides/composer-convert.md
greg-1-anderson: There is no point in linking to the Drush 9 pm:projectinfo entry on Drush Commands, though; if you run pm:projectinfo you just get an error message telling you to use pm:list or composer show

As of Drush 8.4.8 and Drush 9.7.3, though, you can run `pm:list --fields=*` and see all of the available modules along with the project they're part of. This field isn't included in Drush 10.4.0, but it will be in the next release (hopefully available soon)

source/content/guides/composer-convert.md
hanoii: Currently trying to this on an actual project. I set 
```yml
api_version: 1
database:
  version: 10.0
```

on my `pantheon.yml` to prevent using 10.4

but I got 

```
remote: 
remote: PANTHEON ERROR:
remote: 
remote: Changes to `pantheon.yml` detected, but there was an error while processing it:
remote: 
remote: pantheon.yml: 
remote: Validation failed with error:
remote: >   10.0 is not one of [10.4]
remote: 
remote: while validating the following value:
remote: >   10.0
remote: >   ...
remote: 
```

greg-1-anderson: @hanoii The `database:` directive in pantheon.yml is not ready to use, and is dangerous. You must remove the directive if you do not want to use it, because at the moment the only valid value is 10.4. If your pantheon.upstream.yml has the `database:` directive, though, then you must remove it there as well; failure to do so prior to GA could be dangerous. An individual site is not supposed to modify the pantheon.upstream.yml file, though, so you will have extra work to do later if you do this. I would like to reiterate that if you make a mistake during this process, the result could be dangerous. If you think that you may have applied a `database` directive with the value 10.4 in any existing site's multidev, I would recommend deleting that multidev, and start over with a fresh one.

The recommended process to follow today is https://github.com/pantheon-systems/documentation/pull/6314. I would follow this guide to convert an existing drops-8 to integrated composer, even if your intention is to downgrade back to Drupal 8.

hanoii: @greg-1-anderson thanks! So no `database:` directive on either `pantheon.yml` and `pantheon.upstream.yml` means `10.0`? Isn't there anyway for me to override the `database` on my `pantheon.yml` with null or empty object something that would mean the same as no database? 

Anyway, I will try removing it. Do you foresee any issue moving forward that makes me reconsider doing the d8 to a new d9 site worthwhile considering at this point? I assume that  changing the upstream yaml, and then reverting that commit should allow for proper upstream merging from now on once 10.4 is GA.

greg-1-anderson: There is a comment above from mbaynton that I forgot about: it appears that this you can get 10.0 by modifying only pantheon.yml as follows:
```
database:
  {}
```
My honest advice, though, is that you should follow the instructions to convert your site to Drupal 9 by creating a new site, and then move your work back to your existing site after D9 GA.

hanoii: @greg-1-anderson ok. If I move it to a new d9 site I guess I can simply move everything there and then switch domains. Will explore this option. Thanks!

hanoii: FWIW, `database: {}` does work in that I get the push to go through and the site/code looks to have built, howeveer I connected to the mysql site and it reported to be `10.4`. The site worked, but I am not trying without any database at all.

 I did find a few things odd. The `tag pantheon_composerify_1` actually has the code from `master` (not yet composerified), but the deployed code as seen by SFTP looks the new stuff. This seems to be some odd issue with multidev and IC. I also saw the same behavior on a new d9 site I am using for testing where I created a new multidev. There it is less obvious because the tag belongs to the last commit, but there seems to be no artifacts tag on multidevs.

stovak: Here's mine from my last project:
```yaml
api_version: 1
web_docroot: true
php_version: 7.4
drush_version: 10
database:
  version: 10.4
```

stovak: FYI: Decoupled react with D9 backend: https://live-freshdrupalmi.pantheonsite.io

EdwardAngert:
that rebase did not do exactly what I was hoping it would - please hold while I try to clean up my mistakes

EdwardAngert:
sorry @stovak @hanoii @greg-1-anderson @mbaynton, in trying to rebase, I torched the thread of conversations on this page. it's still available on [/files](https://github.com/pantheon-systems/documentation/pull/6145/files), so I'm not going to try to fix this any more than I already have 😶 
I am going through the discussions and trying to incorporate all that excellent info so we can have this ready soon.
I appreciate each of you

source/content/guides/composer-convert.md
EdwardAngert: I've added this a subpoints in https://github.com/pantheon-systems/documentation/pull/6145/files#diff-de5e95860ba4c809f9e43c75a3ff0cd70ad67998ae04beec2df43d8d52958415R200-R216

source/content/guides/composer-convert.md
EdwardAngert: updated to match your latest comment, but I've kept the `git status` to catch people who're just running through steps and might have missed something staged
I can remove it if it's too in the way

greg-1-anderson: `git status` is a good idea.

EdwardAngert:
big changes in my last push
we had the instructions for the upgrade-in-place in the composer-convert doc
I mostly reverted that doc to be about converting a non-composer site to a composer-managed site, and moved much of the work in this PR to the Migrate to Drupal 9 guide

for a preview of what this doc will like like in that doc, check out the built preview: [Migrate Existing Sites
](https://pr-6145--pantheon-docs.my.pantheonfrontend.website/docs/guides/drupal-9-migration/migrate-guided-d9)
(open to a new concise name)

EdwardAngert:
moving this to copy review - but please let me merge it when it's time, just in case this looks ready to go and has all the right 👍  on it, but someone DMs me that something's not ready for it from the platform side of things.

as part of the copy review, I think we should rename from `Migrate Existing Sites` to something closer to `Upgrade a Drupal 8 Site to a Composer-managed Drupal 9 Site on Pantheon` because there's been some understandable confusion about what `migrate` means in this case

source/content/guides/composer-convert.md
```diff
@@ -3,215 +3,38 @@ title: Convert a Standard Drupal 8 Site to a Composer Managed Site
...
-[InvalidArgumentException]
-Could not find a version of package drupal/MODULE_NAME matching your minimum-stability (stable). Require it with an explicit version constraint allowing its desired stability.
-```
+<Alert title="Note" type="info">
 
-It means that one of the modules you are using (or its dependencies) are not stable. If there is not a stable version you can switch to, you may need to adjust the `minimum-stability` setting of `composer.json` to a more relaxed value, such as `beta`, `alpha`, or even `dev`. You can read more about `minimum-stability` [in the documentation for Composer](https://getcomposer.org/doc/04-schema.md#minimum-stability) itself.
+To maintain best practice and to avoid difficult, time-consuming repairs to the site, this doc is written for users with access to Pantheon's [Multidev](/multidev) feature.
 
-#### Themes
+Pantheon support is not available to users who avoid the Multidev steps.
```
greg-1-anderson, line 25: We can direct people to the "Manually Migrate" page (or the renamed version thereof) to follow the "create a new site" path.

source/partials/drupal-8-convert-to-composer.md
```diff
@@ -0,0 +1,255 @@
...
+
+This process involves significant changes to the codebase. We recommend you to do this work on a new branch, as it might take you some time to complete and rolling back changes can be complicated:
+
+1. In your local terminal, change directories to the site project. For example, if you keep your projects in a folder called `projects` in the home directory:
+
+  ```bash{promptUser:user}
+  cd ~/projects/$SITE/
+  ```
+
+1. Add the Pantheon Drupal Upstream as a new remote called `ic`, fetch the `ic` branch, and checkout to a new local branch based on it called `composerify`:
```
greg-1-anderson, line 28: "fetch the `ic` upstream,"

source/content/guides/drupal-9-migration/03-migrate-guided-d9.md
```diff
@@ -1,34 +1,73 @@
...
+<Partial file="drupal-9/prepare-local-environment.md" />
+
+<Partial file="drupal-8-convert-to-composer.md" />
+
+## Change Upstreams
+
+Set the site to use the Drupal 9 Upstream:
+
+```bash{promptUser:user}
+terminus site:upstream:set $site drupal9
```
greg-1-anderson, line 50: We should probably wait and do this at the very end. If it takes the user a long time to convert, and another Drupal 8 upgrade comes out, then they'll want to still be on drops-8 so that they can apply the upstream update to their dev site.

source/partials/drupal-8-convert-to-composer.md
```diff
@@ -0,0 +1,255 @@
...
+
+1. When reviewing the site, take stock of exactly what versions of modules and themes you depend on. One way to do this is to change to run the `pm:projectinfo` Drush command from within a contributed modules folder (e.g. `/modules`, `/themes`, `/themes/contrib`, `/sites/all/themes`, `/sites/all/themes/contrib`, etc.).
+
+  ```bash{promptUser:user}
+  terminus drush $SITE.dev -- pm:projectinfo --fields=name,version --format=table
+  ```
+
+  This will list each module followed by the version of that module that is installed.
+
+1. You can add these modules to your new codebase using Composer by running the following for each module in the `$SITE-composer` directory:
```
greg-1-anderson, line 87: I don't think there is a `$SITE-composer` directory in this guide; we're still working in the original $SITE directory at this point.

source/partials/drupal-8-convert-to-composer.md
```diff
@@ -0,0 +1,255 @@
...
+
+Manually copy custom code from the existing site repository to the Composer-managed directory.
+
+#### Modules and Themes
+
+Modules:
+
+```bash{promptUser:user}
+mkdir -p ../$SITE-composer/web/modules/custom # create the directory if it doesn't already exist
+cp -r modules/custom/awesome_module ../$SITE-composer/web/modules/custom
```
greg-1-anderson, line 198: Instead of the lines above, use:
```
git checkout master modules/custom
git mv modules/custom web/modules/
git commit -m "Copy custom modules"
```

source/partials/drupal-8-convert-to-composer.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+Themes:
+
+```bash{promptUser:user}
+mkdir -p ../$SITE-composer/web/themes/custom # create the directory if it doesn't already exist
+cp -r themes/custom/great_theme ../$SITE-composer/web/themes/custom
```
greg-1-anderson, line 205: Instead of the lines above, use:
```
git checkout master themes/custom
git mv themes/custom web/themes/
git commit -m "Copy custom themes"
```

source/partials/drupal-8-convert-to-composer.md
```diff
@@ -0,0 +1,255 @@
...
+git status # Ensure working tree is clean
+git show master:sites/default/settings.php > web/sites/default/original-settings.php
+diff -Nup --ignore-all-space web/sites/default/settings.php web/sites/default/original-settings.php
+# edit web/sites/default/settings.php and commit as needed
+rm web/sites/default/original-settings.php
+```
+
+The resulting `settings.php` should have no `$databases` array.
+
+#### Configuration
```
greg-1-anderson, line 226: The configuration instructions above already relocated the config to the preferred location, so this section is no longer necessary.

source/partials/drupal-8-convert-to-composer.md
```diff
@@ -0,0 +1,255 @@
...
+
+We don't recommend that you completely overwrite the `settings.php` file with the old one, as it contains customizations for moving the configuration directory you don't want to overwrite, as well as platform-specific customizations.
+
+```bash{promptUser:user}
+git status # Ensure working tree is clean
+git show master:sites/default/settings.php > web/sites/default/original-settings.php
+diff -Nup --ignore-all-space web/sites/default/settings.php web/sites/default/original-settings.php
+# edit web/sites/default/settings.php and commit as needed
+rm web/sites/default/original-settings.php
+```
```
greg-1-anderson, line 222: I'm not sure whether we should do anything about the config directory, e.g. it could maybe be set like this:
```
$settings['config_sync_directory'] = dirname(dirname(dirname(__DIR__))) . '/config';
```

greg-1-anderson: This would have to go in settings.php after settings.pantheon.php is included. Maybe we should fix the Pantheon settings file instead.

source/partials/drupal-8-convert-to-composer.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Deploy to a Multidev
+
+Push the changes to a Multidev called `composerify` to safely test the site without affecting the Dev environment:
+
+```bash{promptUser:user}
+git add .
+git commit -m "convert to composer"
+git push origin composerify && terminus env:create $SITE.dev composerify
+```
```
greg-1-anderson, line 253: We have been committing as we go along, so the `git add` and `git commit` lines are not needed here.

source/partials/drupal-8-convert-to-composer.md
```diff
@@ -0,0 +1,255 @@
...
+1. In your local terminal, change directories to the site project. For example, if you keep your projects in a folder called `projects` in the home directory:
+
+  ```bash{promptUser:user}
+  cd ~/projects/$SITE/
+  ```
+
+1. Add the Pantheon Drupal Upstream as a new remote called `ic`, fetch the `ic` branch, and checkout to a new local branch based on it called `composerify`:
+
+  ```bash{promptUser:user}
+  git remote add ic git@github.com:pantheon-upstreams/drupal-project.git && git fetch ic && git checkout -b composerify ic/master
```
jburianek, line 31: This command causes the `composerify` branch to track `master`:
```
❯ git checkout -b composerify ic/master
Branch 'composerify' set up to track remote branch 'master' from 'ic'.
Switched to a new branch 'composerify'
```

jburianek: You probably want the `--no-track` flag:
```
❯ git checkout --no-track -b newbranch ic/master
Switched to a new branch 'newbranch'
```

jburianek:
```
❯ git branch -vv
  composerify e51c478 [ic/master] Pantheon Drupal 9 update
* newbranch   e51c478 Pantheon Drupal 9 update
```

source/partials/drupal-8-convert-to-composer.md
```diff
@@ -0,0 +1,255 @@
...
+You've now committed the code to the local branch. Deploy that branch directly to a new Multidev and test the site in the browser. If the site doesn't load properly, clear the cache. If there are any issues, utilize the site's logs via `terminus drush $SITE.composerify -- wd-show` to inspect the watchdog logs, or follow the directions in our documentation on [log collection](/logs).
+
+### Deploy to a Multidev
+
+Push the changes to a Multidev called `composerify` to safely test the site without affecting the Dev environment:
+
+```bash{promptUser:user}
+git add .
+git commit -m "convert to composer"
+git push origin composerify && terminus env:create $SITE.dev composerify
```
jburianek, line 252:
```suggestion
git push -u origin composerify && terminus env:create $SITE.dev composerify
```

You probably want the `-u` flag here so that the `composerify` will track the remote `composerify` branch.

Also, I think we called our remote `ic` instead of `origin` earlier.

source/partials/drupal-8-convert-to-composer.md
carolyn-shannon: added --no-track

source/partials/drupal-8-convert-to-composer.md
carolyn-shannon: removed `git add` and `git commit` lines

source/partials/drupal-8-convert-to-composer.md
carolyn-shannon: @greg-1-anderson can you clarify what we should do here?

source/partials/drupal-8-convert-to-composer.md
carolyn-shannon: removed Configuration section.

source/partials/drupal-8-convert-to-composer.md
carolyn-shannon: @greg-1-anderson just checking...should the second line read:
`git mv themes/custom web/themes/custom/` ?

carolyn-shannon: replaced with suggested commands

source/partials/drupal-8-convert-to-composer.md
carolyn-shannon: commands replaced

source/partials/drupal-8-convert-to-composer.md
carolyn-shannon: changed `$SITE-composer` to `$SITE`

carolyn-shannon (commented):
updated core updates page, added a new page to the Drupal 9 migration guide: "Upgrade to Drupal 9"

PR #6184, opened 2020-12-21
Add information about Jetpack

nullvariable:
re-enable xmlrpc for jetpack users

Closes: #6185 

## Summary

**[WordPress Plugins and Themes with Known Issues](https://pantheon.io/docs/plugins-known-issues#jetpack)** - Noted that Jetpack needs xmlrpc enabled in `pantheon.yml`.

**[WordPress Best Practices](https://pantheon.io/docs/wordpress-best-practices#avoid-xml-rpc-attacks)** - Now that XML-RPC is disabled in the upstream, replaced the section on disabling it by `pantheon.yml` with a section on re-enabling it for required use cases.

## Post Launch


pantheondocs:
Lighthouse Audit Results:

| Description | [Live](https://pantheon.io/docs) | [pr-6184](https://pr-6184--pantheon-docs.my.pantheonfrontend.website/docs) | Difference |
| --- | --- | --- | --- |
| Performance Score | `57.99999999999999` | `56.99999999999999` | `-1` |
| Best Practice Score | `79` | `86` | `7` |
| Accessibility Score | `88` | `88` | `0` |

Lighthouse Audit Reports:

- [`pr-6184` HTML Report](https://circleci.com/gh/pantheon-systems/documentation/25459/artifacts/0//lighthouse_data/lighthouse-audit-dev.html)
- [`Pantheon Docs` HTML Report](https://circleci.com/gh/pantheon-systems/documentation/25459/artifacts/0//lighthouse_data/lighthouse-audit-reference.html)


[Preview Environment](https://pr-6184--pantheon-docs.my.pantheonfrontend.website/docs)

alexfornuto (approved):
Thanks @nullvariable! Now that I've authored on this PR, it will require a copy review from someone else in @pantheon-systems/docs-admins.

carolyn-shannon:
`xmlrpc.php` is disabled by default in the Pantheon upstream. We should clarify what enabling it means in terms of support scope, how it might impact site security, and clarify for users how it might impact site traffic metrics. Moving this to blocked status pending further conversation about the best way to address these questions.

carl-alberto:
+1 on this PR, we've been seeing this pop once in a while in the support end since the 5.4.2 https://github.com/pantheon-systems/WordPress/blob/5.4.2/pantheon.upstream.yml Pantheon upstream update as that is the time that is started blocking the WP XMLrpc, we can easily point this out in the doc how to fix it and the customer will be the one liable in protecting their site based from xmlrpc based attacks so most likely they will rely on Jetpack's protect module https://jetpack.com/2015/10/12/jetpack-protection-from-brute-force-xml-rpc-attacks/ for that

carl-alberto:
While this is not in the docs yet, just a quick reference to make Jetpack work in similar cases is adding this in the `pantheon.yml` and having some values for the `protected_web_paths` eg:

```
protected_web_paths_override: true
protected_web_paths:
  - /private
  - /wp-content/uploads/private
```

EdwardAngert:
[preview](https://pr-6184--pantheon-docs.my.pantheonfrontend.website/docs/plugins-known-issues#jetpack)

EdwardAngert:
1. according to [Avoid XML-RPC Attacks](https://pantheon.io/docs/wordpress-best-practices#avoid-xml-rpc-attacks), we suggest that users add xmlrpc to protected webpaths, but it's [already there](https://github.com/pantheon-systems/WordPress/blob/default/pantheon.upstream.yml#L16-L19) - that might need some rewording
2. our SEs generally suggest not using Jetpack, so we might be able to use that as a disclaimer and point users to other resources

carl-alberto (approved):
Everything related to Jetpack and enabling xmlrpc to make it works looks good!

source/content/plugins-known-issues.md
```diff
@@ -523,7 +523,39 @@ ___
...
+protected_web_paths:
+  - /private
+  - /wp-content/uploads/private
+```
+
+This will maintain the normal security settings for other paths, but allows access for XMLRPC.
+
+<Alert title="Note"  type="info" >
+
+Pantheon does not support XML-RPC if it is enabled.
```
hckia, line 545: While we most certainly do not support this, we should probably include more documentation about how to secure a site that wants to utilize Jetpack from XMLRPC.php attacks.

If you disable protection by removing xmlrpc.php from protected paths, you should still try to block all xmlrpc.php requests EXCEPT from allowed IPs.

https://jetpack.com/support/how-to-add-jetpack-ips-allowlist/ 

You can use the list of IPS above, and within our [advanced redirects section for WordPress](https://pantheon.io/docs/advanced-redirects#restrict-access-to-paths-based-on-ip) add those ips to the `is_from_trusted_ip` function under the `trusted_ips` array. Then, modify the `disallow_uri` array to the following...

```
    $disallow_uri = array(
        '/wp-login.php',
        '/wp-admin/',
        '/xmlrpc.php',
    ); 

```

This would need to be tested, but I think it's the more responsible approach. Kudos to Kris.

Also a note on the [Protection Module to prevent brute-force attacks](https://jetpack.com/support/protect/#:~:text=Jetpack's%20Protect%20module%20collects%20information,B%2C%20it%20would%20be%20blocked.).  This will face similar hurdles as security plugins, and under enough traffic could harm the site or the file system itself. It's not a good idea to enable this, will likely earn itself a place under "Plugins with known issues", even though this is a module within a plugin.

hckia (changes requested):
This will work, but we should still address how sites that disable this protection can still protect themselves. The Protection module can cause problems as well.

hckia (approved):
LGTM

wordsmither:
lgtm

PR #6602, opened 2021-08-12
Persistent Cache Public Docs

### LINKED ISSUE #6601: Persistent Cache Public Docs
Concept overview of the persistent cache feature, how it works with AGCDN. This will appear as a section in the AGCDN doc.

michellecolon-pantheon:
Closes #6601 

**[Pantheon Global CDN](https://pantheon.io/docs/global-cdn)** -  Add Persistent Cache section to AGCDN page.

## Remaining Work

- [ ] Create a Persistent Cache Doc- will be done in a separate PR

## Post Launch

### DIFF
--- a/source/content/global-cdn.md
+++ b/source/content/global-cdn.md
@@ -32,16 +32,22 @@ We recommend installing the Pantheon Advanced Page Cache [plugin](https://wordpr
 
 For more details, see [Clearing Caches for Drupal and WordPress](/clear-caches).
 
-### Persistent Cache
+## Persistent Cache
 
-Serve your Drupal or WordPress site, even when it's down. If the server is not responding and can't serve a new copy of a page, it will use a cached version instread of displaying an error. With Persistent Cache, the goal is to provide a seamless, uninterrupted experience for the user.
+Serve your Drupal or WordPress site, even when it's down. If the server is not responding and can't serve a new copy of a page, it will use a cached version instead of displaying an error. With Persistent Cache, the goal is to provide a seamless, uninterrupted experience for the user.
 
-#### How Do I Do the Thing?
+### How Do I Configure This Setting?
+
+On [Drupal](/drupal-cache#drupal-8-performance-configuration) and [WordPress](//wordpress-cache-plugin#pantheon-page-cache-plugin-configuration), you can adjust your CDN edge configuration to serve stale content for a specific amount of time.
 
 
-### Troubleshooting
+### Caching Exceptions
 
-#### How I Know If It's Working?
+Your site’s CMS page-level caching must be correctly configured in order to take advantage of Persistent Cache. 
+
+Users with session-style cookies set, or a `NO_CACHE` cookie set will bypass the cache, and will not see cached content. For best results, set the `NO_CACHE` cookie to persist longer than the site’s page cache (this includes logged in users and authenticated traffic). You can learn more about the exceptions to page caching rules in [Caching: Advanced Topics](caching-advanced-topics#allow-a-user-to-bypass-the-cache). 
+
+### How I Know If It's Working?
 
 
 ## Frequently Asked Questions
--- a/source/content/global-cdn.md
+++ b/source/content/global-cdn.md
@@ -32,6 +32,18 @@ We recommend installing the Pantheon Advanced Page Cache [plugin](https://wordpr
 
 For more details, see [Clearing Caches for Drupal and WordPress](/clear-caches).
 
+### Persistent Cache
+
+Serve your Drupal or WordPress site, even when it's down. If the server is not responding and can't serve a new copy of a page, it will use a cached version instread of displaying an error. With Persistent Cache, the goal is to provide a seamless, uninterrupted experience for the user.
+
+#### How Do I Do the Thing?
+
+
+### Troubleshooting
+
+#### How I Know If It's Working?
+
+
 ## Frequently Asked Questions
 
 ### I already have a CDN. Can I use it with the Pantheon Global CDN?
--- a/source/content/global-cdn.md
+++ b/source/content/global-cdn.md
@@ -92,9 +92,9 @@ To test how stale cache is served, compare the header results of a page refresh:
 1. In a terminal, cURL the site headers filtered for stale cache:
 
   ```bash{promptUser: user}
-  curl --head https://pantheon.io/docs | grep PContext-Resp-Is-StaleHTTP/2 301
+  curl --head https://pantheon.io/docs | grep PContext-Resp-Is-Stale
   ```  
-  If the result includes `PContext-Resp-Is-StaleHTTP/2 301`, the page has been successfully served from stale cache. 
+  If the response headers include `PContext-Resp-Is-Stale`, the page has been successfully served from stale cache. 
   
 </Tab>
 
--- a/source/content/global-cdn.md
+++ b/source/content/global-cdn.md
@@ -52,35 +52,19 @@ Users with session-style cookies set, or a `NO_CACHE` cookie set will bypass the
 
 ### Confirm That Persistent Cache Works
 
-To test how stale cache is served, compare the header results of a page refresh of the site's Dev environment to the same page in Maintenance Mode.
+To test how stale cache is served, compare the header results of a page refresh:
 
-Navigate to the page using [Firefox](https://developer.mozilla.org/en-US/docs/Tools) or [Chrome](https://developer.chrome.com/docs/devtools/), and in the browser's developer tools open the **Network** tab and view the response headers for the page or asset. Refresh the page in the browser for the newest information.
+1. Navigate to the page using [Firefox](https://developer.mozilla.org/en-US/docs/Tools) or [Chrome](https://developer.chrome.com/docs/devtools/), and in the browser's developer tools open the **Network** tab and view the response headers for the page or asset
+1. Go to your site's Dev environment in Maintenance Mode
+1. Clear the cache from either the Advanced Page Cache or from the dashboard
+1. Refresh the page in the browser for the newest information
 
 Examine the results for the `age` and `max-age`.
 
-To examine the headers through the command line, use a `curl` command to examine the headers before and after you set the site to Maintenance Mode:
+To examine the headers through the command line, use a `curl` command. 
 
 ```bash{outputLines: 2-20}
-curl --head https://pantheon.io/docs
-HTTP/2 301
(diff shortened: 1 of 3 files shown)


michellecolon-pantheon:
@mattcallahan Can you please take a look at this PR? 🙂

pantheondocs:
Lighthouse Audit Results:

| Description | [Live](https://pantheon.io/docs) | [pr-6602](https://pr-6602--pantheon-docs.my.pantheonfrontend.website/docs) | Difference |
| --- | --- | --- | --- |
| Performance Score | `59` | `62` | `3` |
| Best Practice Score | `79` | `79` | `0` |
| Accessibility Score | `96` | `96` | `0` |

Lighthouse Audit Reports:

- [`pr-6602` HTML Report](https://circleci.com/gh/pantheon-systems/documentation/29397/artifacts/0//lighthouse_data/lighthouse-audit-dev.html)
- [`Pantheon Docs` HTML Report](https://circleci.com/gh/pantheon-systems/documentation/29397/artifacts/0//lighthouse_data/lighthouse-audit-reference.html)


[Preview Environment](https://pr-6602--pantheon-docs.my.pantheonfrontend.website/docs)

source/content/global-cdn.md
```diff
@@ -32,6 +32,24 @@ We recommend installing the Pantheon Advanced Page Cache [plugin](https://wordpr
 
 For more details, see [Clearing Caches for Drupal and WordPress](/clear-caches).
 
+## Persistent Cache
+
+Serve your Drupal or WordPress site, even when it's down. If the server is not responding and can't serve a new copy of a page, it will use a cached version instead of displaying an error. With Persistent Cache, the goal is to provide a seamless, uninterrupted experience for the user.
+
+### How Do I Configure This Setting?
+
+On [Drupal](/drupal-cache#drupal-8-performance-configuration) and [WordPress](//wordpress-cache-plugin#pantheon-page-cache-plugin-configuration), you can adjust your CDN edge configuration to serve stale content for a specific amount of time.
```
mbaynton, line 41: The settings built into the CMS don't impact how long the CDN will serve stale content, rather they impact how long the CDN is allowed to regard the content as "fresh." We've always supported this, whereas the new feature is not configurable at all by the customer.

source/content/global-cdn.md
```diff
@@ -32,6 +32,24 @@ We recommend installing the Pantheon Advanced Page Cache [plugin](https://wordpr
 
 For more details, see [Clearing Caches for Drupal and WordPress](/clear-caches).
 
+## Persistent Cache
```
mbaynton, line 35: FWIW, I object to the name that the company is converging on for the feature because it implies a specific behavior by the CDN - storing objects with persistence guarantees - that Fastly explicitly disclaims and so we aren't delivering.

I think there are many great ways we could name this feature that convey the customer value without using techy words like "cache," thereby sidestepping potential deception. For example "Enhanced Uptime Protection" or something like that.

source/content/global-cdn.md
```diff
@@ -32,6 +32,24 @@ We recommend installing the Pantheon Advanced Page Cache [plugin](https://wordpr
...
+
+Serve your Drupal or WordPress site, even when it's down. If the server is not responding and can't serve a new copy of a page, it will use a cached version instead of displaying an error. With Persistent Cache, the goal is to provide a seamless, uninterrupted experience for the user.
```
mbaynton, line 37: This sentence is true, but for users familiar with CDN behaviors it leaves uncertainty about what we're really doing with this feature.
```suggestion
Serve your Drupal or WordPress site, even when it's down. If the server is not responding and can't serve a new copy of a page, stale cached versions will be favored over displaying an error. With Persistent Cache, the goal is to provide a seamless, uninterrupted experience for the user.
```

source/content/global-cdn.md
```diff
@@ -32,6 +32,24 @@ We recommend installing the Pantheon Advanced Page Cache [plugin](https://wordpr
...
+
+### How Do I Configure This Setting?
+
+On [Drupal](/drupal-cache#drupal-8-performance-configuration) and [WordPress](//wordpress-cache-plugin#pantheon-page-cache-plugin-configuration), you can adjust your CDN edge configuration to serve stale content for a specific amount of time.
+
+
+### Caching Exceptions
```
mbaynton, line 44: A related suggestion we might want to let users know about (and exactly why the feature is not a "persistent cache") --
The probability that an object will successfully be served stale drops appreciably if its original fresh TTL was less than `3700` seconds. The reason is that such objects are only cached in RAM on Fastly's cache nodes and, as a limited resource, they are more likely to experience cache eviction. In fact, the likelihood that they will experience RAM eviction over the hours-to-days spans needed to retain a large pool of stale resources is extraordinarily high.

So the customer-actionable recommendation here would be to set their TTLs to >= 3700 seconds to fully benefit from this feature.

Fastly reference: https://docs.fastly.com/en/guides/serving-stale-content#why-serving-stale-content-may-not-work-as-expected

EdwardAngert: I added this recommendation to the section

source/content/global-cdn.md
EdwardAngert: I combined two sections and renamed this one `Caching Exceptions - How long does content stay fresh?` for now - not sure I'm totally settled on it, but I think it's closer than giving the impression that the CDN is the configurable part of this feature

source/content/global-cdn.md
```diff
@@ -32,6 +32,57 @@ We recommend installing the Pantheon Advanced Page Cache [plugin](https://wordpr
...
+x-cache: HIT, HIT
+x-cache-hits: 1, 1
+x-timer: S1629819022.932985,VS0,VE1
+pantheon-trace-id: be58e6a03a904fbfa64515ee136ffd34
+vary: Cookie, Cookie
+age: 9654
+accept-ranges: bytes
+via: 1.1 varnish, 1.1 varnish
+content-length: 162
+```
```
EdwardAngert, line 84: I'm not super clear on which would be the best header response to look out for, so this whole set of lines should get reviewed by @mbaynton and @mattcallahan and then we can either add a `//highlight-start` and `-end` to show people what to look out for in the output, or `grep` for a keyword if that's easy enough

EdwardAngert (changes requested):

EdwardAngert:
[preview](https://pr-6602--pantheon-docs.my.pantheonfrontend.website/docs/global-cdn#persistent-cache)

source/content/global-cdn.md
mbaynton: I don't think this response actually demonstrates the feature. This response looks like a healthy cache hit. Also pantheon.io specifically redirects all of the time when you're using `cURL` for some reason, I've noticed this as well but am unsure what's causing it.
To see the feature operating, one must also prevent Fastly's regular old fresh cache for the page from being used. There's basically two ways to do that, either wait more than the `max-age` (86400s here, not fun, may not work anyway since the cache, you know, isn't persistent) or purge the cache, either for the page using for example the Advanced Page Cache module or for the whole site from the dashboard.
Once you've done that, you should see an additional header in the response, `PContext-Resp-Is-Stale: true`, which indicates the reason a non-error response was returned was because of this feature.

EdwardAngert: thanks, @mbaynton! you're right - this snippet doesn't show staled (the text above it just introduces it as a way to view headers) - but that `PContext-Resp-Is-Stale: true` is perfect. I can add another example bash command like

```
curl --head https://pantheon.io/docs | grep PContext-Resp-Is-Stale
```

to this point:

> either wait more than the max-age (86400s here, not fun, may not work anyway since the cache, you know, isn't persistent) or purge the cache, either for the page using for example the Advanced Page Cache module or for the whole site from the dashboard.

should the steps to test this be more like (adding step 2):
1. maint mode
2. clear cache either from advanced page cache or dash
3. check headers

michellecolon-pantheon: @mbaynton - All updated, if you want to take a look 👀

source/content/global-cdn.md
```diff
@@ -52,36 +52,67 @@ Users with session-style cookies set, or a `NO_CACHE` cookie set will bypass the
...
+1. Navigate to your site's Dev environment and set the site to Maintenance Mode.
+  
+1. Clear the cache from either the Advanced Page Cache module or from the Dashboard.
+
+1. In a terminal, cURL the site headers filtered for stale cache:
+
+  ```bash{promptUser: user}
+  curl --head https://pantheon.io/docs | grep PContext-Resp-Is-StaleHTTP/2 301
+  ```  
+  If the result includes `PContext-Resp-Is-StaleHTTP/2 301`, the page has been successfully served from stale cache.
```
mbaynton, line 97:
```suggestion
  If the result includes `PContext-Resp-Is-Stale`, the page has been successfully served from stale cache. 
```
The HTTP/2 part shouldn't appear like that, it may be your terminal interleaving curl's stdout and stderr?

Try
```bash{promptUser: user}
  curl -v https://pantheon.io/docs 2>&1 | grep PContext-Resp-Is-Stale
  ```

File: source/content/global-cdn.md [hunk 7]
```diff
@@ -52,36 +52,67 @@ Users with session-style cookies set, or a `NO_CACHE` cookie set will bypass the
...
+
+1. Navigate to the page using [Firefox](https://developer.mozilla.org/en-US/docs/Tools) or [Chrome](https://developer.chrome.com/docs/devtools/), and in the browser's developer tools open the **Network** tab and view the response headers for the page or asset.
+
+1. Go to your site's Dev environment and set the site to Maintenance Mode.
+  
+1. Clear the cache from either the Advanced Page Cache module or [from the Dashboard](/clear-caches#pantheon-dashboard).
+
+1. Go back to your page and view the Developer Tools, and Refresh for the newest header responses.
+  
+    If the result includes `PContext-Resp-Is-StaleHTTP/2 301`, the page has been successfully served from stale cache.
```
mbaynton, line 111:
```suggestion
    If the response headers include `PContext-Resp-Is-Stale`, the page has been successfully served from stale cache. 
```

michellecolon-pantheon (commented):
@mbaynton - Please have a look and let me know if there is anything else that should be updated!

mbaynton:
Sure @michellecolon-pantheon looks great

michellecolon-pantheon:
@mbaynton - TYSM!

EdwardAngert (approved):

PR #6983, opened 2022-01-19
[UpdateTool - Terminus Information] Update commands and releases.

pantheon-upstream:


kporras07:
I found the error with this and I'm fixing it in update-tool. Once ready, a new PR will be created

PR #7318, opened 2022-05-04
Updates info on YITH WooCommerce plugins with MPDF

morganestes:
## Summary

**[WordPress Plugins and Themes with Known Issues](https://pantheon.io/docs/plugins-known-issues)** - Updates the YITH filters for MPDF file storage.

Since the same issue affects multiple YITH plugins that use the MPDF library, this makes the filter function reusable so it can be used as a callback for multiple filters.

Instead of adding two separate sections for each plugin, I combined them into a single section that can be easily updated as new plugins and filter names are needed.

Also updated the code sample to say the file is `functions.php` instead of `wp-config.php` since you can't use the `wp_upload_dir` function then.

## Effect

Changes the specific plugin section to a more general one, which is a new change for the page.

Code sample assumes the user's PHP version supports argument and return types. If that's not the case then the code will fail unless those are removed when pasting.

## Post Launch

### DIFF
--- a/source/content/plugins-known-issues.md
+++ b/source/content/plugins-known-issues.md
@@ -1361,26 +1361,44 @@ ___
 1. Define the [FS_METHOD in the wp-config](#define-fs_method).
 ___
 
-## YITH WooCommerce Request a Quote
+## YITH WooCommerce Extensions with MPDF library
 
-<ReviewDate date="2022-04-8" />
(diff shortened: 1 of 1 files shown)


joa-pan:
Thank you for this contribution @morganestes 🎉

joa-pan:
Currently testing this plugin with CSE

PR #7319, opened 2022-05-04
This is to avoid confusion.

v0jpantheon:
Since the user is in `code/wp-content/uploads`, we need clearer instructions that the user may need to go back to `code/wp-content` directory. It will be easier to follow.


joa-pan:
Thank you @v0jpantheon. Much appreciated 🎉

joa-pan (approved):
LGTM

PR #7331, opened 2022-05-10
Drupal CORS

v0jpantheon:
Just to be fair with both platforms.

CORS reference (Conincidentally this is specifically for pantheon):
https://drupal.stackexchange.com/questions/245903/how-do-i-set-up-cors#answer-247816


v0jpantheon:
We may need to have a dedicated CORS page for drupal if we really wanted to dive into the topic.

Only related article I can find in Drupal CORS for pantheon is this one:
https://github.com/pantheon-systems/drops-8/issues/22

whitneymeredith:
LGTM

azovsky:
Hi, I added CORS for my Drupal 9 site on Pantheon, but 

How to add CORS on for "direct" files, not just for pages that "generated via Drupal" ?

... for files there is no `access-control-allow-origin` header.

Thank you!

PR #7434, opened 2022-06-15
Reduce confusing context

v0jpantheon:
This might be not very clear for some of the users. Instead, we should add more straightforward directions like go to code/ (use `cd code`).


joa-pan (approved):
LGTM

PR #7527, opened 2022-07-25
Added Drupal 9 settings for tmp files.

v0jpantheon:
Added Drupal 9 tab settings for tmp files.


v0jpantheon:
We can also add a verification by accessing the page, like other tabs:

Verify the setting by accessing admin settings page: `/admin/config/media/file-system`
Screen Shot 2022-07-26 at 3 36 41 AM

PR #7862, opened 2022-10-28
Workaround fix for SSH for macos ventura

v0jpantheon:
there's some issues with some customers that's causing ssh to suddenly fail after updating to MacOS Ventura.

Updating docs to accommodate workaround


v0jpantheon:
Hi team, just checking if this somehow got removed in the latest docs? We may not be able to reference this to the customers anymore if it got removed.
https://pantheon.io/docs/ssh-keys#connections-fail-with-no-matching-host-key-type-found-their-offer-ssh-rsa

Not sure if there are other workarounds than this.

PR #8064, opened 2023-01-18
Minor update

v0jpantheon:
since its alternative to the terminus, it should be `you can` instead of `you might`


whitneymeredith (approved):
LGTM

PR #8090, opened 2023-01-31
[CMS-1257] Custom upstreams commands recent changes.

kporras07:
## Summary

**[Use Upstream with Integrated Composer](https://pantheon.io/docs/guides/integrated-composer/ic-upstreams)** - Update docs on how to add and update upstream dependencies.

- [ ] CMSO-1230 should be done before merging this.

**Release**:
- [ ] When ready
- [ ] After date: $DATE

## Post Launch


source/content/guides/custom-upstream/02-create-custom-upstream.md
```diff
@@ -56,8 +56,6 @@ Follow the steps for your preferred Git host below to create your repository.
 
   As a company, Pantheon is trying to use [more inclusive language in our repositories](https://pantheon.io/blog/diversity-equity-and-inclusion-pantheon?docs). While our team works to make Custom Upstreams less reliant on older naming conventions, new Custom Upstreams currently default to using the `master` branch name.
 
-  Git's default naming convention differs from GitHub's. If you don't initialize the repository on GitHub, Git will assign the default branch name as `master` when you clone the repository locally.
```
greg-1-anderson, line 59: Note for docs: the deleted text was true for old versions of git, but current versions now also default to `main` for the initial default branch.

source/content/guides/integrated-composer/05-ic-upstreams.md
```diff
@@ -29,47 +29,31 @@ Upstream refers to the source code that is hosted in the Pantheon code repositor
...
 
-1. Change into the Upstream's configuration directory:
-
-   - Drupal:
-
-    ```bash{promptUser: user}
-    cd upstream-configuration
-    ```
-
-   - WordPress:
```
greg-1-anderson, line 42: Above the docs refer to `https://github.com/pantheon-upstreams/wordpress-project`. We should decide on a strategy for WordPress, i.e. are we going to copy the upstream management code from drupal-composer-managed into wordpress-project, or are we going to recommend the Bedrock Composer project here? We probably shouldn't publish this page until we decide; if it is going to be a while before we do either, then maybe we could make a note on how to copy this code from drupal-composer-mananaged yourself. Such instructions would be helpful for folks on older upstreams (although maybe we want to just tell them to use the conversion tool to get on the current upstream).

source/content/guides/integrated-composer/05-ic-upstreams.md
```diff
@@ -29,47 +29,31 @@ Upstream refers to the source code that is hosted in the Pantheon code repositor
...
+1. Run the `composer upstream-require` command for each dependency:
 
     ```bash{promptUser: user}
-    cd upstream-config 
+    composer upstream-require drupal/pkg-name [--no-update]
     ```
 
-1. Run the `composer require` command for each dependency:
+     - `--no-update` tells Composer to disable automatic updates of the dependency. This makes Composer faster when adding dependencies to the Upstream as shown here. This is useful if you don't want to pin versions for your dependencies. Do not include this option if you want to pin specific versions for your dependencies.
```
greg-1-anderson, line 40: It's not necessary to specify `--no-update` here; the script will automatically include `--no-update` if you're not pinning versions, and it will automatically put in `--no-install` if you are.

Use `composer update-upstream-dependencies` to start pinning changes.

source/partials/ic-upstream-structure.md
```diff
@@ -30,7 +30,7 @@ code/
     - `composer.json`: Composer automatically updates `composer.json` with customizations for the upstream. Avoid manually modifying this file.
 - `pantheon.upstream.yml`: The `build_step: true` directive in `pantheon.upstream.yml` enables the build step.
 
-When a site is created, Pantheon runs `composer install`, generates a `composer.lock` file, and commits it back to the site’s code repository. To avoid potential merge conflicts after applying an upstream update, do not commit the `composer.lock` file to the upstream repository.
+When a site is created, Pantheon runs `composer install`, generates a `composer.lock` file, and commits it back to the site’s code repository. To avoid potential merge conflicts after applying an upstream update, do not commit the `composer.lock` file from your upstream root to the upstream repository.
```
greg-1-anderson, line 33: We could maybe add a little helpful advice for working with custom upstreams locally:
```
Since the `composer.lock` file should be committed for individual sites created from your custom upstream, it should not be listed in your .gitignore file. To make it easier to avoid accidentally commit a `composer.lock` file when working directly with your custom upstream, run the following command:

    echo /composer.lock >> .git/info/exclude

This will have the same effect as ignoring the file in the top-level .gitignore file, except that the `exclude` file is not saved in the repository when you push changes back to the remote origin. You therefore must re-run this command every time you make a new local clone of your upstream if you would like to protect against inadvertently committing a composer.lock file.
```

greg-1-anderson: It isn't necessary to ever make a composer.lock file in the custom upstream, but it might happen if folks test `composer update` in their upstream. I am therefore somewhat undecided about how important it is to include this advice. It's helpful, so I am inclined to include it.

source/content/guides/integrated-composer/05-ic-upstreams.md
kporras07: This is no longer valid as this is now a composer package that could be pulled anywhere

greg-1-anderson (approved):

whitneymeredith (approved):
LGTM

PR #8346, opened 2023-04-16
Explain how to select the version of NodeJS

joshkoenig:
## Summary

https://docs.pantheon.io/guides/decoupled/overview/manage-settings does not (yet) explain how to select the version of NodeJS that is used for the Front End Site runtime. This is a very important thing for developers to know, as new versions of frameworks will require newer versions of NodeJS. 

Let's make sure people can find out that `.nvmrc` is the place to specify this.


wordsmither (approved):
lgtm

PR #8619, opened 2023-07-21
Update 02-drush-versions.md "Available Drush Versions"

jrannelv:
Drupal 11 says it is available, but not really. We only support this version, Drush 11 has to be installed as a site-local Drush

Closes #

## Summary

## Effect

*
*

### Dependencies and Timing

**Release**:
- [ ] When ready
- [ ] After date: $DATE

## Post Launch

### DIFF
--- a/source/content/guides/drush/02-drush-versions.md
+++ b/source/content/guides/drush/02-drush-versions.md
@@ -27,6 +27,12 @@ Pantheon runs Drush versions based on your Drupal version:
 
 We recommend managing your site through Composer. Visit the [Build Tools Workflow](/guides/build-tools/) for information on how to use Composer to manage Drupal sites on Pantheon, or the [Convert to Composer](/guides/composer-convert) guide to convert an existing site to a Composer-managed site.
 
+<Alert title="Note"  type="info" >
+
+Drush 11 is only available via Composer. We only support this version, Drush 11 has to be installed as a site-local Drush.
+
+</Alert>
+
 ## Verify Current Drush Version
 
 You can use [Terminus](/terminus/) to verify the current version of Drush running on your Pantheon site:


rachelwhitton:
Thanks for the PR @jrannelv! 

I think we can remove content around managing Drush versions via `pantheon.yml` since newly created Drupal 9 and 10 sites default to site-local Drush 11 and 12 respectively. Newly created Drupal 7 and 8 sites default to Drush 8, which is the only compatible version according to the table in this doc. Is there a non-deprecated use case for managing Drush via `pantheon.yml`? 

What do you think? 
cc @stevector

rachelwhitton:
@greg-1-anderson can you review and weigh in on the question I asked in my comment above? 
> Is there a non-deprecated use case for managing Drush via pantheon.yml?

source/content/guides/drush/02-drush-versions.md
```diff
@@ -27,6 +27,12 @@ Pantheon runs Drush versions based on your Drupal version:
 
 We recommend managing your site through Composer. Visit the [Build Tools Workflow](/guides/build-tools/) for information on how to use Composer to manage Drupal sites on Pantheon, or the [Convert to Composer](/guides/composer-convert) guide to convert an existing site to a Composer-managed site.
 
+<Alert title="Note"  type="info" >
+
+Drush 11 is only available via Composer. We only support this version, Drush 11 has to be installed as a site-local Drush.
```
greg-1-anderson, line 32: Drupal 8 (no longer supported) starts with Drush 10: https://github.com/pantheon-systems/drops-8/blob/default/pantheon.upstream.yml#L10. Note there is a comment about "Drush 10" above that only applies to Drupal 8.

Drupal 9 & 10 add Drush via Composer. See https://www.drush.org./12.x/install/ for supported Drush versions, and what versions of Drupal they work with.

namespacebrian:
I think only Drush 8 works with Drupal 7, right?   in which case there's no good reason for D7 sites to change their drush version via `pantheon.yml`.  I feel D8 is sufficiently EOL that IMO we should avoid documenting "supported" configurations for D8 because D8 falls pretty squarely in the "_possible_ but not _'supported'_" category.

So I agree with @rachelwhitton that there's no good reason to mention choosing the drush version via `pantheon.yml` at all.  The dichotomy between site-local vs `pantheon.yml` causes a lot of confusion for customers around the term "supported"..

IMO we should just document that drush version is configured via composer.json and link out to drush's own documentation about which versions are still supported by the community as the point of reference for what is "supported".

stevector:
I started #9010 as a larger overhaul. @namespacebrian can I get your thoughts there on how I handled site-local vs. global? As long as `pantheon.yml` does indeed change Drush versions, I think we need to acknowledge it. If Terminus directly invoked the `drush` in `vendor` when it was there then I'd be fine downplaying the global drush more.

PR #8650, opened 2023-08-18
PDS theme

mel-miller:
## Summary

This PR updates the site theme to use the Pantheon Design System. 

There will be subsequent, and much smaller, theme updates in the future, but this is the first pass at PDS implementation. 

If I noticed accessibility issues while working on a particular component, I fixed them but did not fix any outside of the scope I was working in. 

**Release**:
- [x] When ready
- [ ] After date: $DATE


source/data/home.yaml
```diff
@@ -1,111 +1,107 @@
...
+  title: "Pantheon Docs"
   call-to-action:
-    title: 'Getting Started Guide'
-    sub-title: 'New to the Pantheon platform? Check out our step-by-step guide to learn all the basics.'
-    url: '/guides/getstarted/'
-    type: 'guide'
+    subtitle: "New to the Pantheon platform? Check out our step-by-step guide to learn all the basics."
+    url: "/guides/getstarted/"
+    linkText: Getting started guide
+    imageSrc: "https://cdn.bfldr.com/MEM5087K/at/wprcks7sx3wgqghb3q84p5/WEB-2x1-ProductDetailFrontEndSites-Hero_v2.psd?format=png&crop=1187%2C801%2Cx11%2Cy198"
```
stevector, line 7: What is this domain name? Can we instead put this image inside this repo?

mel-miller: @stevector — This is a Brandfolder CDN link. Brandfolder is our new DAM that is replacing Bynder. One of the reasons we chose it was for CDN capability. We hope to eventually host these decorative-type images from there for all sites. 

Are you seeing performance issues?

src/layout/header/index.js
```diff
@@ -1,205 +1,145 @@
...
-                    <Link id="troubleshoot" to="/troubleshoot">
-                      Troubleshoot
-                    </Link>
-                  </li>
-                </div>
-              </div>
+// Links for NavMenu component.
+const mainNavigationLinks = [
+  {
+    linkContent: (
```
stevector, line 16: It strikes me as a little odd to have these 4 links in the top left of the header with differing behaviors.

Screenshot 2023-09-28 at 2 37 08 PM


* The logo is like a home button next to Docs Home. But it goes to the top level `pantheon.io`. I think that's a little confusing but maybe fine.
* "Docs Home" next to a drop down labeled "Documentation" feels duplicative in a different way.
* Then "Support" goes to `dashboard.pantheon.io`. Which is the subdomain used by the links on the right of the header nav.

I don't want to hold up this PR unnecessarily. I just see opportunity for further improvement. Perhaps a small change that would alleviate my main concern would be changing "Documentation" to "Topics" or "Docs Topics" since it is a drop down of Topics

mel-miller: That's fine. I can certainly change it to "Topics".  

This exercise is mainly to theme what was already there and try to improve if possible. The main reason for adding docs home is that on the current site, there is no way to get to the homepage from the navbar. Would you rather the Pantheon logo go to docs home?

stevector: > The main reason for adding docs home is that on the current site, there is no way to get to the homepage from the navbar.

Looks like that on the current design, the "Documentation" link that triggers the drop down nav is itself a link to the docs home page. I think I'd prefer to keep that behavior and then drop the separate "Docs Home" link

mel-miller: The menu component is not currently built to act that way for accessibility best practices. Top-level links are buttons that trigger the menus and not links themselves. See example: https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/

What if we just put the "Docs home" link as the top item in the dropdown for now — then after the IA recommendations are in, we can reevaluate if we need a different component?

mel-miller: Also, do you want to move support to the right side?

stevector: > Also, do you want to move support to the right side?

I don't have a strong opinion there. Let's not block this PR on it.

rachelwhitton (approved):

PR #8665, opened 2023-08-31
Restore AddSearch Experiences (Mostly)

rachelwhitton:
### Replaces #8664  
If we want to keep debugging I'm down but also not sure if we want to sink much more time in this specific 3rd party search solution so I'm proposing this as an alternative to #8664 

## Effect 
Fixes the busted AddSearch experience across `/search` ~and 404 pages~

## Summary
Building off of some strong progress achieved by @stevector which replaced the stale AddSearch scripting with up to date installation techniques, this PR:

 - Fixes up `/search`, by binding the existing search form via CSS selector to Steve's new Search UI instance, and includes a first pass style alignment for the results, also fixes pagination 
 - ~Fixes the search experience across 404 pages, by generating a ${searchpath} query paramater based on requested URL, allowing a visitor the enter the autofilled search form which then loads the results on the separate results `/search` page~

PR #8707, opened 2023-09-27
Fix separate search results page

rachelwhitton:
replace #8665 

The search field after the PDS Theme was implemented has a better experience than the (at the time) current production theme, so I switched efforts to this PR and I'm getting expected behaviors now on local and the preview env (in #8665 separate search results page worked locally but not on the preview) 

Expected Search Experience: 
* When you are using the search widget (everywhere except `/search`) and you hit enter, you are now directed to the separate search results page with the query passed in and results shown on page load
* When you are using the separate search results page (`/search`) the results should upate as you type, hitting enter in this search field on this page should not load a new page but it should update the search URL paramater in the address bar

- [x] Outstanding: Style separate search results

### DIFF
--- a/src/layout/SearchBar/index.js
+++ b/src/layout/SearchBar/index.js
@@ -17,7 +17,7 @@ const SearchBar = ({ page }) => (
     className="pds-spacing-pad-block-start-l pds-spacing-pad-block-end-2xl"
   >
     <InputText
-      id="search-field"
+      id="search"
       aria-label="Search Pantheon Documentation"
       placeholder="Search Pantheon Documentation"
       type="search"
--- a/src/pages/search.js
+++ b/src/pages/search.js
@@ -5,36 +5,66 @@ import SEO from "../layout/seo"
 import { Container } from "@pantheon-systems/pds-toolkit-react"
 
 class Search extends React.Component {
-  componentDidMount() {
-    //On page load...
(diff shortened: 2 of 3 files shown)


mel-miller (approved):
It works! 

The search results page is themed. I'm going to also theme the dropdown search results, but I'll do that in a separate PR so as to not hold this one up. 

I'm not sure how much control you have over the markup from addsearch, but if possible the "number-of-results" should probably be an h2. Alternatively, the search hit titles could be h2 instead of h3. Either solution would work for proper heading levels on this page for accessibility purposes. I wouldn't hold up the whole thing for this, as it can always be fixed later, but I thought I'd mention it in case it was an easy change.

stevector (approved):

stevector:
> I'm not sure how much control you have over the markup from addsearch, but if possible the "number-of-results" should probably be an h2. Alternatively, the search hit titles could be h2 instead of h3. Either solution would work for proper heading levels on this page for accessibility purposes. I wouldn't hold up the whole thing for this, as it can always be fixed later, but I thought I'd mention it in case it was an easy change.

I'd prefer we handle that in a follow up.

stevector:
Also, thanks @mel-miller! I'm passing this along internally and hope to have it merged soon.

PR #8747, opened 2023-10-16
Updated collect-logs-sftp.sh.txt to skip tedious StrictHostKeyChecking

ccharlton:
Added a useful parameter in the SFTP command that skips the need to ask the user if it's okay to connect with each appserver before downloading the logs. It's pretty tedious for the user without this flag, especially on site plans with many appservers.

Closes #

## Summary

## Effect

*
*

### Dependencies and Timing

**Release**:
- [ ] When ready
- [ ] After date: $DATE

## Post Launch

### DIFF
--- a/source/scripts/collect-logs-sftp.sh.txt
+++ b/source/scripts/collect-logs-sftp.sh.txt
@@ -18,13 +18,13 @@ if [ $COLLECT_LOGS == true ]; then
 echo 'COLLECT_LOGS set to $COLLECT_LOGS. Beginning the process...'
 for app_server in $(dig +short -4 appserver.$ENV.$SITE_UUID.drush.in);
 do
-    echo "get -R logs \"app_server_$app_server\"" | sftp -o Port=2222 "$ENV.$SITE_UUID@$app_server"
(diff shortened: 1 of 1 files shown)


ccharlton:
The alternative option the script can use is `BatchMode=yes`, which also suppresses the connection prompt. The difference is, `BatchMode` will not append the server connection info to the user's `known_hosts` file. Does anyone have a reason we should adopt one over the other for customers?

namespacebrian (approved):
I'm in favor of this change.  

It prevents the customer from having to type "yes" for each appserver while the script is running.  They have no way of verifying the authenticity of a host key anyway, so typing "yes" to allow the connection is inherently blind in any scenario.

@kporras07 has expressed concern about the optics of our docs telling people to use an option that disables a security feature, but my view is that that feature doesn't actually provide any security.. it simply forces you to type "yes" to say you trust the remote host, which presumably you already decided when you pulled this script from the docs and decided to run it.   If people prefer, we could add a note saying that that option simply disables a prompt asking you if you trust the remote host, but I don't personally feel that's necessary.

PR #8802, opened 2023-12-14
Release notes MVP

stevector:


rachelwhitton:
@stevector Thinking about the URL here, rather than `/releasenotes` what do we think about `/releases` or `/release-notes`? (I think my preference is `/releases`)

stevector:
> Thinking about the URL here, rather than /releasenotes what do we think about /releases or /release-notes? (I think my preference is /releases)

Ah, yes, `/releasenotes` is an artifact of when I was troubleshooting Gatsby and was building `/release-notes` and `/releasenotes` using different commands. I'll update the file path accordingly.

I prefer `/release-notes` to `/releases` because this page is fundamentally a list of the notes. `/releases` implies a listing of software to be downloaded.

stevector:
@mel-miller I think I've hit the point where I'm out of my depth in React land and need you to take it the rest of the way. The `MenuButton` works on my local machine in Gatbsy develop mode with 
```
"callback": () => { window.location.href = `/release-notes/${category.slug}` },
```

But the fully built site is throwing errors when using the `MenuButton` https://pr-8802-documentation.appa.pantheon.site/release-notes/account-management

example:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'callback')
```

I don't know if it's better to keep debugging this JS or to convert these to proper links.

mel-miller:
@stevector -- I didn't think of this yesterday, but yeah in the built Gatsby site `window` is not defined right away. You need to run a check to see if the window is ready before executing those functions.  Here's some info about that: https://www.gatsbyjs.com/docs/debugging-html-builds/#how-to-check-if-window-is-defined

But overall, I was thinking about this earlier today and I think I'm going to add a menu item type to the existing menu button to accept regular links instead of building a whole new component. I'll keep you posted.

rachelwhitton (approved):

source/releasenotes/2024-01-23-wordpress-6-4-2-security-update.md
```diff
@@ -0,0 +1,13 @@
+---
+title: WordPress 6.4.2 security updates
+published_date: "2024-01-25"
```
rachelwhitton, line 3: Update to match filename date `published_date: "2024-01-23"`

source/releasenotes/2024-01-24-wp-cli-v2-9-0-release-now-available-on-pantheon.md
```diff
@@ -0,0 +1,21 @@
+---
+title: WP-CLI v2.9.0 release now available on Pantheon
+published_date: "2024-01-25"
```
rachelwhitton, line 3: @IngridKwok can you update this so it matches filename? 

`published_date: "2024-01-24"`

source/releasenotes/2024-01-24-wp-cli-v2-9-0-release-now-available-on-pantheon.md
```diff
@@ -0,0 +1,21 @@
...
+categories: [wordpress]
+---
+
+We're thrilled to announce the availability of WP-CLI v2.9.0 on the Pantheon platform.
+
+WP-CLI is easily accessible for any WordPress site on Pantheon through our CLI, [Terminus](https://docs.pantheon.io/terminus). Manage your WordPress installations efficiently with the power of WP-CLI at your fingertips.
```
rachelwhitton, line 9: @IngridKwok Can you replace all instances of absolute links like this from your entries today? We want relative links, so in this case it would be `[Terminus](/terminus)`

If you use absolute link for a relative page, the behavior changes. Absolute links will open in a new window which should only happen when linking outside of the docs.pantheon.io site

source/releasenotes/2024-01-25-pantheon-release-notes-a-new-year-a-new-chapter.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Pantheon release notes - A new year, a new chapter
+published_date: "2024-01-25"
+categories: [documentation, user-interface]
+---
+
+As we embark on a new year, we're excited to unveil a transformation. This January marks the introduction of **Pantheon release notes**, a rebranding of our changelog. Our aim is to elevate your platform experience by providing even more valuable insights into product updates.
+
+**What's new?**
```
rachelwhitton, line 9: @IngridKwok what do you think about using H3s instead of bolded text for each subsection of a given entry?  
Screenshot 2024-01-25 at 1 24 09 PM

PR #8807, opened 2023-12-19
Update 04-shibboleth-sso.md

Unifex:
When using this on a clean install (no vendor dir) the symlink was placed inside the already existing directory. Removing before linking did the trick for me.


rachelwhitton:
Thank you! +1

PR #8819, opened 2024-01-10
Traffic is only counts on Live

rachelwhitton:
* Document that traffic is counted for the Live environment only
* Add @pantheon-systems/platform-edge-routing to codeowners file for the traffic guide


conorbauer23:
Looks great! Thanks, Rachel.

PR #8843, opened 2024-01-24
google crawling verification

stevector:
@kyletaylored is working on a tool that needs to crawl docs.pantheon.io. This file is necessary to verify that the domain is safe to crawl.


stevector:
@rachelwhitton I think this is safe to just merge. If you agree, please do so

PR #8848, opened 2024-01-26
Exam instructions

stevector:
Bringing over some content from https://certification.pantheon.io/

PR #8862, opened 2024-02-07
Relocate Exam Instructions from Certification Registry Site to Pantheon Docs

rachelwhitton:
Replaces #8848 since there were build errors there that can only be replicated on that single branch, same code here builds without error

This PR relocates exam info from the Certification Registry Site to Pantheon Docs
It also moves the about info to a new path: `/certificaiton/about` so that the `/certification` path can be used for a new landing page, which links to the about page and the study guide and a new `/certification/exam` page which houses the relocated content from the separated certification.pantheon.io site 

- [x] Peer review from @pantheonsteve completed via Zoom. Requested one change to crosslink Cert Registry site within a CTA on the landing page

## Post Launch Todos: 
* @rachelwhitton and @stevector can cowork to add redirects for certification.pantheon.io (301 certification.pantheon.io/faq to docs.pantheon.io/certification/exam), first we need to decide implementation method (redirect module vs PHP; @pantheonsteve has no preference) - we have site admin permissions (pantheon perms and drupal admin) to do this ourselves when ready 
* Defer content removal from cert registry site to @pantheonsteve once he's back from OOO


source/data/landings.yaml
```diff
@@ -1567,3 +1567,23 @@
               url: "/object-cache/cli"
             - text: "Safely Remove Object Cache"
               url: "/object-cache/remove"
+- title: "Pantheon WebOps Certification"
```
rachelwhitton, line 1570: Link back to Cert Registry site with a CTA like "See who's certified!"

PR #8867, opened 2024-02-14
[CMSP-695] WordPress 6.5 font library documentation and behavior

jazzsequence:
## Summary

**[Release Notes](https://docs.pantheon.io/release-notes)** - Release notes for WordPress 6.5 including documentation about the Font Library
**[WordPress Configuration Guide](https://docs.pantheon.io/guides/wordpress-configurations)** - Adds a new guide for using the WP Font Library, how our filter works and how to override it, and considerations about either approach.

## Effect

The WordPress Font Library introduces a new type of asset to WordPress sites, fonts. This opens up fundamental questions about how best to make use of these assets. As an opinionated platform, our belief is that, on day one, the font library should _just work_. As such, we have [introduced code in our `pantheon-mu-plugin`](https://github.com/pantheon-systems/pantheon-mu-plugin/blob/main/inc/fonts.php) to change the default behavior of the WP Font Library such that the default location for fonts to be stored is a (writeable on Pantheon) `wp-content/uploads/fonts/` directory. This diverges from the WP core functionality of storing them in `wp-content/fonts`. 

In a version controlled platform, the underlying question to consider here is: Are fonts more like **media files** (e.g. "I don't care where they are written and I should be able to add them to my live site directly.") or are they more like **plugins and themes** (e.g. "I want to manage the fonts in use on my site and commit them to version control.")?

We have opted for the former (fonts are like media files) whilst allowing teams to restore the default functionality by overriding our choice.

This PR documents the change in WordPress core and our approach.

- [ ] Update docs with functional links to release notes.
- [ ] Update docs with functional links to Font Library docs.

### Dependencies and Timing

- [ ] Request for comments

**Release**:
- [x] When ready
- [ ] After date: March 26, 2024 (WordPress 6.5 anticipated release date)

## Post Launch


stevector:
Thanks @jazzsequence for thinking through these changes. I'll need to set aside some thinking time too!

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
```diff
@@ -0,0 +1,64 @@
...
+
+For more information, refer to [this Gutenberg issue](https://github.com/WordPress/gutenberg/issues/58696).
+
+</Alert>
+
+## Considerations
+
+The distinction between the two approaches for handling fonts depends what works best for your workflow and whether you consider your fonts to be handled more like media files or more like themes and plugins.
+
+**If you consider fonts to be more like media files**, use the default Pantheon behavior added in our `mu-plugin`. This makes it easier to _install fonts onto production sites_ but perhaps more difficult to _synchronize_ those fonts on the lower environments (test and dev).
```
pwtyler, line 59: Using the default Pantheon filter, do fonts move between environments with a "clone files" workflow as expected?

jkudish:
Giving this a +1 - I think the approach makes sense. Treat it like WP does (generic upload) by default but with the option to switch to version controlled approach if desired.

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
jazzsequence: There are a few things I need to test. That's one. I also have a couple open questions that I posted in `#core-editor` Slack: https://wordpress.slack.com/archives/C02QB2JS7/p1707937998405009

pwtyler: I smashed that subscribe button for more jazzsequence gutenberg goodness.

bensheldon:
> Are fonts more like media files ... or are they more like plugins and themes...

I'll offer that I think of a font as the latter, because I'm not sure how one would make use of an uploaded font without _also_ modifying the theme file (other than making the change with inline styles but in that case one could also reference a font asset inline as a generic upload as well).

jazzsequence:
@bensheldon 
> I'm not sure how one would make use of an uploaded font without also modifying the theme file (other than making the change with inline styles but in that case one could also reference a font asset inline as a generic upload as well).

If it helps to contextualize the feature, the Font Library is built into the Full Site Editor -- meaning that fonts and styles, while _informed by_ the theme, are _independent_ of the theme. The idea is that you would download a theme (that's compatible with FSE) and that theme might give you 4 fonts to work with. But you don't like those fonts for whatever reason, and you want to use one that you've either previously uploaded or are uploading (or installing from Google) new. That font could then be activated next year when you switch to the Twenty Twenty Five theme (or whatever).

Themes can use the Font Library to install/download their own fonts via the API it adds as well, but the Font Library/management aspect of it is kind of independent of the theme.

jazzsequence:
👀  [Doc Preview](https://pr-8867-documentation.appa.pantheon.site/guides/wordpress-configurations/wordpress-font-library)

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
jazzsequence: Here's what I'm seeing:

1. Install font on live site
2. Pull Live db down to Dev and sync files
3. The font appears in the font manager but is not _active_
4. Clicking into the font, selecting it and clicking update _activates_ it

So, to answer your question, it does appear that they are pulled down when you clone the DB and files (I suspect you would _need_ to clone the DB and files in order to get them), but there is an additional step to actually _use_ them on the lower environment you pulled down to. I will update the doc.

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
```diff
@@ -60,5 +60,16 @@ The distinction between the two approaches for handling fonts depends what works
 
 **If you consider fonts to be more like themes and plugins**, use the WordPress default behavior and the `pantheon_modify_fonts_dir` filter. This ensures that the fonts are installed across environments, but does not necessarily mean that fonts that are _physically installed_ on your lower environments (e.g. they exist in the `wp-content/fonts` directory) are recognized by WordPress. (They may still need to be added in the WordPress admin.)
 
+## Troubleshooting
+
+### "No font faces were installed" error message
+
+![No font faces were installed]((../../../images/wordpress-configurations/07-no-font-faces-installed.png))
+
+There are two reasons you might receive an error message like the above that says that no fonts were installed and asks if the parent directory is writable by the server, both assume you have _disabled_ our modification to the directory where fonts are installed and are using the WordPress core handling (uploading to `wp-content/fonts/`).
```
pwtyler, line 69:
```suggestion
If you have disabled our modification and are using WordPress's default handling (uploading to `wp-content/fonts/`), there are two reasons you might receive an error message reporting that no fonts were installed and asks if the parent directory is writable by the server.
```
Be more clear with "this only happens if you've changed the Pantheon setting"

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
```diff
@@ -0,0 +1,87 @@
...
+
+![Font is installed but not active](../../../images/wordpress-configurations/07-font-installed-not-active.png)
+
+If you've cloned your database and files from your live environment to Dev or Test and the font appears in your Font Library but is not "active" (it displays a message like "0/1 variants active"), you can click into the font, select the variant you want, and click update. 
+
+![Update Font](../../../images/wordpress-configurations/07-font-update.png)
+
+### I've uploaded fonts to my `/fonts` directory, but they aren't showing up in the Font Library
+
+Fonts need to be installed via the Font Library in the WordPress admin. This is because, in the backend, they are stored as a post type. Therefore, simply having a font in your `/fonts` directory does not necessarily mean it will be recognized by WordPress.
```
pwtyler, line 84: Isn't this the state if we're tracking fonts in version control and pushing up "as code" then? What is the next step then?

jazzsequence: Good point. I guess you'd need to install it via the admin (or in code) again? I'm still waiting for clarification in WP core Slack because the other option would be cloning the DB from a lower environment. I tried exporting a WXR and importing it into a local copy of the same site and I got a Gutenberg error when I tried to open the font library, so that's a no-go. When you upload a font into the Font Library where the files already exist, you get an error (that a font with that name already exists) but it _does_ show up in the font library. 

There's an open ticket dealing with fonts still being listed after the _files_ have been deleted: https://github.com/WordPress/gutenberg/issues/58375

There's also an open ticket for exporting fonts in a WP Export: https://github.com/WordPress/gutenberg/issues/54226 -- this one might be the missing next step, particularly if you could export _just_ fonts.

There isn't (yet) a ticket dealing with font _files_ existing but I thought I remembered seeing a discussion about it. I will open an issue since my Slack thread hasn't gotten any attention.

jazzsequence: https://github.com/WordPress/gutenberg/issues/59102

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
pwtyler: it's curious that activation status doesn't carry through db clone the way plugin activation status would.

jazzsequence:
Just posting this update as there seems to be an emerging theme in the GitHub issues around the argument of "fonts are like media files" vs "fonts are like themes/plugins" in favor of "like media files". The decision may be made for us.

See: https://github.com/WordPress/gutenberg/issues/59102#issuecomment-1948436852 and also https://github.com/WordPress/gutenberg/issues/58375, which is a tangentially related issue.

I guess this opens up the question of whether to include documentation to restore the WordPress default behavior _at all_ if the intended end goal was always to emulate the media library experience (my vote would be no, and perhaps we remove our filter to override and just let users set their own `font_dir` filter or remove our filter).

pwtyler:
I am +1 on removing our introduced filter (although it could still be removed or filtered at another priority) and instead documenting here why it is unlikely one would _want_ to override the mu's decision— unless/until we see any changes in core's process that might make the whole process more flexible.

jazzsequence:
Updated the docs to remove references to "restoring core behavior" and clarified some of our reasoning and how to override if you still _really_ wanted to. Bumping this out of draft since it seems likely that WordPress core is going to err on the side of fonts are like media.

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
jazzsequence: I described this somewhere, so I might be repeating myself, but I had a similar experience when I cloned a DB from live to Dev with theme styles. It kept everything else, but the styles that were active on Live were not active on Dev until I set them. I'm not sure what this tells me exactly, because surely they are stored in the database somewhere, but whatever the full site editor is doing with styles, it's doing the same thing with fonts. `¯\_(ツ)_/¯`

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
```diff
@@ -0,0 +1,78 @@
...
+
+For more information, refer to [this Gutenberg issue](https://github.com/WordPress/gutenberg/issues/58696).
+
+</Alert>
+
+## Considerations
+
+WordPress handles fonts more like **media files** than **plugins or themes**. This means that when fonts are added to one Pantheon site environment, they will not necessarily exist in your other WordPress environments (e.g. installing a font on Dev does not mean it will push to Test when you deploy). In WordPress, fonts have two parts, there is an underlying font post type (similar to the `attachment` post type for media files) and the physical font files themselves. Without the font post type existing in the database, WordPress has no way of knowing that a font is installed (in the same way that WordPress has no way that a particular image exists in the `/uploads` directory if it was not uploaded via the media library).
```
pwtyler, line 51:
```suggestion
WordPress handles fonts more like **media files** than **plugins or themes**. This means that when fonts are added to one Pantheon site environment, they will not necessarily exist in your other environments (e.g. installing a font on Dev does not mean it will push to Test when you deploy). In WordPress, fonts have two parts, there is a font post type (similar to the `attachment` post type for media files) and the font files themselves. Without the font post type existing in the database, WordPress has no way of knowing that a font is installed (in the same way that WordPress has no way that a particular image exists in the `/uploads` directory if it was not uploaded via the media library).
```

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
```diff
@@ -0,0 +1,78 @@
...
+
+If you intend to override the Pantheon behavior, care should be taken that you are not doing so with the assumption that committing font files to your repository will make them available automatically on your other environments.
```
pwtyler, line 53:
```suggestion
If you intend to override the Pantheon behavior, know that committing font files to your repository will not make them available automatically on your other environments.
```

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
```diff
@@ -0,0 +1,78 @@
...
+![Update Font](../../../images/wordpress-configurations/07-font-update.png)
+
+Whether or not a font is "active" is similar to whether or not a plugin or theme is active. If a font is not active, it means the font is installed and recognized by WordPress but it cannot be used on the site until you activate it.
+
+### I've uploaded fonts to my `/fonts` directory, but they aren't showing up in the Font Library
+
+Fonts need to be installed via the Font Library in the WordPress admin. This is because fonts contain data that is stored in a post type. Simply having a font in your `/fonts` directory does not mean it will be recognized by WordPress.
+
+## More Resources
+* [Font Library Gutenberg Tracking Ticket](https://github.com/WordPress/gutenberg/issues/55277)
```
pwtyler, line 78: Are there any WordPress core docs that could be linked here?

source/releasenotes/2024-03-26-wordpress-6.5.md
pwtyler, line 1: If the intention is to merge these docs before 6.5 drops (which I think is valid), this should be removed and drafted in a new PR.

source/content/guides/wordpress-configurations/07-wordpress-font-library.md
jazzsequence: No, not yet. Otherwise they would be linked.

source/releasenotes/2024-03-26-wordpress-6.5.md
jazzsequence: Because this includes release notes, my thought is that when WordPress 6.5 is released, this PR would get updated and merged shortly after it's available on Pantheon/while we're pushing it out.

jazzsequence: I've updated the PR to remove the release note so we can publish this before 6.5 drops and not wait.

jazzsequence:
Last update before I think we're good to go with this: 

I've updated the PR to remove the release note. I'll be writing up a release note separately for WordPress 6.5 so I will include the Font Library stuff in that.

Some of the logic has changed (again) since this was first drafted: now default WordPress core behavior is _if the `wp-content` directory is writeable_, fonts are installed in `wp-content/fonts` otherwise, fonts are installed in `wp-content/uploads/fonts`. This means it always works, but it creates a discrepancy between local development environments and Pantheon environments if left to the default behavior. I've updated the docs to reflect this, explain the problem, and reshaped some of the language about overriding. 

Finally, the infinite loop bug when using `wp_get_upload_dir` inside the `font_dir` filter was resolved, so I removed the note. This should be g2g once I get a ✅.

source/releasenotes/2024-03-26-wordpress-6.5.md
jazzsequence: ...and added a new release note for the Pantheon MU Plugin update.

source/releasenotes/2024-03-11-pantheon-advanced-page-cache-1.5.0.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Enhancements in Pantheon Advanced Page Cache v1.5.0
 published_date: "2024-03-11"
-categories: [wordpress]
+categories: [wordpress, plugins]
```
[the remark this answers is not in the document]
jazzsequence: Didn't realize the other PR existed. 👍

PR #8916, opened 2024-04-02
Dash creds auth method retiring soon

rachelwhitton:
Adds a warning alert to inform readers that you cannot rely on dash creds as an sftp auth method after April 30th


kelly-mcgonnigle:
approved! thanks

stevector (commented):
This change seems pretty significant. @IngridKwok my understanding is that changes of this magnitude should have accompanying release notes, yes?

PR #8927, opened 2024-04-10
Add Media Bulk Upload to Incompatible Modules

dficker:
Added Media Browser Plus, a Drupal 8/9/10 module that is probably incompatible with the platform.

Closes #

## Summary

**[Drupal Modules with Known Issues](https://docs.pantheon.io/modules-known-issues)** - Add Media Bulk Upload to modules with known issues. And add verbage to encourage customers to not even try to work around it, because it doesn't really work.

## Effect

* There are many tickets in ZenDesk about customers having issues with Media Bulk Upload module, so it likely uses tmp folders in an unsupported way. Though it seems like it may only break on test/live.
* Add it to the list of incompatible modules.

- [ ] Maybe do some testing to verify that it doesn't work?

### Dependencies and Timing

**Release**:
- [ ] When ready
- [ ] After date: $DATE

## Post Launch

### DIFF
--- a/source/content/modules-known-issues.md
+++ b/source/content/modules-known-issues.md
@@ -271,6 +271,15 @@ ___
 
 ___
 
+## [Media Bulk Upload](https://www.drupal.org/project/media_bulk_upload)
+<ReviewDate date="2024-06-24" />
+
+**Issue**: The default configuration for this module has been known to cause issues related to the "Upload location" setting, which out of the box uses the `/tmp` directory.
+
+**Solution**: Configure the "Upload location" to `public://sites/default/files/private` You must include the `public://` protocol for this solution to work.
+
+___
+
 ## [Node export webforms](https://www.drupal.org/project/node_export_webforms)
 
 **Issue**:  This module requires the use of the `tmp` directory. Refer to the [Using the tmp Directory](#using-the-tmp-directory) section below.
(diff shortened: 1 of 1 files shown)


rachelwhitton:
@ryanshoover do you have anyone on your team that could help provide technical review on this PR?

rachelwhitton:
@lcatlett can you do a technical review on this PR?

yuriybabenko:
I just installed this module on a fresh D10 site and see that the module's configuration allows specifying the filepath ("upload location") to which files will be stored during the upload process. It seems like this can be used to easily get around Pantheon's `/tmp` challenge. 

I'd suggest re-phrasing the wording in the Known Issues list to emphasize _potential_ challenges with out-of-the-box configuration (ie. user doesn't specify an "upload path"), and highlight that a configured "upload path" may be required.

Screenshot 2024-04-18 at 3 59 34 PM

rachelwhitton:
@yuriybabenko what upload location should we recommend for this module?

stevector:
> The description in the module's form suggests this would only be a temporary location in any case, and the module would move the files to the 'proper' location right after upload.

Does that work @yuriybabenko ? Are any customers using either of these modules successfully?

yuriybabenko:
@stevector - I looked at this again and tried it out. The module's configuration page isn't very detailed in terms of what's expected, but the following configuration path for "Upload location" does work on Pantheon: `public://sites/default/files/private`.  Note that the `public://` protocol part is a must - module does not work without it.

rachelwhitton:
@yuriybabenko thanks for passing along these details! I just updated the PR, can you review one last time before I merge?

yuriybabenko (approved):
Looks great!

PR #8953, opened 2024-04-19
[CMSP-993 CMSP-1055] New default cache TTL changes

jazzsequence:
## Summary
* **[Release Notes](https://docs.pantheon.io/release-notes)** - Release note for new Cache max age settings in the mu-plugin https://github.com/pantheon-systems/pantheon-mu-plugin/pull/38 and Site Health page https://github.com/pantheon-systems/pantheon-mu-plugin/pull/39
* **[Pantheon WordPress Cache Plugin Configuration](https://docs.pantheon.io/guides/wordpress-configurations/wordpress-cache-plugin)** - Updates recommendation for cache max age and adds notes about using the `pantheon_cache_default_max_age` filter
* **[Caching](https://docs.pantheon.io/guides/frontend-performance/caching)** - Updates recommendation for cache max age.

### Dependencies and Timing

- [x] https://github.com/pantheon-systems/pantheon-mu-plugin/pull/38 is merged
- [x] https://github.com/pantheon-systems/pantheon-mu-plugin/pull/39 is merged
- [x] https://github.com/pantheon-systems/pantheon-mu-plugin/pull/40 is merged
- [x] https://github.com/pantheon-systems/pantheon-mu-plugin/pull/41 is merged
- [x] 1.4.0 release of [`pantheon-mu-plugin`](https://github.com/pantheon-systems/pantheon-mu-plugin) is available

**Release**:
- [x] When ready
- [ ] After date: $DATE

## Post Launch


source/content/guides/wordpress-configurations/05-wordpress-cache-plugin.md
```diff
@@ -39,12 +39,22 @@ You can increase the default time to live value to improve the chances that a vi
 
 1. Modify the **Default Cache Time**.
 
-    You should strike a balance between freshness of content and speed. We recommend a minimum of 600 seconds. If you can increase the setting to 30 minutes (1800 seconds) or 1 hour (3600 seconds), many more requests will hit the Edge Cache. Every page served from the Edge Cache won't hit your application container's PHP workers or MySQL database, which means faster page load times and a better user experience for site visitors.
+    You should strike a balance between freshness of content and speed. We recommend a minimum of 600 seconds. If you can increase the setting to 1 hour (3600 seconds) or even 1 week (604800 seconds), many more requests will hit the Edge Cache. Every page served from the Edge Cache won't hit your application container's PHP workers or MySQL database, which means faster page load times and a better user experience for site visitors.
```
scottbuscemi, line 42: The premise of this first sentence is misguided. With PAPC, content should always be fresh since it's cleared. 

Let's explain that, and recommend 1 week. No need to mention 600 seconds or 1 hour.

source/releasenotes/2024-04-xx-pantheon-page-cache-default-ttl.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Default Cache TTL bumped to one week
+published_date: "2024-04-xx"
+categories: [wordpress, plugins]
+---
+
+With the latest [1.4.0 release](https://github.com/pantheon-systems/pantheon-mu-plugin/releases) of the Pantheon MU Plugin, we have updated the default cache TTL to one week. Previously, the default value was `600` (10 minutes). This change is intended to improve the performance of your site by reducing the number of cache clears that are run automatically.
```
scottbuscemi, line 7: With the latest [1.4.0 release](https://github.com/pantheon-systems/pantheon-mu-plugin/releases) of the Pantheon MU Plugin, we have updated the default cache TTL to one week. Previously, the default value was 10 minutes. This change is intended to improve the performance of your site by reducing the number of cache clears that are run automatically.

scottbuscemi: We should also note here that the Pantheon Advanced Page Cache plugin auto-clears content, which is why we're increasing the default cache TTL.

Also should add a note that this change does not the TTL for existing sites.

scottbuscemi (changes requested):

rachelwhitton:
There are a few more references to the previous default, for example: https://github.com/pantheon-systems/documentation/pull/8953/files#diff-1cd1e29f23a107b75b778ae3d0d9c4d1b26471ebad093a28bed851877ffdb9d4R18 

And a few sections on this page: https://docs.pantheon.io/guides/wordpress-pantheon/plugins#pantheon-page-cache

source/releasenotes/2024-04-xx-pantheon-mu-plugin-1-4-0-update.md
```diff
@@ -0,0 +1,15 @@
...
+
+You can modify this new default value using a filter built into the MU Plugin. For more information refer to our [WordPress Cache Plugin Configuration](/guides/wordpress-configurations/wordpress-cache-plugin#override-the-default-ttl).
+
+Included in this release are some changes to the Site Health page in the WordPress admin dashboard. We have removed recommended Site Health tests that were previously recommending things that were not possible on Pantheon's read-only filesystem and we have added new tests that check your Redis object cache configuration. If Redis is not enabled for a site, you will see a recommendation in the Site Health page, and you will see new recommendations if Redis is active but you are not using an object caching plugin like [Object Cache Pro](/object-cache/wordpress).
+
+Finally, 1.4.0 adds a filter that allows you to skip cache control headers for specific requests. This is useful if you have a specific page that you do not want to cache. For more information refer to our [GitHub repository documentation](https://github.com/pantheon-systems/pantheon-mu-plugin?tab=readme-ov-file#hooks).
```
stevector, line 15:
```suggestion
Finally, 1.4.0 adds a filter that allows you to skip cache control headers for specific requests. This is useful if you have a specific page that you do not want to cache. For more information refer to our [GitHub repository documentation](https://github.com/pantheon-systems/pantheon-mu-plugin?tab=readme-ov-file#hooks). Special thanks to [Eric Caron](https://github.com/ecaron) for [contributing the suggestion](https://github.com/pantheon-systems/pantheon-mu-plugin/issues/37) that became the 'pantheon_skip_cache_control' filter.
```

source/releasenotes/2024-04-xx-pantheon-mu-plugin-1-4-0-update.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Pantheon MU Plugin v1.4.0 update
+published_date: "2024-04-xx"
+categories: [wordpress, plugins]
+---
+
+With the latest [1.4.0 release](https://github.com/pantheon-systems/pantheon-mu-plugin/releases) of the Pantheon MU Plugin, we have updated the default cache TTL to one week. Previously, the default value was 10 minutes. This change is intended to improve the performance of your site by reducing the number of cache clears that are run automatically. The [Pantheon Advanced Page Cache](https://wordpress.org/plugins/pantheon-advanced-page-cache) plugin auto-clears content when pages are published or updated which is why we are increasing the default TTL. We highly recommend using the Pantheon Advanced Page Cache plugin to ensure that your site is performing optimally.
```
stevector, line 7:
```suggestion
With the latest [1.4.0 release](https://github.com/pantheon-systems/pantheon-mu-plugin/releases) of the Pantheon MU Plugin, we have updated the default cache max-age used to one week. Previously, the default value was 10 minutes. This change is intended to improve the performance by increasingly the likelihood that an unexpired cached copy of a page is present within the CDN. Often such increases in max-age settings carry the risk that site visitors see stale content that should be replaced by newly published posts. Pantheon greatly mitigates this risk with The [Pantheon Advanced Page Cache](https://wordpress.org/plugins/pantheon-advanced-page-cache) plugin. This plugin granularly clears cached responses using post IDs and other metadata. We highly recommend using the Pantheon Advanced Page Cache plugin to ensure that your site is performing optimally.
```

stevector: We should say "max-age" and not "TTL" when the header value is literally "max-age".

Part of the reason I think this pedantry is valuable is that I have encountered a lot of customers who think this value is a the exact amount of time the response will be stored in the cache. "Time To Live" encourages that interpretation. It implies that this is the amount of time the response will live in the cache.

But that's not what it is. It is the _maximum_ amount of time the response should be kept. It is not a guarantee that the the response will be kept that long.

source/content/guides/wordpress-configurations/05-wordpress-cache-plugin.md
```diff
@@ -39,12 +39,22 @@ You can increase the default time to live value to improve the chances that a vi
...
 
-    You should strike a balance between freshness of content and speed. We recommend a minimum of 600 seconds. If you can increase the setting to 30 minutes (1800 seconds) or 1 hour (3600 seconds), many more requests will hit the Edge Cache. Every page served from the Edge Cache won't hit your application container's PHP workers or MySQL database, which means faster page load times and a better user experience for site visitors.
+    When [Pantheon Advanced Page Cache](https://wordpress.org/plugins/pantheon-advanced-page-cache) is active, your site content should always be fresh and served quickly from Pantheon's Edge Cache. This is why we recommend caching content for a minimum of 1 week. Every page served from the Edge Cache won't hit your application container's PHP workers or MySQL database, which means faster page load times and a better user experience for site visitors.
 
 1. Click **Save Changes**.
 
 ![WordPress Pantheon Cache Plugin settings](../../../images/WordPress_Pantheon-Cache-Settings.png)
 
+### Override the default TTL
+
```
stevector, line 49: we should use `max-age`, not `ttl` in this filter

stevector (dismissed):
@jazzsequence @scottbuscemi I made some comments here on "TTL" vs. "max-age" but I'm not sure if that's a linguistic battle worth fighting since we have TTL all over the place. What do you think?

source/content/guides/wordpress-configurations/05-wordpress-cache-plugin.md
stevector: Or should we? https://github.com/pantheon-systems/documentation/pull/8953#pullrequestreview-2020329072

source/releasenotes/2024-04-xx-pantheon-mu-plugin-1-4-0-update.md
jazzsequence:
```suggestion
Finally, 1.4.0 adds a filter that allows you to skip cache control headers for specific requests. This is useful if you have a specific page that you do not want to cache. For more information refer to our [GitHub repository documentation](https://github.com/pantheon-systems/pantheon-mu-plugin?tab=readme-ov-file#hooks). Special thanks to [Eric Caron](https://github.com/ecaron) for [contributing the suggestion](https://github.com/pantheon-systems/pantheon-mu-plugin/issues/37) that became the `pantheon_skip_cache_control` filter.
```

jazzsequence: @stevector I copied your language and just added backticks around the filter name.

source/releasenotes/2024-04-xx-pantheon-mu-plugin-1-4-0-update.md
jazzsequence: The setting in the WordPress admin is "TTL", though. We can change that, but having two different phrases that mean the same thing would be just as confusing.

@scottbuscemi up to you whether we want to change the description in WordPress admin.

jazzsequence:
@rachelwhitton The original code snippet isn't _wrong_. There is a, perhaps small but significant difference in the approach between the two methods. The new filter returns _no_ `cache-control` headers at all if you use `add_filter( 'pantheon_skip_cache_control', '__return_true' )` on specific pages/paths. The documentation in the example sends the cache control headers with `max-age=0`. I think the end result is likely the same and this might be a good opportunity to show off the new filter in documentation, but it does not remove the complicated part (IMO the regex) and using the filter is not necessarily _more correct_ (for my money, I might rather send `max-age=0` than _nothing_).

I guess my question is, is it worth adding _more_ code snippets to suggest using the filter as an _alternative_ to what's currently there, but not remove it?

source/releasenotes/2024-04-xx-pantheon-mu-plugin-1-4-0-update.md
stevector: > We can change that, but having two different phrases that mean the same thing would be just as confusing.

Yes, using both without explaining would be even more confusing.

After reading https://en.wikipedia.org/wiki/Time_to_live I am willing to drop this concern. Our usage does fit with with the meaning of TTL and I don't think it's worth the effort to change our terminology across the board. While I think people would be less likely to get confused if we used "max-age" everywhere instead of "TTL" there still would be some misinterpretation, changing UI copy creates swirl/confusion, etc

stevector:
> I think the end result is likely the same

In some automated testing of caching a few years ago we found that it was possible to get cache hits from responses with a  `max-age=0` header. Right now I'm struggling to remember why I was in a position to discover that. But yeah, setting `max-age=0` is not enough without `no-cache` (and `must-revalidate`??)

> The original code snippet isn't wrong.

This strikes me as a situation where the [Diátaxis documentation framework](https://diataxis.fr/) should help. Is the documentation here a tutorial, explanation, guide, or reference? I think this "bypassing cache" page is closest to a "How-To Guide." My expectation is that people come to this page in the context of attempting to accomplish a specific work task and they need the necessary steps to accomplish the task of bypassing cache for a real website (different from a general tutorial that's teaching caching concepts).

As a "How to Guide" it is acceptable to include "if" statements.
```
* If you want to ensure that the CDN does not cache the response do XYZ
* If you want to set a very specific cache header on your own and you want to Pantheon's plugin to set no header, the do ABC
```

> I guess my question is, is it worth adding more code snippets to suggest using the filter as an alternative to what's currently there, but not remove it?

We run the risk of supplying the reader with too many options and not enough guidance on how to pick an option.

source/releasenotes/2024-04-xx-pantheon-mu-plugin-1-4-0-update.md
jazzsequence: We've got an internal ticket to change "ttl" to "max age" in the mu-plugin as part of the 1.4.0 update to the mu-plugin.

scottbuscemi:
@stevector Any chance for a review so we can merge this in?

stevector (approved):
Woo! Thanks for the updates @scottbuscemi and @jazzsequence.

Shall I merge now?

scottbuscemi:
@jazzsequence Should we wait until the next WP release? Otherwise the docs won't match customer experience.

stevector:
> Should we wait until the next WP release? Otherwise the docs won't match customer experience.

@scottbuscemi Probably? Could we take a few minutes to diagram out the interrelated parts and when we take which actions? That could be valuable internally and maybe more so externally. Like, I could interview you and @jazzsequence about the moving pieces in play (like https://www.youtube.com/watch?v=6pDMf9e3ZNM) There's value and trust-building in making visible how we handle these sorts of interrelated updates.

scottbuscemi:
Let's indeed hold off until the next WP release since that's when it'll be update-able for most customers.

PR #8964, opened 2024-04-25
WIP - DRYing out mdx component

stevector:
We have a lot of MDX embeds that reused overlapping sets of "shortcodes". I think the codebase will be more comprehensible if we abstract it out.


stevector:
Nevermind! How did I forget that I already made this idea in https://github.com/pantheon-systems/documentation/commits/main/src/components/mdxWrapper.js

PR #8966, opened 2024-04-25
WIP - Mdxwrapper usage

stevector:


src/templates/resource.js
```diff
@@ -1,57 +1,11 @@
 import React from "react"
```
stevector, line 1: I don't know if or where resource.js is used. that uncertainty should block merging.

stevector (commented):
I don't know if or where resource.js is used. that uncertainty should block merging.

src/templates/video.js
```diff
@@ -1,55 +1,11 @@
 import React from "react"
```
stevector, line 1: I'm not sure if or where this template is used. Don't merge prior to figuring that out

stevector:
Replaced with https://github.com/pantheon-systems/documentation/pull/8968

PR #8967, opened 2024-04-26
Delete src/templates/resource.js

stevector:
It appears this file is not used at all, but it does through a warning  during the Gatsby build process.


stevector:
@mel-miller This PR is not at all urgent. Just some clean up that should get your review.

mel-miller (approved):
Yeah, I'm pretty sure we do not need that file.

stevector:
Thanks @mel-miller! Assigning to @rachelwhitton who may know something else about the history of the file, or just merge.

PR #8969, opened 2024-04-26
[BUGS-7678] WP config HTTP_HOST documentation updates

jazzsequence:
## Summary

* **[WordPress Known Issues](https://docs.pantheon.io/wordpress-known-issues)** - Adds `Undefined array key HTTP_HOST` warning as a known issue
* **[WordPress Multisite - Configure](https://docs.pantheon.io/guides/multisite/config)** - Updates recommended configuration for WP multisite to use `PANTHEON_HOSTNAME`
* **[Configure Your wp-config.php File](https://docs.pantheon.io/guides/php/04-wp-config-php.md)** - adds notes about `PANTHEON_HOSTNAME` constant
* **[Release Notes](https://docs.pantheon.io/release-notes)** - Release note for the change made in the Pantheon MU Plugin

- [x] https://github.com/pantheon-systems/pantheon-mu-plugin/pull/42 is merged
- [x] https://github.com/pantheon-systems/WordPress/pull/382 is merged
- [x] https://github.com/pantheon-systems/wordpress-composer-managed/pull/119
- [x] Version 1.4.1 of the mu-plugin is released
- [ ] A new version of WordPress is released (holding on pushing docs updates until the change goes out to users on the default (non-Composer) WordPress upstream along with the next release.

**Release**:
- [ ] When ready
- [ ] After date: $DATE

## Post Launch

### DIFF
--- a/source/content/guides/multisite/03-config.md
+++ b/source/content/guides/multisite/03-config.md
@@ -167,7 +167,7 @@ Complete the steps below after spinning up a new WPMS site from the correct Cust
     }
   }
   define( 'MULTISITE', true );
-  define( 'SUBDOMAIN_INSTALL', <?php echo $subdomain_install ? 'true' : 'false'; ?> );
+  define( 'SUBDOMAIN_INSTALL', false ); // Set this to TRUE for Subdomains
   define( 'DOMAIN_CURRENT_SITE', $hostname );
   define( 'PATH_CURRENT_SITE', '/' );
   define( 'SITE_ID_CURRENT_SITE', 1 );
--- a/source/content/guides/multisite/03-config.md
+++ b/source/content/guides/multisite/03-config.md
@@ -71,6 +71,7 @@ Make sure [Terminus](/terminus) is installed and [authenticated](/terminus/insta
 1. Locate the configuration added by WP-CLI, and *modify* the line that sets `DOMAIN_CURRENT_SITE` to a hardcoded URL. We have provided a constant in `wp-config-pantheon.php`, `PANTHEON_HOSTNAME` that defaults to a dynamic URL (`$_SERVER['HTTP_HOST']`, when available) for web requests, while providing a fallback for non-web requests (notably workflows like search and replace) that do not have a `$_SERVER['HTTP_HOST']` value.
   
   ```php:title=wp-config.php
+  define( 'WP_ALLOW_MULTISITE', true );
   define( 'MULTISITE', true );
   define( 'SUBDOMAIN_INSTALL', false ); // Set this to TRUE for Subdomain installs.
   define( 'DOMAIN_CURRENT_SITE', PANTHEON_HOSTNAME );
@@ -120,6 +121,7 @@ Complete the steps below after spinning up a new WPMS site from the correct Cust
 1. Locate the `/* That's all, stop editing! Happy Pressing. */` line, and add the following code above this line to enable the WPMS configuration.
 
   ```php:title=wp-config.php
+  define( 'WP_ALLOW_MULTISITE', true );
   define( 'MULTISITE', true );
   define( 'SUBDOMAIN_INSTALL', false ); // Set this to TRUE for Subdomains
   define( 'DOMAIN_CURRENT_SITE', PANTHEON_HOSTNAME );
--- a/source/content/guides/multisite/03-config.md
+++ b/source/content/guides/multisite/03-config.md
@@ -68,20 +68,43 @@ Make sure [Terminus](/terminus) is installed and [authenticated](/terminus/insta
 
 1. Open the `code` folder in your SFTP client, and download your site's `wp-config.php` file.
 
-1. Locate the configuration added by WP-CLI, and *modify* the line that sets `DOMAIN_CURRENT_SITE` from a hardcoded URL to a dynamic URL `$_SERVER['HTTP_HOST']`. This automatically detects the URL in each environment. You must replace this variable. For example:
-
+1. Locate the configuration added by WP-CLI, and *modify* the line that sets `DOMAIN_CURRENT_SITE` to a hardcoded URL. You will want to default to a dynamic URL (`$_SERVER['HTTP_HOST']`), however some environments on Pantheon (notably workflows that run containerized versions of WP-CLI like search and replace) will not have a `$_SERVER` global. For this reason, we recommend using the following configuration that provides fallbacks if `$_SERVER['HTTP_HOST']` is unavailable.
+  
   ```php:title=wp-config.php
-  define( 'WP_ALLOW_MULTISITE', true );
+  $hostname = 'www.yourdomain.com'; // The domain of the network. Use as a fallback if $_SERVER['HTTP_HOST'] is not available.
+  if ( !empty( $_ENV['PANTHEON_ENVIRONMENT'] )) {
+    $site_name = $_ENV['PANTHEON_SITE_NAME'];
+    // Override $hostname value as needed.
+    switch ( $_ENV['PANTHEON_ENVIRONMENT'] ) {
+      case 'live':
+        // Fall back to the default $hostname if $_SERVER['HTTP_HOST'] is not available.
+        $hostname = isset( $_SERVER['HTTP_HOST'] ) ? $_SERVER['HTTP_HOST'] : $hostname;
+        break;
+      case 'test':
+        $hostname = isset( $_SERVER['HTTP_HOST'] ) ? $_SERVER['HTTP_HOST'] : 'test-' . $site_name . '.pantheonsite.io';
+        break;
+      case 'dev':
+        $hostname = isset( $_SERVER['HTTP_HOST'] ) ? $_SERVER['HTTP_HOST'] : 'dev-' . $site_name . '.pantheonsite.io';
+        break;
+      case 'lando':
+        $hostname = $site_name . '.lndo.site';
+        break;
+      default:
+        $hostname = $_ENV['PANTHEON_ENVIRONMENT'] . '-' . $site_name . '.pantheonsite.io';
+        break;
+    }
+  }
   define( 'MULTISITE', true );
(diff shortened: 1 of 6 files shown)


source/content/guides/multisite/03-config.md
```diff
@@ -68,20 +68,43 @@ Make sure [Terminus](/terminus) is installed and [authenticated](/terminus/insta
 
 1. Open the `code` folder in your SFTP client, and download your site's `wp-config.php` file.
 
-1. Locate the configuration added by WP-CLI, and *modify* the line that sets `DOMAIN_CURRENT_SITE` from a hardcoded URL to a dynamic URL `$_SERVER['HTTP_HOST']`. This automatically detects the URL in each environment. You must replace this variable. For example:
-
+1. Locate the configuration added by WP-CLI, and *modify* the line that sets `DOMAIN_CURRENT_SITE` to a hardcoded URL. You will want to default to a dynamic URL (`$_SERVER['HTTP_HOST']`), however some environments on Pantheon (notably workflows that run containerized versions of WP-CLI like search and replace) will not have a `$_SERVER` global. For this reason, we recommend using the following configuration that provides fallbacks if `$_SERVER['HTTP_HOST']` is unavailable.
```
stevector, line 71: > containerized versions

I think it'd be clearer and more accurate to say non-web requests. web requests also hit WordPress running in a container.

> will not have a `$_SERVER` global
Is that true? Or are they just missing `$_SERVER['HTTP_HOST']`

source/content/guides/multisite/03-config.md
```diff
@@ -68,20 +68,43 @@ Make sure [Terminus](/terminus) is installed and [authenticated](/terminus/insta
...
+  
   ```php:title=wp-config.php
-  define( 'WP_ALLOW_MULTISITE', true );
+  $hostname = 'www.yourdomain.com'; // The domain of the network. Use as a fallback if $_SERVER['HTTP_HOST'] is not available.
+  if ( !empty( $_ENV['PANTHEON_ENVIRONMENT'] )) {
+    $site_name = $_ENV['PANTHEON_SITE_NAME'];
```
stevector, line 76: In PR about checking that array keys exist before using them, check this key too :)

stevector (changes requested):
@jazzsequence could we do something in a plugin that made copy/pasting a big `switch` statement unnecessary? That might be too big a request for this PR.

source/content/guides/multisite/03-config.md
```diff
@@ -120,9 +143,32 @@ Complete the steps below after spinning up a new WPMS site from the correct Cust
...
+        break;
+      default:
+        $hostname = $_ENV['PANTHEON_ENVIRONMENT'] . '-' . $site_name . '.pantheonsite.io';
+        break;
+    }
+  }
   define( 'MULTISITE', true );
-  define( 'SUBDOMAIN_INSTALL', false ); // Set this to TRUE for Subdomains
-  define( 'DOMAIN_CURRENT_SITE', $_SERVER['HTTP_HOST'] );
+  define( 'SUBDOMAIN_INSTALL', <?php echo $subdomain_install ? 'true' : 'false'; ?> );
```
pwtyler, line 170: no sure I follow what is going on here? What is this inline conditional?

pwtyler:
>  could we do something in a plugin that made copy/pasting a big switch statement unnecessary? That might be too big a request for this PR.

I would love to see a version of this automated as part of the site launch workflow for wpms sites

source/content/guides/multisite/03-config.md
jazzsequence: I copy/pasted a large part of this from the WPCM repository. This may have been from that and it might not be a thing we want/need here.

source/content/guides/multisite/03-config.md
jazzsequence: If we are on a `PANTHEON_ENVIRONMENT` (e.g. `$_ENV['PANTHEON_ENVIRONMENT']` is not empty), we can make a lot of assumptions. One of those assumptions is that `PANTHEON_SITE_NAME` would _also_ exist. The issue is not the `isset` itself, the issue is that in some environments (predominantly unified job runner workflows that do not actually spin up a live server) the `$_SERVER` superglobal is not defined. If `$_ENV['PANTHEON_ENVIRONMENT']` is defined, we don't need to worry about that.

source/content/guides/multisite/03-config.md
jazzsequence: > > will not have a $_SERVER global
> Is that true? Or are they just missing $_SERVER['HTTP_HOST']

I haven't verified 100% (because I don't know that there's a way to jump into those environments and run commands outside of running workflows, so we'd basically need to break something existing to test or write a lot of new Ygg internals code to make something new for testing), but I'm relatively sure this is accurate. However, we can be more specific and _just_ say that HTTP_HOST is not set, because that's the part that really matters.

jazzsequence:
@stevector 
> could we do something in a plugin that made copy/pasting a big switch statement unnecessary? That might be too big a request for this PR.

the fundamental ask here is that there needs to be a fallback domain that is customer-specific and not reliant on server variables. We could simplify this somewhat, but there's every change that it will break things because we have different environments that need different rules (and using `$_SERVER['HTTP_HOST']` _was_ the way around that). We can't add something like a single filter or something that could be used in code and remove all the other stuff because this is before WordPress execution even starts, so there's nothing to hook into. 

Ultimately, the configuration file (be it `config/application.php` or WP Composer Managed or `wp-config.php`) is customer code that we _can't_ make changes to without breaking their sites. The change _has_ to come from the user, at least as far as the fallback hostname is concerned. As @pwtyler mentions, we can build something new into a workflow for creating WordPress multisites _specifically_, but that would likely be a whole new site creation workflow (e.g. Create New Site -> WordPress -> WordPress Multisite) that doesn't exist today.

source/content/guides/multisite/03-config.md
jazzsequence: yeah this was copy/pasta from the mu-plugin where we're detecting something from a previous step. I've removed.

source/content/guides/multisite/03-config.md
jazzsequence: I've updated to be clearer about _just_ talking about `$_SERVER['HTTP_HOST']` (and not `$_SERVER` more generally) and updated the language to use "web requests"/"non-web requests".

stevector:
> Ultimately, the configuration file (be it config/application.php or WP Composer Managed or wp-config.php) is customer code that we can't make changes to without breaking their sites.

What about hiding some of this complexity in `wp-config-pantheon.php` ? Even declaring a function there (or in prepend.php) that simplified the setting of these variables would be better than this much copy/pasting.

jazzsequence:
> What about hiding some of this complexity in wp-config-pantheon.php ? Even declaring a function there (or in prepend.php) that simplified the setting of these variables would be better than this much copy/pasting.

The problem is this _is_ in the mu-plugin. Specifically, it's in the [instructions for setting up WordPress multisite that displays in the admin when you are setting up a multisite](https://github.com/pantheon-systems/pantheon-mu-plugin/blob/main/inc/network/includes-network.php#L462-L482).  We are [making a change in the MU Plugin already with https://github.com/pantheon-systems/pantheon-mu-plugin/pull/42. This PR is updating our documentation so it matches the guidance that's already in the mu-plugin. This is something that people are _already_ copy/pasting.

jazzsequence:
MU plugins are loaded _after_ `wp-config.php`, so there's nothing we could put into the MU plugin that customer code could expect (e.g. if we wanted to define a constant _in the MU plugin_ and have customers use that as a fallback in their `wp-config.php`).

HOWEVER, it _is_ conceivable that we could define a constant in `wp-config-pantheon.php` and `config/application-pantheon.php` (for Composer Managed) that customers could override in their `wp-config.php`. This is code that I'm playing with...

**Pantheon-maintained code** (exists in the WordPress upstream)
```php
	if ( ! defined( 'PANTHEON_HOSTNAME' ) && ! isset( $_SERVER['HTTP_HOST'] ) ) {
		$site_name = $_ENV['PANTHEON_SITE_NAME'];
		$hostname = $_ENV['PANTHEON_ENVIRONMENT'] . "-$site_name.pantheonsite.io";
		define( 'PANTHEON_HOSTNAME', $hostname );
	}
```

This would be the customer code (default):
```php
define('DOMAIN_CURRENT_SITE', PANTHEON_HOSTNAME);
```

However, they could customize it like this if they wanted:
```php
define( 'PANTHEON_HOSTNAME', [
	'dev' => 'pantheondev.chrisreynolds.io',
	'test' => 'pantheontest.chrisreynolds.io',
	'live' => 'pantheon.chrisreynolds.io',	
] );
$hostname = isset( $_SERVER['HTTP_HOST'] ) ? $_SERVER['HTTP_HOST'] : PANTHEON_HOSTNAME[ $_ENV['PANTHEON_ENVIRONMENT'] ];
define('MULTISITE', true);
define('SUBDOMAIN_INSTALL', false);
define('DOMAIN_CURRENT_SITE', $hostname);
define('PATH_CURRENT_SITE', '/');
define('SITE_ID_CURRENT_SITE', 1);
define('BLOG_ID_CURRENT_SITE', 1);
```

This is a change in the approach because we'd need to make upstream changes, so this is a @scottbuscemi Product decision.

jazzsequence:
Discussed in PBR. We're going to try the alternative approach I outline in my last comment. tl:dr; this will mean essentially one (optional) change to the customer's `wp-config.php` file (or `config/application.php` if using Bedrock/Composer) for multisite and we can remove the switches entirely. I need to do additional testing to ensure it works for all the things, but I will move this PR back to draft while I'm working this angle.

stevector:
Thanks @jazzsequence!

rwagner00 (approved):

jazzsequence:
We've merged everything, but we're going to hold off on releasing docs until the next WordPress update goes out and pushes the https://github.com/pantheon-systems/WordPress upstream update out with it.

Issue #8961, opened 2024-04-24
Domain Validation - Procedure Changes on various pages


eabquina:
Re: [Relaunch Existing Pantheon Site](https://docs.pantheon.iorelaunch)

Priority: High

## Issue Description:
The following pages now need to be updated to consider the Domain Validation procedures

https://docs.pantheon.io/relaunch#relaunch-procedure
Step 4 where we need to add TXT record first before DNS change

https://docs.pantheon.io/guides/global-cdn/https#provision-https
No longer correct, not automatic unless exempted or Support triggered

https://docs.pantheon.io/relaunch#expected-disruption
Add copy that the disruption is dependent to the TXT record domain validation procedure

## Suggested Resolution
Add Domain Validation considerations on relevant sections

rachelwhitton:
https://docs.pantheon.io/relaunch#how-does-the-enforcement-of-domain-verification-impact-the-established-relaunch-procedure

@eabquina thanks for this issue! The FAQ above indicates that relaunch procedure is effectively exempt from validation, during testing early on I recall that to be true but has the behavior recently changed? 

@conorbauer23 can you review and weigh in?

rachelwhitton:
Additional pages that need updates: 


https://docs.pantheon.io/guides/domains/custom-domains
https://docs.pantheon.io/guides/launch/domains/
https://docs.pantheon.io/guides/launch/configure-dns/
https://docs.pantheon.io/guides/global-cdn/https
https://docs.pantheon.io/guides/global-cdn/global-cdn-faq
https://docs.pantheon.io/cloudflare

PR #8999, opened 2024-05-16
Relaunch procedure now requires domain validation

rachelwhitton:
The relaunch procedure is no longer "effectively exempt" from domain validation, this PR updates the procedure doc so that it instructs users to verify ownership and sets the expectation that the disruption (404 unknown site error) will resolve only once domain validation is completed 

This PR partially addresses #8961 but a more thorough audit is required to fully resolve the issue


conorbauer23 (approved):
Looks good!

PR #8805, opened 2023-12-18
WIP - Adding Playwright Tests

stevector:
Replacing https://github.com/pantheon-systems/documentation/pull/8804

PR #9007, opened 2024-05-20
Visual Regression with Playwright

stevector:
Replacing https://github.com/pantheon-systems/documentation/pull/8805 with a narrower usecase for Playwright


stevector:
@rachelwhitton can you review now?

PR #9008, opened 2024-05-20
Delete redundant Drush page

stevector:
The content on https://docs.pantheon.io/drush seems to thin to be worth keeping. It's a very short tutorial and I think for the purposes of a tutorial, https://docs.pantheon.io/guides/drush/drupal-commandline is better.


stevector:
I'm marking as a draft because there may be other clean up around Drush to do together.

rachelwhitton:
Can we move this to #9010 ? With a note in the description about where we need to redirect after deletion

rachelwhitton:
Rachel to move this into existing PR and create a separate issue to audit SST video content `layout: video`

PR #9046, opened 2024-06-13
Add note on site name length to Add Site page

morganestes:
Since there is a hard limit of 51 characters in the site name, we should let folks know ahead of time about it instead of having to guess.

Closes #

## Summary

**[Add a Site](https://docs.pantheon.io/guides/getstarted/addsite)** - Adds a callout note about site name length restrictions in step 3.

**Release**:
- [x] When ready

## Post Launch


stevector:
Thanks @morganestes and @dylanbaumann! I've put some edits in https://github.com/pantheon-systems/documentation/pull/9053

PR #9081, opened 2024-07-10
checking for search js before adding it

stevector:
While working locally with `npm run develop` I regularly get this error while clicking around within guides.

Screenshot 2024-07-09 at 9 09 25 PM

As far as I can tell, the error is resulting from running `document.body.appendChild(script)` repeatedly with the same value. While there might be a more elegant or systematic fix, this PR just checks to see if there already is a script tag adding the external JS file before trying to add it.

@rachelwhitton can you confirm that you also have seen this error while working locally and confirm that this the error disappears with this PR? If so, we should ask Mel to review/approve.

### DIFF
--- a/src/components/addSearch.js
+++ b/src/components/addSearch.js
@@ -21,15 +21,17 @@ class AddSearch extends React.Component {
       }
     }
     const script = document.createElement("script")
-    script.setAttribute(
-      "src",
-      `https://cdn.addsearch.com/v5/addsearch-ui.min.js?key=a7b957b7a8f57f4cc544c54f289611c6&id=search_widget${
-        resultPage.includes("search") ? "&type=resultpage" : ""
-      }`
-    )
-    script.setAttribute("defer", true)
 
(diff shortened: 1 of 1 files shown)


rachelwhitton:
@stevector can you share a recording of how to replicate this issue? I started by running `npm run develop` on the main branch locally, and then I went to a few random guides and clicked around using the lefthand nav and I haven't been able to replicate the error shared above in your PR description

stevector:
hmmm, maybe we should troubleshoot on our Zoom later. Just clicking around the site is how I get this error.

jazzsequence:
I wasn't able to replicate just "clicking around", but I did reproduce the issue running in codespaces with `npm run develop` when I tried to use the search.
Screenshot 2024-08-15 at 11 06 58 AM

jazzsequence:
The original error screenshotted also references search. Maybe it's search specific? Our docs search does seem to periodically not work, this seems related.

Issue #8636, opened 2023-08-08
"Without regex" example uses regex


alisonjo315:
As far as I can tell, line 55 expects a regex (`$matches[1]`):
https://github.com/pantheon-systems/documentation/blob/0a2ce819ff3dcee974a3ed24f95c2891592fb06d/source/content/guides/redirect/07-php.md?plain=1#L50-L57

I'm not sure what the proper code is, the only other version I've been able to find is for when you also need to force HTTPS in settings.php, not just the primary domain (i.e. https://docs.pantheon.io/guides/domains/primary-domain#redirect-with-php )

Is it as simple as replacing `$matches[1]` with `mysite` ? -- i.e. line 55 would be:
```
    header('Location: https://www.mysite.com/');
```

dficker:
I agree. I don't understand what this section of this page is supposed to do:

https://docs.pantheon.io/guides/redirect/php#redirect-traffic-to-your-live-sitedomain-from-sitepantheonio

As written, these snippets will redirect all requests that match a certain criteria to the home page. These would be more helpful if it kept the path string like the code at the top of this page? Very confusing.

I've never noticed this part of the page before and I don't understand how these got here and what they're for. They don't seem useful.

alisonjo315:
Yeah! -- after I posted this issue, I realized that, too -- I don't know what the value is, of those code snippets.  (The "without regex" example also does have a mistake, but yeah also, "who is this for?")

alisonjo315:
Here's what we have on a site that happens to have newrelic, too (in `settings.php`) -- a bit more concise than the longer code block with the force HTTPS bit.  The newrelic part is obviously "optional" (and not relevant to all sites). but I'm including it for the sake of sharing, because this is literally the code we have working well on a real-live site (minus the fake domain).

```php
if (isset($_SERVER['PANTHEON_ENVIRONMENT']) && php_sapi_name() != 'cli' && $_ENV['PANTHEON_ENVIRONMENT'] === 'live') {
  $primary_domain = 'mysite.com';
  if ($_SERVER['HTTP_HOST'] != $primary_domain) {
    # Name transaction "redirect" in New Relic for improved reporting (optional)
    if (extension_loaded('newrelic')) {
      newrelic_name_transaction("redirect");
    }
    header('HTTP/1.0 301 Moved Permanently');
    header('Location: https://'. $primary_domain . $_SERVER['REQUEST_URI']);
    exit();
  }
}
```

[...]

PR #9086, opened 2024-07-10
simplifying PHP redirects section

stevector:
Closing #8636. The short examples on https://docs.pantheon.io/guides/redirect/php feel duplicative of the CMS-specific examples higher on the page and the more detailed examples on the next page. I think we should just remove them.

### DIFF
--- a/source/content/guides/redirect/07-php.md
+++ b/source/content/guides/redirect/07-php.md
@@ -18,49 +18,9 @@ permalink: docs/guides/redirect/php
 
 If your site configuration prevents you from setting the primary domain from the platform level, you can use PHP redirects. However, redirecting the platform domain will break the screenshot of your site in the Personal Workspace, and might complicate troubleshooting for our [Support](/guides/support/contact-support/) team.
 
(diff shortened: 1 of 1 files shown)


source/content/guides/redirect/07-php.md
```diff
@@ -18,49 +18,12 @@ permalink: docs/guides/redirect/php
...
-```php:title=wp-config.php%20or%20settings.php
- if($_SERVER['HTTP_HOST'] == 'old.domain.com') {
-    header('HTTP/1.0 301 Moved Permanently');
-    header('Location: https://new.domain.com/');
-  }
-```
-
-### Using RegEx
+## Redirect Platform Domains (`.pantheonsite.io`)
+We do not recommend redirecting [platform domains](/guides/domains/platform-domains), especially on Live production environments, as it restricts our ability to provide support for scenarios where 3rd party services are utilized prior to the domain resolving to Pantheon (e.g., you're stacking your own custom CDN service on top of Pantheon's infrastructure).
```
stevector, line 26:
```suggestion
We do not recommend redirecting away from [platform domains](/guides/domains/platform-domains), especially on Live production environments, as it restricts our ability to provide support for scenarios where 3rd party services are utilized prior to the domain resolving to Pantheon (e.g., you're stacking your own custom CDN service on top of Pantheon's infrastructure). 
```

PR #9092, opened 2024-07-12
Clarify WPMU unsupported use case and copy edits

### LINKED ISSUE #9079: WPMU confusion: multiple TLDs are okay, WP Multi Network is not
Re: [WordPress Multisite: Unsupported Use Cases](https://docs.pantheon.io/guides/multisite/#unsupported-use-cases)

Priority: Medium

## Issue Description:
* Last item in the list of [unsupported use cases for WP Multisite on Pantheon](https://docs.pantheon.io/guides/multisite/#unsupported-use-cases) is confusing and gives the impression that we do not support connecting multiple top-level domains (TLDs) to a single Multisite application.
  > [WordPress Multi-Network](https://wordpress.org/plugins/wp-multi-network/) installations where multiple domains can be added aside from subdomains and subdirectories.

* The following section, titled ["Request a WordPress Multisite"](https://docs.pantheon.io/guides/multisite/#request-a-wordpress-multisite) talks about the supported alternative for [WP Multi Network](https://wordpress.org/plugins/wp-multi-network/), which is to request a WP Multisite custom upstream for your workspace, allowing you to create new networks on-demand by selecting the WP Multisite framework during site creation. However, because it's in a separate section, it's not clearly positioned as the supported alternative for the unsupported [WP Multi Network](https://wordpress.org/plugins/wp-multi-network/) use case.

## Suggested Resolution
* Revise copy of the last [unsupported use case item](https://docs.pantheon.io/guides/multisite/#unsupported-use-cases), do not mention multiple TLDs in this copy since that scenario is not exclusive to this use case and is actually supported. Redefine this use case as "a network of one or more multisites on a single instance, commonly achieved via plugins like [WP Multi Network](https://wordpress.org/plugins/wp-multi-network/)." Provide context for why this usage is not supported.
  * Add a sub-bullet that positions WP Multisite custom upstream as the supported alternative, and provide context for why this alternative is better.
* Revise the section ["Request a WordPress Multisite"](https://docs.pantheon.io/guides/multisite/#request-a-wordpress-multisite) with more distinction between a single network requirement (I need one WP Multisite) vs a requirement to create new and separate networks self-serve (I need a WP Multisite custom upstream). Right now the copy assumes anyone asking for WP Multisite needs it delivered as a custom upstream

rachelwhitton:
Closes #9079 

* Multiple TLDs are supported in a single instance, but multiple networks are not. 
* Copy edits

### DIFF
--- a/source/content/guides/multisite/01-introduction.md
+++ b/source/content/guides/multisite/01-introduction.md
@@ -35,12 +35,18 @@ We do not support uses of WordPress Multisite that run functionally-different or
 
 - Software as a service (SAAS) products
 - Agencies using one WordPress installation to support several customers
-- [WordPress Multi-Network](https://wordpress.org/plugins/wp-multi-network/) installations where multiple domains can be added aside from subdomains and subdirectories.
+- A network of one or more multisites on a single instance, commonly achieved via plugins like [WordPress Multi-Network](https://wordpress.org/plugins/wp-multi-network/).
+  - The supported alternative for this use case is to maintain a Custom Upstream that is based on Pantheon's WordPress Multisite framework, see the following section.
 
 ## Request a WordPress Multisite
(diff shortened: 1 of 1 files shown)


jazzsequence (approved):
This looks good to me from a technical perspective. I'm a little more fuzzy on the internal processes/policies implied here, but I _think_ those are all accurate as well. Would defer to @scottbuscemi or others, but it has the ring of truth.

ccharlton:
Guidance steps read fine to me. Much more informative than what was there before.

PR #9102, opened 2024-07-15
Fixing glossary TOC

### LINKED ISSUE #8992: Broken Table of Contents on Glossary
https://docs.pantheon.io/glossary/

The specific terms for the letter A are showing up in the TOC
Screenshot 2024-05-13 at 4 21 42 PM

stevector:
Fixes #8992 

The existing use of `h3`s is what causes the individual terms to show up in the Table of Contents. I'm getting rid of those and instead using the more semantically appropriate combination of `dl`, `dt`, and `dd` for a list of definitions.

### DIFF
--- a/source/content/guides/php/02-php-versions.md
+++ b/source/content/guides/php/02-php-versions.md
@@ -16,7 +16,7 @@ permalink: docs/guides/php/php-versions
 reviewed: "2022-12-13"
 ---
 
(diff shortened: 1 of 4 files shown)


stevector:
the internal consensus is to go with `dt` and `dd` and add classes to make the visuals work

PR #9110, opened 2024-07-16
wrapping cookie in check for development mode

### LINKED ISSUE #9098: Suppress cookie banner during local development
https://github.com/pantheon-systems/documentation/blob/29447a8ba02d5ce597e90e12fcdd0d6b231c4031/gatsby-ssr.js#L12

stevector:
closes #9098, but I'm really not sure if I should bother mucking with this type of code in Gatsby vs spending time exploring an eventual replacement.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -8,8 +8,11 @@ require("dotenv").config({
   path: `.env.${process.env.NODE_ENV}`,
 })
 
+
+
+
 // Gatsby Configuration, Options, and Plugins
-module.exports = {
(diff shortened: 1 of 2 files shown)


rachelwhitton:
@stevector all Playwright tests are impacted and timeout at the same step:

do-click-to-accept-cookies

If the above step is removed from each test, they continue to fail due to screenshot comparison failure:
dont-click-to-accept-cookies

This is true for both `end-to-end:local` and `end-to-end:compare-pr-to-live`

PR #9115, opened 2024-07-19
disambiguating environment logs

### LINKED ISSUE #8626: Log Files on Pantheon Doc Update
Re: [Log Files on Pantheon](https://docs.pantheon.io/guides/logs-pantheon)

Priority: Medium

## Issue Description:
There is no indication of which log contains build_step issues when composer fails to run. Discussions with support end up in a back and forth "change this, lets see"... "okay here is the next error in the log". It would save Pantheon support time if we could access this log directly. Perhaps we can, but the docs have no mention of composer or build_step.

## Suggested Resolution
Document which log contains the errors.

stevector:
Closes #8626 

I'd like input from Pantheors in PDE but we should merge anyway next week if there is no objection.

### DIFF
--- a/source/content/guides/logs-pantheon/01-introduction.md
+++ b/source/content/guides/logs-pantheon/01-introduction.md
@@ -1,5 +1,5 @@
 ---
-title: Log Files on Pantheon
+title: Environment Log Files on Pantheon
 subtitle: Introduction
 description: Use logs on Pantheon to help find, debug, and isolate potential problems.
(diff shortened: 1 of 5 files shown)


stevector:
I should get review from PDE after the build completes

rachelwhitton (approved):

stevector:
@rachelwhitton I'm assigning you to merge next week or earlier if we get a ➕ from @IngridKwok

PR #9121, opened 2024-07-24
Simplify PR template

rachelwhitton:
This PR simplifies our template for when new PRs are open. Most of the time, no one interacts with the template because it's too complex.


rachelwhitton:
Reviewed with @stevector on zoom together

PR #9133, opened 2024-07-25
Delete source/content/sftp.md

stevector:
https://docs.pantheon.io/sftp redirects to https://docs.pantheon.io/guides/sftp so this page isn't really used.

But it is still present in the codebase and still visible in PR preview environments where AGCDN directs aren't present. For instance: https://pr-9132-documentation.appa.pantheon.site/sftp

For clarity, we should just delete it


source/content/sftp.md
```diff
@@ -1,41 +0,0 @@
...
-contenttype: [doc]
-innav: [true]
-categories: [sftp]
-cms: [--]
-audience: [development]
-product: [--]
-integration: [--]
----
-
-<Youtube src="ktesh9SiHfc" title="On Server Development" />
```
rachelwhitton, line 19: @stevector this video does not exist anywhere else in the docs. Do we want to preserve it somewhere like we did for other videos from the 2016 "Self-Serve Training" initiative

stevector: I'm ok with losing it from the Docs site. I'm currently working with pre-sales on some updated demo videos.

PR #9140, opened 2024-07-31
Update 04-incident-remediation.md

joshualieb:
Update - synthetic responses require an AGCDN subscription, customers cannot request these be implemented without one.

## Summary

**[Doc Page Title](https://docs.pantheon.io/doc-title)** - Synthetic responses require an AGCDN subscription, customers cannot request these be implemented without one.

### DIFF
--- a/source/content/guides/disaster-recovery/04-incident-remediation.md
+++ b/source/content/guides/disaster-recovery/04-incident-remediation.md
@@ -27,7 +27,7 @@ In the event of an attack, exploit, or other issue related to the global edge, f
 
 ## Edge Failover
 
-Recovery of a site may take minutes, but time to recovery can be uncertain. Additional features can be implemented at the edge layer to provide a more immediate solution. These features minimize downtime while the specific issues are remediated.
+Recovery of a site may take minutes, but time to recovery can be uncertain. Additional features can be implemented at the CDN layer to provide a more immediate solution. These features minimize downtime while the specific issues are remediated.
 
 ### Experience Protection
 
--- a/source/content/guides/disaster-recovery/04-incident-remediation.md
+++ b/source/content/guides/disaster-recovery/04-incident-remediation.md
@@ -27,7 +27,7 @@ In the event of an attack, exploit, or other issue related to the global edge, f
 
 ## Edge Failover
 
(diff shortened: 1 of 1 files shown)


source/content/guides/disaster-recovery/04-incident-remediation.md
```diff
@@ -27,15 +27,15 @@ In the event of an attack, exploit, or other issue related to the global edge, f
 
 ## Edge Failover
 
-Recovery of a site may take minutes, but time to recovery can be uncertain. Additional features have been implemented at the Global CDN level to provide a more immediate solution. These features minimize downtime while the specific issues are remediated.
+Recovery of a site may take minutes, but time to recovery can be uncertain. Additional features can be implemented at the global edge to provide a more immediate solution. These features minimize downtime while the specific issues are remediated.
```
rachelwhitton, line 30: @joshualieb the previous copy was clearer to me because it specified a product by name (Global CDN), if that's not accurate and should be AGCDN instead - can we say AGCDN rather than "global edge" which is vague?

rachelwhitton (dismissed):

source/content/guides/disaster-recovery/04-incident-remediation.md
joshualieb: The tricky part is that experience protection is generally available and now part of the GCDN, but synthetic responses are only available with AGCDN, so it is not accurate to say that both of these are associated with GCDN, which is why I genericized it. Maybe we change it to something like:

> Recovery of a site may take minutes, but time to recovery can be uncertain. You can minimize downtime for your users while specific issues are remediated by taking advantage of several features that provide a response even when the backend CMS is down.

rachelwhitton: @joshualieb how's this? https://github.com/pantheon-systems/documentation/pull/9140/commits/4decd0ef753032f2bf31e1c6212eec5422fa9b8b

Because "global edge" feels too similar to "global CDN", I prefer to say "edge layer" when generalizing

rachelwhitton: Reviewed further with @jazzsequence - landed on CDN layer vs the other variations

PR #9151, opened 2024-08-05
Update docs: ssh-keys.md

sekedus:
## Summary

Doc: **[Generate and Add SSH Keys](https://docs.pantheon.io/ssh-keys)**

ㅤ
Add `-m PEM` to `ssh-keygen -t rsa`

Run `ssh-keygen -t rsa -m PEM` to generate a new RSA key pair, and the private key is saved in PEM format.

This format is sometimes required for compatibility with certain software or systems, and fix error `invalid privatekey: [B@12a3456b`

More: ["Invalid privatekey" when using JSch](https://stackoverflow.com/a/53783283/7598333)


rachelwhitton:
Tested this with @jazzsequence and it worked as expected. Thank you!

rachelwhitton (approved):

PR #9155, opened 2024-08-07
Update cloudflare.md

conorbauer23:
Update to include a section about needing access to the /.well-known/ path.

## Summary

**[Cloudflare Config](https://docs.pantheon.io/cloudflare)** - Updating to include a section about needing to allow access to the well-known path for Lets Encrypt cert renewal. This is related to a few bugs cards Edge Routing has seen as of late where customers were using CF as a CDN in front of their sites, and their cert expired due to geographic blocking:

https://getpantheon.atlassian.net/browse/BUGS-8271

### DIFF
--- a/source/content/dns-providers/cloudflare.md
+++ b/source/content/dns-providers/cloudflare.md
@@ -110,6 +110,20 @@ You can configure Cloudflare's CDN as an additional layer on Pantheon's Global C
 
    ![Example DNS only](../../images/cloudflare-full.png)
 
+### Restrict Content Based on Geographic Location
+
+If you're using Cloudflare's IP Geolocation feature, you will need to read the `CF-IPCountry` header and set `Vary: CF-IPCountry` on all responses.
+
+### Access to the /.well-known Path Is Needed
+
+If you're using Cloudflare's services to control or block traffic, an exception must be made for the `/.well-known` path in Cloudflare's config. If the `/.well-known` path is not accessible, Lets Encrypt may not be able to issue a certificate.
+
+### Cache Invalidation Best Practices
+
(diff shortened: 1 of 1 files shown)


source/content/dns-providers/cloudflare.md
```diff
@@ -144,6 +144,10 @@ A **CAA Record** specifies which certificate authority (**CA**) can issue HTTPS
 
 If you're using Cloudflare's IP Geolocation feature, you will need to read the `CF-IPCountry` header and set `Vary: CF-IPCountry` on all responses.
 
+## Access to the /.well-known/ Path Is Needed
+
+If you're using Cloudflare's blocking services to block traffic from specific regions, an exception must be made for the /.well-known/ path in CloudFlare's config. If the /.well-known path is not accessible, Lets Encrypt may not be able to issue a certificate.
```
jazzsequence, line 149: Inconsistency between `/.well-known` and `/.well-known/` -- pick whichever one you think is most accurate/correct and go with it. I would tend to lean toward `/.well-known` unless the context is specifically about the _contents_ of the directory (it reads to me like this is more about the directory _itself_. 

It would also be helpful to make the paths visually distinct by wrapping them in backticks for code formatting.
```suggestion
If you're using Cloudflare's blocking services to block traffic from specific regions, an exception must be made for the `/.well-known` path in CloudFlare's config. If the `/.well-known` path is not accessible, Lets Encrypt may not be able to issue a certificate. 
```

Let me know (or just merge the suggestion) if this works for you @conorbauer23

rachelwhitton:
Todo: I'll relocate the last 3 sections of this doc to the relevant section (option 2) as these only apply to the stacked CDN configuration option. Waiting for final updates from @conorbauer23 before I proceed

rachelwhitton (approved):

PR #9160, opened 2024-08-08
update rest api no-cache example code

### LINKED ISSUE #8675: Bypassing Cache with HTTP Headers Doc Update
Re: [Bypassing Cache with HTTP Headers](https://docs.pantheon.iocache-control)

Priority: Medium

## Issue Description:

I think that there may be a bug in the WordPress code example in the Bypassing Cache with HTTP Headers documentation. I recently needed to ensure that an endpoint under /wp-json was not cached. I found the article mention ed above and followed the instructions however, the endpoint in question was still cached. After reading through the code and tracing the hooks through WordPress core, I was able to discern a potential issue and create a solution.

The issue was identified in this block which is hooked into core WordPress filter `rest_post_dispatch`:

```php
// Re-define the send_header value with any custom Cache-Control header
function filter_rest_post_dispatch_send_cache_control( $response, $server ) {
    $server->send_header( 'Cache-Control', 'no-cache, must-revalidate, max-age=0' );
    return $response;
}
```

Here, the header is immediately sent by calling a method of parameter 2 `$server`. Producing an effect goes against the semantics of a filter which is intended to conditionally modify and return the value passed as parameter one.

Also, It was not producing the results that I was expecting. I believe that this is because it sends a `Cache-Control` header too early and, somewhere down the line, a subsequent call to `header()`  overwrites it.

I was able to refactor the example to call the `header()` method of parameter one `$response` instead. This method stores the header in memory instead of sending it right away.

```php
function filter_rest_post_dispatch_send_cache_control( $response, $server ) {
  $response->header('Cache-Control', 'no-cache, must-revalidate, max-age=0');
  return $response;
}
```

## Suggested Resolution

Update the documentation incorporating a working WordPress solution for bypassing cache in REST API requests,

jazzsequence:
Fixes #8675

## Summary

**[Bypassing Cache with HTTP Headers](https://docs.pantheon.io/cache-control)** - Updates the code snippet example for bypassing cache headers for rest requests

### DIFF
--- a/source/content/cache-control.md
+++ b/source/content/cache-control.md
@@ -118,14 +118,14 @@ $regex_json_path_patterns = array(
 foreach ($regex_json_path_patterns as $regex_json_path_pattern) {
   if (preg_match($regex_json_path_pattern, $_SERVER['REQUEST_URI'])) {
       // re-use the rest_post_dispatch filter in the Pantheon page cache plugin
-      add_filter( 'rest_post_dispatch', 'filter_rest_post_dispatch_send_cache_control', 12, 2 );
+      add_filter( 'rest_post_dispatch', 'filter_rest_post_dispatch_send_cache_control', 12 );
       break;
   }
 }
 
 // Re-define the send_header value with any custom Cache-Control header
-function filter_rest_post_dispatch_send_cache_control( $response, $server ) {
-    $server->send_header( 'Cache-Control', 'no-cache, must-revalidate, max-age=0' );
+function filter_rest_post_dispatch_send_cache_control( $response ) {
+    $response->header( 'Cache-Control', 'no-cache, must-revalidate, max-age=0' );
     return $response;
 }
 ```


stevector (commented):
@jazzsequence did you try this code on a site on the platform? If so, I'll approve/merge.

jazzsequence:
@stevector I haven't yet. I want to do that before we merge.

jazzsequence:
@stevector 
So, I think the _approach is correct_ -- in that we should not semantically be performing an action inside something intended to _filter_ content -- however, the solution _does not work_. I think this is actually related to [this issue](https://github.com/pantheon-systems/WordPress/issues/398) ([related community slack conversation](https://pantheon-community.slack.com/archives/C2GJ3JG7Q/p1722060406985489)). The underlying issue is, _it doesn't matter_ if the application sends `no-cache` headers to the CDN, the CDN _does not change its behavior_ when it receives those headers and continues to serve cached copies. 

Additionally, implementing the suggested change causes 502 errors if a REST API endpoint does not have a cached copy yet. For example, if I go to `/wp-json/wp/v2/pages/` on a site where I haven't already hit that endpoint, I get a 502. However, if I remove the code and visit that endpoint and then re-enable it, I will always get the cached version of that endpoint (until the cache expires). I think the _actual_ solution here is to remove this recommendation until the underlying issue is resolved (and open a docs issue that we come back to this if/when that issue is resolved). It's currently sitting inside a feature request (FEAT-1420).

jazzsequence:
lol it would help if I used the right header name (`Cache-Control` not `Cache Control`) in my test code 🙃 
Nevermind everything, this is good 😂

PR #9162, opened 2024-08-08
Add vitest to GitHub actions

### LINKED ISSUE #9083: Automate running of vitest
I'm adding vitest in https://github.com/pantheon-systems/documentation/pull/9047

It should run on PRs

jazzsequence:
Fixes #9083
Replaces #9161 

## Summary
Adds a GitHub actions workflow to run `npx vitest src/components` on PRs that change files in `src/` (can be altered to limit that scope to `src/components/` or if the workflow itself is updated.

Also allows for manual triggering (`workflow_dispatch`) in the Actions tab.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,5 +19,5 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 'latest'
-      - run: npm install
+      - run: npm ci
       - run: npx vitest src/components
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,23 @@
+name: Test
+on:
+  pull_request:
+    paths:
+      - 'src/**'
+      - '.github/workflows/test.yml'
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 'latest'
+      - run: npm install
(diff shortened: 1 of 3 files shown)


jazzsequence:
Failing tests are expected per https://github.com/pantheon-systems/documentation/issues/9083#issuecomment-2276310831. Successful run can be seen in https://github.com/pantheon-systems/documentation/actions/runs/10307574827/job/28533128469?pr=9161

.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 'latest'
+      - run: npm install
```
stevector, line 22: I think we want `npm ci` https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci

.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+      - run: npx vitest src/components
```
stevector, line 23: Should there be a line break after this line?

jazzsequence: Doesn't need to be. More a matter of stylistic preference and readability.

.github/workflows/test.yml
jazzsequence:
```suggestion
      - run: npm ci
```

stevector (approved):
Thanks @jazzsequence. I made the filename more specific.

jazzsequence:
I left it open ended in case we wanted to use it for other types of tests, but since we're running on specific paths, that's probably fine.

PR #9201, opened 2024-09-03
561 Error hard to diagnose

jms-pantheon:
## Summary

**5xx Level Errors](https://docs.pantheon.io/guides/errors-and-server-responses/5xx-errors#error-561-no-site-detected)** - Diagnosing 561 errors was not clear, make it more apparent that `index.php` is the key piece.


jms-pantheon:
Whoops thanks @rachelwhitton good catch on the duplicate content

PR #9219, opened 2024-09-11
Set published date with hash based timestamp

### LINKED ISSUE #9126: Bug: RSS subscribers are not updated when multiple release notes are published in a single day
Publishing more than one release note in a given day does not update RSS subscribers, we suspect that is because the entries have the same value for pubDate (e.g., `<pubDate>Wed, 24 Jul 2024 00:00:00 GMT</pubDate>`)

rachelwhitton:
Closes #9126 

Fixes known RSS bug by setting an arbitrary hour, minute, and second for each release note within the RSS feed by deriving a hash of the note. 

Adds new release note: [Bug fix for Pantheon release notes RSS feed](https://pr-9219-documentation.appa.pantheon.site/release-notes/2024/09/rss-bug-fix)

### DIFF
--- a/source/releasenotes/2024-09-12-rss-bug-fix.md
+++ b/source/releasenotes/2024-09-12-rss-bug-fix.md
@@ -3,4 +3,6 @@ title: Bug fix for Pantheon release notes RSS feed
 published_date: "2024-09-12"
 categories: [documentation]
 ---
-Previously, our release note publishing process used Markdown and Git with dates set only by year, month, and day. This caused RSS readers, including our Community Slack integration, to ignore multiple release notes with the same date. To address this, we now include an arbitrary hour, minute, and second for each release note in the RSS feed, derived from a hash of the note.
+RSS readers will now detect multiple release notes from the same date.
+
+Our release note publishing process runs through Markdown and Git with published dates set manually only at the granularity of year, month, and day. RSS readers (like our Community Slack integration) ignore multiple items with identical dates and times. We have worked around this limitation by setting an arbitrary hour, minute, and second for each release note within the RSS feed by deriving a hash of the note.
--- a/source/releasenotes/2024-09-12-rss-bug-fix.md
+++ b/source/releasenotes/2024-09-12-rss-bug-fix.md
@@ -1,0 +1,6 @@
+---
+title: Bug fix for Pantheon release notes RSS feed
+published_date: "2024-09-12"
+categories: [documentation]
+---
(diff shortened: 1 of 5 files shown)


jazzsequence (approved):

rachelwhitton:
Added multiple test entries in a single day with the same published date and they were all 3 picked up by my RSS reader
Screenshot 2024-09-11 at 5 31 01 PM

@mel-miller can you review?

mel-miller (approved):
@rachelwhitton -- TBH, I'm not that familiar with this type of thing. But the solution seems good to me and I don't see anything dangerous introduced.

jazzsequence (approved):
lgtm

source/releasenotes/2024-09-12-rss-bug-fix.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Bug fix for Pantheon release notes RSS feed
+published_date: "2024-09-12"
+categories: [documentation]
+---
+A few months ago we [added an RSS feed](release-notes/2024/07/rss-feed) to Pantheon release notes, and today we're improving the reliability of that experience. Previously, RSS readers would not pick up new release note entries if more than one was published on the same day. This bug is now fixed, ensuring a more reliable notification experience.
```
stevector, line 6:
```suggestion
RSS readers will now detect multiple release notes from the same date.

Our release note publishing process runs through Markdown and Git with published dates set manually only at the granularity of year, month, and day. RSS readers (like our Community Slack integration) ignore multiple items with identical dates and times. We have worked around this limitation by setting an arbitrary hour, minute, and second for each release note within the RSS feed by deriving a hash of the note.
```

jazzsequence (commented):
update is a bit more granular than strictly necessary, but I like the breakdown and I don't think it's a problem. I imagine getting into the weeds discussing the arbitrariness of the timestamp might lead to questions about notes being ordered incorrectly (or call attention to the discrepancy between where release notes appear in readers vs on the website).

source/releasenotes/2024-09-12-rss-bug-fix.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Bug fix for Pantheon release notes RSS feed
+published_date: "2024-09-12"
+categories: [documentation]
+---
+Previously, our release note publishing process used Markdown and Git with dates set only by year, month, and day. This caused RSS readers, including our Community Slack integration, to ignore multiple release notes with the same date. To address this, we now include an arbitrary hour, minute, and second for each release note in the RSS feed, derived from a hash of the note.
```
stevector, line 6: This is not accurate. We still use Markdown and Git.

PR #9228, opened 2024-09-18
Add OpenSSH requirements to 02-install.md

codyfishman:
Adding details about OpenSSH requirements for Terminus nested Drush and WP-CLI commands.

Verified this information in https://pantheon.slack.com/archives/C1P99MH3M/p1726675141658079. 

## Summary

### DIFF
--- a/source/content/terminus/02-install.md
+++ b/source/content/terminus/02-install.md
@@ -56,21 +56,18 @@ Terminus does not work with the following platforms:
 ### Required Packages
 
 - PHP Version 7.4 or later
-   - You can check your PHP version by running `php -v` from a terminal application. You must have the [php-xml extension](https://secure.php.net/manual/en/dom.setup.php) for:
-    - mbstring
-    - XML
-    - [cURL](https://secure.php.net/manual/en/curl.setup.php)
-    - [CLI](http://www.php-cli.com)
+  - You can check your PHP version by running `php -v` from a terminal application.
+  - You must have the [php-xml extension](https://secure.php.net/manual/en/dom.setup.php) for:
+     - mbstring
+     - XML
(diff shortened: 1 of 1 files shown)


source/content/terminus/02-install.md
```diff
@@ -66,6 +66,9 @@ Terminus does not work with the following platforms:
 
 ### Recommended Packages
 
+- [OpenSSH 7.8](https://github.com/PowerShell/Win32-OpenSSH/wiki/) or later
+   - You can check your OpenSSH version by running `ssh -V` from a terminal application. 
+   - This is required package for executing nested Drush or WP-CLI commands.
```
rachelwhitton, line 71: Since the copy says this package is required, it should be moved from "Recommended Packages" to the above section, "Required Packages".

Should this also be added to the [projects README](https://github.com/pantheon-systems/terminus?tab=readme-ov-file#required-packages)? 

@kporras07 @stovak can you review and weigh in?

kporras07: I like the idea of calling openssh out here (and maybe even moving it to required); I'm not sure I like linking it to PowerShell Win32-OpenSSH but I don't know of a better link. Would it make sense to NOT link anything here and allow customers to get the right package based on their OS?

kporras07: 100% agreement on also getting it to project's README. I can create a quick PR once we figure out how we want to do it here

rachelwhitton: @kporras07 I reviewed this again with @jazzsequence during docs coworking this morning, we decided to remove the link entirely (we don't link out to PHP in that same list so I think it's fine without), move OpenSSH up to the required packages section, and remove the recommended packages section. Can you do another review when convenient?

kporras07 (approved):
LGTM

PR #9231, opened 2024-09-20
Add the "Internal Exception" Troubleshooting

ersonjn:
## Summary

**[Pantheon YML Troubleshooting](https://docs.pantheon.io/pantheon-yml#troubleshooting)** - This is to give the Pantheon customers and clients temporary official documentation for the workaround while this is still being addressed for better error handling/messaging.

### DIFF
--- a/source/content/pantheon-yml.md
+++ b/source/content/pantheon-yml.md
@@ -328,6 +328,21 @@ remote: Valid versions are: 1
 
 While our parser will reject a `pantheon.yml` that is invalid, it won't necessarily give you the exact reason the file is invalid. Syntax errors are the most common reason for an invalid `pantheon.yml` file.
 
+### There was an internal error processing the request
+
+The Platform will throw an "internal error" when you try to push a commit that includes a `pantheon.yml` changes to a **_branch without an associated multidev_**. The error message will resemble:
+
+```none
+remote: PANTHEON ERROR:
+remote:
+remote: The specific error returned was: Internal exception
+remote:
+remote: There was an internal error processing the request
+remote:
+```
+
+To resolve this, create multidev first before pushing the `pantheon.yml` changes.
+
 ### Why can’t I update the PHP version on my Multidev?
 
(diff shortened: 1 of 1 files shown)


source/content/pantheon-yml.md
```diff
@@ -328,6 +328,21 @@ remote: Valid versions are: 1
...
+remote: PANTHEON ERROR:
+remote:
+remote: The specific error returned was: Internal exception
+remote:
+remote: There was an internal error processing the request
+remote:
+```
+
+To resolve this, create multidev first before pushing the `pantheon.yml` changes.
+
```
rachelwhitton, line 345: I could not reproduce this issue on a d10 Integrated Composer performance small site and a vanilla WP sandbox site. In both cases, the push was successful without existing multidevs. 

On both test sites, I checked out a new branch locally and made changes to `pantheon.yml` (upgraded PHP versions), committed, and pushed with the following result: 
```
remote:
remote: PANTHEON NOTICE:
remote:
remote: Skipping code sync, no Multidev environments were found for branch "php8-3".
remote:
```

@ersonjn can you provide the full error message output for the support issue related to this docs PR?  I expect something else caused this internal error and the surrounding output should help identify. Or if you have the ticket number? 

@jazzsequence can you recommend an SME for this?

jazzsequence: @rachelwhitton Not sure about SME but @greg-1-anderson might have some context...

greg-1-anderson: I'm not sure about the cause of the reported error. It happens intermittently, and rarely. I think it only happens if the name selected refers to a multidev that used to exist, but no longer does. I don't think that you can reproduce this simply by creating and deleting a multidev; there is something more complex at work.

Since this problem happens only rarely, and the symptoms might also be indicative of a different platform problem, I don't think that we should include this workaround in our public docs.

rachelwhitton: Thanks all! I'll close this PR for now

PR #9257, opened 2024-10-14
WP S3 plugin needs full package with vendor folder

jms-pantheon:
## Summary

**[AWS S3 Setup for WordPress](https://docs.pantheon.io/guides/wordpress-developer/wordpress-s3#install-and-deploy-s3-uploads-plugin)** - Download instructions were unclear.

### DIFF
--- a/source/content/guides/wordpress-developer/07-wordpress-s3.md
+++ b/source/content/guides/wordpress-developer/07-wordpress-s3.md
@@ -10,8 +10,8 @@ audience: [development]
 product: [--]
 integration: [aws]
 tags: [files]
-contributors: [sarahg,carl-alberto]
-reviewed: "2023-05-19"
+contributors: [sarahg,carl-alberto,jms-pantheon,jazzsequence]
+reviewed: "2024-10-23"
 showtoc: true
 permalink: docs/guides/wordpress-developer/wordpress-s3
 ---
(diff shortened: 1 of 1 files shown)


rachelwhitton:
assigning to @jazzsequence to consider testing with the latest version of the plugin and updating docs

jms-pantheon:
Reviewing this now, it looks like that plugin is removing the manual install. Perhaps we should just refer people to that project's readme for install instructions

jazzsequence:
Looking at the HM repo now, and in particular the line:

> [Composer's autoloader](https://getcomposer.org/doc/01-basic-usage.md#autoloading) must be loaded before S3 Uploads is loaded.

I'm thinking that it's not possible to install S3 Uploads _without_ composer. We should call that out specifically in the [opening paragraph](https://docs.pantheon.io/guides/wordpress-developer/wordpress-s3#integrate-s3-with-wordpress) as a requirement. I also don't think that WP Offload Media _requires_ a license (I am using it on my personal site without one), but it's required if you want the urls to match your domain (rather than the S3 bucket or DO space or whatever). So maybe we can lean heavier into using that plugin instead.

Given we just [link out to WP Offload Media for docs/instruction there](https://docs.pantheon.io/guides/wordpress-developer/wordpress-s3#multisite-compatibility), I think we should probably do the same for S3 Uploads (maybe with a note that WP-CLI commands in HM's docs must be wrapped in Terminus syntax for use on a Pantheon site.

jms-pantheon:
Thanks @jazzsequence that looks like a good move

jazzsequence:
Merging!

PR #9272, opened 2024-10-18
[UpdateTool - Terminus Information] Update commands and releases to version 3.6.0.

pantheon-upstream:
[UpdateTool - Terminus Information] Update commands and releases to the latest Terminus version.


rachelwhitton:
Not sure why this popped up when we already had an update for this version about a month ago via #9229

PR #9293, opened 2024-11-04
 npm cleanup

stevector:
cleaning up some npm dependencies


gatsby-config.js
```diff
@@ -235,14 +235,6 @@ module.exports = {
     },
     `gatsby-transformer-sharp`,
     `gatsby-plugin-sharp`,
-    {
-      resolve: "gatsby-plugin-web-font-loader",
```
stevector, line 239: It appears we weren't using this font at all.

stevector:
@rachelwhitton and @jazzsequence this one is not urgent, but I cleaned up some of our NPM dependencies, mostly by removing packages. I tracked the operations I did in [cleanup.txt](https://github.com/pantheon-systems/documentation/pull/9293/files#diff-043616eca9b10a557d218738b0811410e1324d254ea1b4f7be889ffda255ce65)

cleanup.txt
```diff
@@ -0,0 +1,14 @@
+npm uninstall algoliasearch
```
stevector, line 1: I should delete this file before merging

rachelwhitton (approved):
I ran local VRT with passing results, looks good to me 👍 

screencapture-localhost-9323-2024-11-05-10_22_07 (1)

jazzsequence (approved):
this isn't updating anything, it's just removing unused things?

stevector:
@jazzsequence it also contained some updates triggered by `npm update msgpackr` which I ran to fix a single notice in `npm audit`. Then `npm audit fix` did some other additional updates. Resolving most of the warning seems like it would require a major version update to Gatsby which I don't think is worth doing.

PR #9299, opened 2024-11-13
Clarify what counts for known bots, incognito bots and load testing traffic limits

rachelwhitton:
## Summary

**[Measuring Site Traffic](https://docs.pantheon.io/guides/account-mgmt/traffic#why-doesnt-pantheons-traffic-metrics-match-my-other-analytics)** - Updates the table so that the following are distinguished, rather than grouped together in the existing row for "Automated traffic from bots or load testing": 

* Known and Self-identified Bots = No, Sometimes
* Unknown and Incognito Bots = Yes, Sometimes
* Load Testing and Third-party Monitoring = Yes, Sometimes


ccharlton (commented):
LGTM. Thank you.

PR #9309, opened 2024-11-15
Update Local Object Cache Pro config for Lando

RayHollister:
## Summary

**[Enable Object Cache Pro for WordPress](https://docs.pantheon.io/object-cache/wordpress)** - Added note after 6c advising Lando users to disable Redis to prevent it from deactivating their server. Suggested by @pwtyler and tested by myself.

### DIFF
--- a/source/content/addons/object-cache/howto/wordpress.md
+++ b/source/content/addons/object-cache/howto/wordpress.md
@@ -97,11 +97,11 @@ terminus redis:enable <site>
 	<Alert title="Note" type="info">
 	**Note for Lando users:**
 
-	If you are using Lando for your local development, add the following to your `wp-config.php` file to deactivate Redis to prevent it from disabling your Lando server:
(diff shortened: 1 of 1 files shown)


jazzsequence:
This is what I've got set and seems to be working locally:

```php
$ocp_settings = [
    // All your OCP defaults, except use the following for the token:
    'token' => isset( getenv( 'OCP_LICENSE' ) ) ? getenv( 'OCP_LICENSE' ) : '',
     // ...other stuff...
];

if ( isset( $_ENV['LANDO'] && 'ON' === $_ENV['LANDO'] ) {
    $ocp_settings['serializer'] = 'php';
    $ocp_settings['compression'] = 'none';
}

define( 'WP_REDIS_CONFIG', $ocp_settings );
```

Note: I'm getting a License error locally even when the token is correct. I've a side conversation with Till from OCP about that, but I don't think the license impedes functionality, so I think just an empty string for that as a default would still work.

jazzsequence:
Closing in favor of https://github.com/pantheon-systems/documentation/pull/9327

PR #9321, opened 2024-11-21
fixing muted attribute in release note

stevector:
recreating #9320.

### DIFF
--- a/source/releasenotes/2024-11-20-content-publisher-beta.md
+++ b/source/releasenotes/2024-11-20-content-publisher-beta.md
@@ -9,7 +9,7 @@ Pantheon is now accepting customers into a [private beta for our Content Publish
 
 With Pantheon Content Publisher, your team can now safely publish content directly from Google Docs to Drupal, WordPress, or Next.js sites — quickly and seamlessly. Skip the extra steps and empower everyone who writes and collaborates to participate in web publishing.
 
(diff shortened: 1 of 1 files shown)


source/releasenotes/2024-11-20-content-publisher-beta.md
```diff
@@ -9,7 +9,7 @@ Pantheon is now accepting customers into a [private beta for our Content Publish
 
 With Pantheon Content Publisher, your team can now safely publish content directly from Google Docs to Drupal, WordPress, or Next.js sites — quickly and seamlessly. Skip the extra steps and empower everyone who writes and collaborates to participate in web publishing.
 
-<a href="https://pantheon.io/blog/introducing-pantheon-content-publisher"><video style="border-radius:0.9rem;height:auto;max-width:100%;" loop muted="muted" autoplay="true">
+<a href="https://pantheon.io/blog/introducing-pantheon-content-publisher"><video style="border-radius:0.9rem;height:auto;max-width:100%;"  loop="true" muted="true" autoplay="true" playsinline="true">
```
stevector, line 12: @christianyates @malikkotob even though the guidance I'm seeing on the web would suggest doing this as `loop muted autoplay playsinline` instead of including the `=true`, I'm including it because the WYSIWYG in the Drupal site keeps adding back `=""`. So for consistency I'm setting a value in both places.

PR #9327, opened 2024-11-22
Add Local Lando config for OCP and suggest Secrets for license storage

jazzsequence:
## Summary

**[Enable Object Cache Pro for WordPress](https://docs.pantheon.io/object-cache/wordpress)** - Updates Lando configuration to configure OCP locally for Lando (alternative to #9309 which recommends disabling OCP). Additionally, this PR suggests using Pantheon Secrets for storing the license key for Composer-based installs.

### DIFF
--- a/source/content/addons/object-cache/howto/wordpress.md
+++ b/source/content/addons/object-cache/howto/wordpress.md
@@ -123,16 +123,8 @@ Refer to the [official Object Cache Pro documentation](https://objectcache.pro/d
 	composer config --auth http-basic.objectcache.pro token $(terminus remote:wp <site>.<env> -- eval "echo getenv('OCP_LICENSE');")
 	```
 	
-	This will pull the Object Cache Pro license token directly into the `auth.json` file.
+	This will pull the Object Cache Pro license token directly into the `auth.json` file. You can open the `auth.json` file locally to ensure that it has a structure that looks like this:
 	
-	**Manually:**
-
-	1. Create an `auth.json` file in your directory.
-	
-	1. Run `terminus remote:wp <site>.<env> -- eval "echo getenv('OCP_LICENSE');")` to output the token to your terminal, then copy it into the `password` field in the next step.
-
-	1. Add the following code to the `auth.json` file.
-
 		```json
 		{
 			"http-basic": {
--- a/source/content/addons/object-cache/howto/wordpress.md
+++ b/source/content/addons/object-cache/howto/wordpress.md
@@ -142,7 +142,7 @@ Refer to the [official Object Cache Pro documentation](https://objectcache.pro/d
 	
 	</Alert>
 
-1. Open your `composer.json` file and locate the `repositories` section. If it doesn't exist, add it as shown below:
+1. Open your `composer.json` file and locate the `repositories` section. If it doesn't exist, add it in the "repositories" section as shown below:
 
 	```json
 		repositories: [
--- a/source/content/addons/object-cache/howto/wordpress.md
+++ b/source/content/addons/object-cache/howto/wordpress.md
@@ -302,7 +302,7 @@ Lando's [Pantheon recipe](https://docs.lando.dev/plugins/pantheon/) includes Red
 			type: redis:6.0
 	```
 
-	- This ensures that the Redis version in your Lando environment matches the 6.x environment on Pantheon.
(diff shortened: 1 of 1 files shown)


source/content/addons/object-cache/howto/wordpress.md
```diff
@@ -309,6 +390,18 @@ Refer to the [official Object Cache Pro documentation](https://objectcache.pro/d
...
+	```php
+	if ( isset( $_ENV['LANDO'] ) && 'ON' === $_ENV['LANDO'] ) {
+		add_filter( 'http_request_args', function ( $args ) {
+			$args['sslverify'] = false;
+			return $args;
+		} );	
+	}
+	```
+	
+	- You may wish to make this file local-only by adding it to your `.gitignore` file. This error will not cause any issues with the functioning of Object Cache Pro or the behavior of the plugin on Pantheon but it might prevent you from being able to make updates to the plugin locally.
```
pwtyler, line 404: - Why is this section down here and not under "Local Config" above?
- Which file do you mean by "make this file local-only"? Are you advocating putting ^ in a wp-config-local.php?

jazzsequence: * It's down here rather than under `Local Config` because I don't actually think it causes any problems (you can still use the plugin locally, but the connections to WP.org for the News widget and updates from OCP's release API don't work), I think it's a Lando-specific issue (I don't think DDEV has this problem), and, I can't prove this, but Till mentioned that he thought Bedrock had built a fix for this issue which means it wouldn't be a problem on Composer-based installs. I don't want to recommend adding configuration for local development as a default, it should be "if you feel like you need this".
* You can't put this code in a wp-config file. wp-config is too early in the runtime for `add_filter` to be a recognized function. It needs to be in a `mu-plugin` as described. The "local-only" suggestion is so that you aren't shipping code that has no relevance to your production site to your git repository, you _only_ need this in a local environment.

pwtyler: Ah, I missed the line about mu-plugin. I was a bit confused because I thought you had to do hacky workarounds to use add_filter in wp-config 😆

pwtyler (approved):

source/content/addons/object-cache/howto/wordpress.md
jazzsequence: I mean, if you really wanted to do hacky workarounds you could `require_once` whatever file has `add_filter` but I wouldn't do it.

source/content/addons/object-cache/howto/wordpress.md
```diff
@@ -152,7 +150,7 @@ Refer to the [official Object Cache Pro documentation](https://objectcache.pro/d
 	git add auth.json && git commit -m "Add Object Cache Pro auth token."
```
stevector, line 150: It seems odd that we're recommending committing the license to git. That shouldn't be necessary with Secrets Manager.

jazzsequence: The auth.json is really only for local development/updates. Once we're on Pantheon, the licence is provided by Pantheon environment variables. The convenience of committing it to the repository is that the token is shared between users of the repository (vs everyone having to create their own auth.json), but I've actually tested and confirmed that you don't need auth.json committed to the repository at all. I'm going to walk back this suggestion and add a note that everyone would need to individually pull down/create an auth.json themselves but that it should be gitignored.

jazzsequence:
Removed recommendation to use Secrets. Reworked some other bits and pieces and also removed the recommendation to commit the `auth.json` file to the git repository (and added notes about how you need it if you're going to update the plugin locally with composer).

@pwtyler @stevector would appreciate you both taking another look

source/content/addons/object-cache/howto/wordpress.md
```diff
@@ -295,7 +291,55 @@ Refer to the [official Object Cache Pro documentation](https://objectcache.pro/d
 
 	- If you are using WordPress Multisite, subsites do not get their own configuration or graphs. Navigate to `/wp-admin/network/settings.php?page=objectcache` to view network-wide configuration and graphs. This is the only screen throughout the network that displays this information.
 
-### Additional Considerations
+## Local configuration with Lando
```
stevector, line 294: @jazzsequence did you do all these steps yourself? If not, one of us should do them on our machines to confirm accuracy. If you did, I'll mark the PR as approved.

jazzsequence: yes, but I can check the redis config in the lando.yml one more time

stevector (approved):
thanks for the updates!

PR #9334, opened 2024-11-27
update the custom upstream intro to add glossary terms

### LINKED ISSUE #9333: Add explanation of "platform framework" to glossary
If such an entry existed already we'd link to it in this PR: https://github.com/pantheon-systems/documentation/pull/9332/files

Glossary entries have to be sourced from somewhere. I'm not sure if we should also add a page that is just about platform frameworks or include it on an existing page.

jazzsequence:
I am hypothesizing that glossary terms are added based on the <dfn> tag. I am further hypothesizing that if we leave the markdown syntax on this page intact, the new glossary terms would not be added correctly, therefore I have converted all the markdown bullets to html list items.

I have added "framework" on this page because it makes the most sense to be included on a page about upstreams.

fixes #9333

## Summary

**[Custom Upstreams](https://docs.pantheon.io/guides/custom-upstream)** - adds framework to the list of terminology and makes all terms `<dfn>` elements for the glossary

### DIFF
--- a/source/content/guides/custom-upstream/01-custom-upstream.md
+++ b/source/content/guides/custom-upstream/01-custom-upstream.md
@@ -126,20 +126,16 @@ Support for Custom Upstreams requires that the externally hosted upstream reposi
 
 Here are definitions for commonly used terms:
 
-- **Upstream**: A repository that acts as a parent for another repository, like [Pantheon's WordPress Upstream](https://github.com/pantheon-systems/wordpress). The next two definitions are specific types of Upstreams.
-
-- **Custom Upstream**: A repository restricted to members of an organization, containing a common codebase for new sites. This type of repository is a child repository to Pantheon's core upstreams ([WordPress](https://github.com/pantheon-systems/wordpress),[Drupal (Latest Version)](https://github.com/pantheon-upstreams/drupal-composer-managed), [Drupal 7](https://github.com/pantheon-systems/drops-7)) and acts as a parent for site level repositories.
-
-- **Public Upstream**: A repository that is open to all Pantheon users which contains a common codebase for new sites, like [Panopoly](https://github.com/populist/panopoly-drops-7).
-
-- **Repository**: A collection of files packaged in a single directory under version control.
-
-- **Remote Repository**: A central version control location, such as a repository residing on GitHub, Bitbucket, or GitLab.
-
-- **Upstream Updates**: Code changes that are made once in a parent (upstream) repository, then applied "downstream" to child repositories. This is how Pantheon's one-click updates work.
-
-- **Site Repository**: Child repository where upstream updates are applied and site specific customizations are tracked, similar to your site's codebase on Pantheon.
-
+<ul>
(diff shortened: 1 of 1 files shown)


source/content/guides/custom-upstream/01-custom-upstream.md
```diff
@@ -126,20 +126,38 @@ Support for Custom Upstreams requires that the externally hosted upstream reposi
...
+
+<p><dfn id="upstream-updates"> Upstream Updates</dfn> are code changes that are made once in a parent (upstream) repository, then applied "downstream" to child repositories. This is how Pantheon's one-click updates work.</p>
+
+### Site Repository
+
+<p>A <dfn id="site-repository">site repository</dfn> is a child repository where upstream updates are applied and site specific customizations are tracked, similar to your site's codebase on Pantheon.</p>
+
+### Framework
+
+<p>The <dfn id="upstream-framework">upstream framework</dfn> (or site framework) determines the server configuration for a given CMS. It includes things like Nginx configuration, the relevant CLI tool to install (Drush or WP-CLI), etc. This setting is <em>not visible</em> to users and once a framework has been set for a site (based on how the upstream was configured), it cannot be changed without <a href="https://docs.pantheon.io/guides/custom-upstream/switch-custom-upstream">changing the upstream</a> itself.</p>
```
rachelwhitton, line 160: > This setting is <em>not visible</em> to users and once a framework has been set for a site (based on how the upstream was configured), it cannot be changed without <a href="https://docs.pantheon.io/guides/custom-upstream/switch-custom-upstream">changing the upstream</a> itself.

^ I don't think is accurate, because: 

* The site framework **is** visible from the site dashboard and also visible from the CLI `terminus site:info` 
  * the upstream framework isn't visible from a non-admin dashboard page but customers can see it from the CLI `terminus upstream:info` 
* A site's framework **cannot** be changed after site creation. Switching a site's upstream will only work when the current upstream and the new upstream have matching frameworks - it will not switch the site framework ever, and if you try to switch the site to an upstream with a different framework the workflow will fail. 
  * A custom upstream's framework **can** be changed after creation but only from the admin debug panel, so customers would have to contact support to do that  

screencapture-dashboard-pantheon-io-sites-452cb6f4-4068-4275-bf10-fc3dd5670274-2024-12-04-15_13_20

Screenshot 2024-12-04 at 3 35 05 PM

rachelwhitton (changes requested):

source/content/guides/custom-upstream/01-custom-upstream.md
jazzsequence, line 160:
```suggestion
<p>The <dfn id="upstream-framework">upstream framework</dfn> (or site framework) determines the server configuration for a given CMS. It includes things like Nginx configuration, the relevant CLI tool to install (Drush or WP-CLI), etc. This setting is visible to users via Terminus and in the Site Settings. Once a framework has been set for a site (based on how the upstream was configured), it cannot be changed through the UI or Terminus. If you need to change your site's framework, you would need to contact Support.</p>
```

@rachelwhitton ☝️ Better?

rachelwhitton (approved):

PR #9342, opened 2024-12-05
Docs for cms buttons in site dash

### LINKED ISSUE #9142: The Site Dashboard Doc Update
Re: [The Site Dashboard](https://docs.pantheon.io/site-dashboard)

Priority: Low

## Issue Description:
This doc never actually mentions anything about the `Visit {env} Site` and `Site Admin` buttons. It describes everything else in the site dashboard _except_ what these buttons do.

## Suggested Resolution
Add docs about the two buttons, screenshots of the buttons (maybe in dev/test/live mode or maybe that's overkill) and describes what they're for.

rachelwhitton:
Fixes #9142 
## Summary

* [Site Dashboard](https://pr-9342-documentation.appa.pantheon.site/site-dashboard#left-navigation): 
  * Document the behavior of the CMS buttons in the site dash (Visit Site and Site Admin). 
  * Note platform domain is expected by default unless overridden by custom primary domain. 
  * Refresh stale image at top of the doc.


rachelwhitton:
@jazzsequence can you review when convenient?

source/content/dashboard.md
```diff
@@ -21,7 +21,7 @@ Easily visualize and manage sites, users, traffic, metrics, visual regression te
 
 Our site dashboard provide a real-time interface into Pantheon. It includes everything you need to build, launch, and run sites. Create new projects, manage team access, interface via SFTP or Git (version control), and track miscellaneous tools like backups. It also allows you to deploy and scale your sites.
 
-![The Pantheon Dashboard](../images/dashboard/new-dashboard/2024/site-dashboard.png)
+![The Pantheon Dashboard](../images/dashboard/new-dashboard/2024/_site-dashboard.png)
```
jazzsequence, line 24: Nit: would probably prefer replacing the old screenshot or renaming the new one (e.g. `site-dashboard-2024.png`) rather than adding an underscore. What happens with the next redesign? And the next? etc. It sets a precedent that I don't think we'd want to perpetuate in the future.

source/content/site-dashboard.md
```diff
@@ -57,11 +58,21 @@ This environment is for your live site, where content creators actively post and
...
 
-- **Commit Log:** Review a list of all code changes by viewing the commits. Click an individual commit to see the log message and list of changes.
+- **Site Admin**: This will open your sites admin login page for the given environment in a new tab. If you're not already logged into your sites admin interface, this button will open `/user/login` for Drupal and `/wp-login.php` for WordPress. Once you've logged in and have an active admin session, this button will take you to `/user` for Drupal and `/wp-admin` for WordPress.
+
+<Alert title="Note" type="info">
+
+By default, the buttons described above will open links that are served via their respective [platform domains](/guides/domains/platform-domains) (e.g., `test-example.pantheonsite.io`). If you want these buttons to open your site using your own custom domain instead (e.g., `test.example.com`) you can connect a [custom domain](/guides/domains/custom-domains) then configure it as the [primary domain](/guides/domains/primary-domain) for the target environment. This is typically only done for the Live environment as part of launch however you can do this anytime for any environment based on your preferences and requirements.
```
jazzsequence, line 69: Probably not worth overloading this note but WordPress multisites break if you set a platform domain, so this can't actually be done (or shouldn't) for WordPress multisites that host multiple domains (even for live environments).

jazzsequence (approved):
A couple notes but not worth blocking or rabbit holing on. Agree with @stevector re: stray `<br />` tag, but also not explicitly blocking.

source/content/site-dashboard.md
```diff
@@ -27,6 +27,7 @@ On the top left of the page, you will see the following information:
 - If applicable, a yellow **Upgrade** button, that will allow you to [upgrade the site](/guides/account-mgmt/plans/site-plans).
 - If applicable, the name of the Professional Workspace containing this site.
 
+<br />
```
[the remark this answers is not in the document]
rachelwhitton: I think there needs to be more space above a new paragraph following a list so I threw in the `<br />` but I can do a separate PR to address that in the template styles instead of this approach 
Screenshot 2024-12-18 at 9 28 40 AM

source/content/site-dashboard.md
rachelwhitton: We could warn about that in the primary domain doc that's crosslinked in this note? https://docs.pantheon.io/guides/domains/primary-domain

source/content/dashboard.md
rachelwhitton: I'll rename this to `../images/dashboard/new-dashboard/2024/site-dashboard-2024.png` as suggested

I don't feel strongly about the renaming convention so long as the name is changed then we'll get the fresh version. 

For background though, I added `2024` subdir to the path in an earlier PR which contained mostly the same images just with the nav cropped out based on PDEs recommendations, in a follow up PR I generated fresh screenshots from the product with the new nav shown for context and added `_` to the filenames because it was an easy modifier to do a search and replace for and I had already done the `2024` modifier previously in the folder structure

source/content/site-dashboard.md
rachelwhitton: @jazzsequence can you review the latest commit when convenient? https://github.com/pantheon-systems/documentation/pull/9342/commits/a2190f3a9f7ac684cd2a39a97eaff896e351eb73

rachelwhitton:
reviewed in docs coworking - ready to go, just bump reviewed dates

PR #9343, opened 2024-12-06
[UpdateTool - Terminus Information] Update commands and releases to version 3.6.1.

pantheon-upstream:
[UpdateTool - Terminus Information] Update commands and releases to the latest Terminus version.


jazzsequence:
@stevector @rachelwhitton  To follow up:

* these PRs are created through automation
* these PRs update a number of terminus-related things, including bumping the current release numbers in docs and the available commands
* these PRs _actually do_ update the latest terminus version and the changelogs, but the changelogs show notes from the _last_ version rather than the _current_ version -- still trying to figure that out

Unfortunately the PRs do not automatically trigger multidevs, so I had to create my own branch off this and push it to see the effect. However, I believe we can/should just merge them because overall I believe they are a net benefit. We (by which I mean "I") can continue to explore why automation is recording the wrong changelog notes.

PR #9348, opened 2024-12-12
Update supported-drupal.md

greg-1-anderson:
Reading through the existing text, it is not immediately clear that contrib LTS is separate from core LTS as an offering. This PR adds a bit of redundancy to emphasize that.

I am also pointlessly fiddling with the alignment of the checkmarks, which are offset on rows with popups.

## Summary

**[Doc Page Title](https://docs.pantheon.io/doc-title)** - Make support policy for D7 contrib LTS more clear.


jazzsequence (approved):
LGTM

PR #9374, opened 2025-01-14
Clarify header overflow for too many headers

jms-pantheon:
## Summary

**[5xx Level Errors](https://docs.pantheon.io/guides/errors-and-server-responses/5xx-errors#503-header-overflow)** - The GCDN has a hard limit on number of headers, and Pantheon uses a fair number so the CMS must stay under 40 additional headers. Previously we made no mention of this limit.


rachelwhitton:
@rachelwhitton to review and merge today

PR #9387, opened 2025-01-23
Fix sidebar nav for Errors and Server Responses 

rachelwhitton:
## Summary

Multipage guide, [Errors and Server Responses](https://docs.pantheon.io/guides/errors-and-server-responses) is nested too deep. This PR pulls the guide up a layer in the sidebar nav so we can see subpages too 

Instead of: 
Support > Troubleshooting > Timeouts and Errors > Errors and Server Responses

It's now: 
Support > Troubleshooting > Errors and Server Responses

### DIFF
--- a/src/components/omniSidebarNav/submenus/support.js
+++ b/src/components/omniSidebarNav/submenus/support.js
@@ -36,9 +36,9 @@ const support = () => {
         ]),
         simpleLink('/timeouts', 'Timeouts and Errors', [
           simpleLink('/timeouts', 'Timeouts on Pantheon'),
-          getGuideDirectory('guides/errors-and-server-responses'),
           simpleLink('/debug-connections', 'Debugging Connectivity Issues'),
         ]),
+        getGuideDirectory('guides/errors-and-server-responses'),
         simpleLink('/mime-types', 'MIME Types'),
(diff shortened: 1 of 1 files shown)


src/components/omniSidebarNav/submenus/support.js
```diff
@@ -36,9 +36,9 @@ const support = () => {
         ]),
         simpleLink('/timeouts', 'Timeouts and Errors', [
           simpleLink('/timeouts', 'Timeouts on Pantheon'),
-          getGuideDirectory('guides/errors-and-server-responses'),
           simpleLink('/debug-connections', 'Debugging Connectivity Issues'),
         ]),
```
rachelwhitton, line 40: @stevector @jazzsequence Does this layer still make sense without the multipage guide nested in it? Should the last 2 left here be moved up a layer as well?

First image shows the layer still intact w/2 left in it
Screenshot 2025-01-23 at 2 57 12 PM

compared to this one, which moves the whole thing up a layer 
Screenshot 2025-01-23 at 2 57 58 PM

stevector:
@rachelwhitton yeah, bring them all out. A submenu with only two children isn't worth keeping.

PR #9388, opened 2025-01-23
Update plugins-known-issues.md for Solid Security debugging

dylanbaumann:
The CSE team is cleaning up some of our playbooks and found documentation about a WordPress Plugin issue that was not talked about publicly. After reviewing within our colleagues, we decided this would be best handled by moving to the public facing documentation, so customers can potentially self-resolve the problem.

## Summary

**[WordPress Plugins and Themes with Known Issues](https://docs.pantheon.io/plugins-known-issues#solid-security-previously-ithemes-security)** - adding documentation for a known plugin issue with [Solid Security (former iThemes)](https://docs.pantheon.io/plugins-known-issues#solid-security-previously-ithemes-security), and the two known solutions to address the problem. 

The source is [this internal Pantheon CSE Playbook](https://getpantheon.atlassian.net/wiki/spaces/CS/pages/2860974234/Solid+Security+former+iThemes+Hide+Backend+URL+Forgot+Password+-+Retaining+itsec-hb-token+Parameter), which did not contain any CSE-only information, and was deemed suitable to move into the public docs to help customers self-resolve issues.

### DIFF
--- a/source/content/plugins-known-issues.md
+++ b/source/content/plugins-known-issues.md
@@ -626,6 +626,31 @@ ___
 
 **Solution:** Modifications to `wp-config.php` should be done in Dev or Multidev environments, then deployed forward to Test and Live.
 
+**Issue 3:** Unable to reset WordPress password or register new WordPress users when "Hide Backend" feature is enabled due to `itsec-hb-token` parameter being stripped from URL redirections.
+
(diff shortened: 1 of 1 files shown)


jazzsequence (approved):

jazzsequence:
This LGTM. Created an issue (linked above) in the Pantheon MU Plugin repo. The MU plugin compatibility layer already registers this plugin as being "Partially Compatible" on Pantheon with a link to that section of the docs, so users would already know that some amount of manual intervention is required, but I think we can probably get the actual code fixes integrated automagically.

PR #9389, opened 2025-01-23
Git reset Dev to Live for Integrated Composer sites

### LINKED ISSUE #9382: Git on Pantheon Guide Doc Update
Re: [Git on Pantheon Guide](https://docs.pantheon.io//guides/git/undo-commits)

Priority: Medium

## Issue Description:

https://docs.pantheon.io/guides/git/undo-commits#reset-dev-environment-to-live
This command only works if the site is not using Integrated Composer. If it is using Integrated Composer, this will force push the build artifact to dev/master, and that will cause Integrated Composer to fail with merge conflicts.

## Suggested Resolution

We probably need to clarify that this will not work with Integrated Composer. And maybe come up with some way to do the "git reset" to the 2nd-to-latest commit on live?

rachelwhitton:
Fixes #9382 

## Summary

**[Terminus > Examples](https://pr-9389-documentation.appa.pantheon.site/terminus/examples#reset-dev-environment-to-live)** 
**[Git > Undo Git Commits](https://pr-9389-documentation.appa.pantheon.site/guides/git/undo-commits#reset-dev-environment-to-live)** 

Add tabs recommending the appropriate procedure based on the given scenario: 
* Integrated Composer sites 
  * Updated procedure, resets dev to the second most recent commit from Live, avoiding build artifacts 
* All others 
  * No changes to the original procedure, still resets dev to the latest commit from Live, no build artifacts in play

### DIFF
--- a/source/content/guides/git/05-undo-commits.md
+++ b/source/content/guides/git/05-undo-commits.md
@@ -179,14 +179,12 @@ You can reset your Dev environment history to match the current state of your Li
 
 <Tab title="All others" id="all-others" active={true}>
 
-1. Identify the most recent commit deployed to Live.
+Run the commands below to overwrite the history on Dev's codebase to reflect Live (replace `<site>` with your site's name):
 
-1. Run the command below to overwrite the history on Dev's codebase to reflect Live (replace `<site>` with your site's name):
-
-    ```bash{promptUser: user}
-    git reset --hard `terminus env:code-log <site>.live --format=string | grep -m1 'live' | cut -f 4`
-    git push origin master -f
-    ```
+```bash{promptUser: user}
+git reset --hard `terminus env:code-log <site>.live --format=string | grep -m1 'live' | cut -f 4`
+git push origin master -f
+```
 
 </Tab>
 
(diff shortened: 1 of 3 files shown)


rachelwhitton:
@dficker can you review when convenient?

source/content/guides/git/05-undo-commits.md
```diff
@@ -174,6 +175,10 @@ You can revert a past commit that has been pushed to your Test or Live environme
 
 You can reset your Dev environment history to match the current state of your Live environment using [Terminus](/terminus). The method in this section is destructive and should be used with caution. It does not clone the Live environment's database or files down to Dev. However, it does reset the Dev environment's codebase.
 
+<TabList>
+
+<Tab title="All others" id="all-others" active={true}>
+
 1. Identify the most recent commit deployed to Live.
```
dficker, line 182: Actually, the Git/Terminus command does the identifying. So it seems like this is not really a "Step". I think it was added as a numbered list of steps by mistake a while back.

source/content/guides/git/05-undo-commits.md
```diff
@@ -183,6 +188,30 @@ You can reset your Dev environment history to match the current state of your Li
...
+
+<Tab title="Integrated Composer" id="integrated-composer">
+
+<Alert title="Note" type="info">
+
+We've adjusted the following steps for [Integrated Composer sites](/guides/integrated-composer), so that you reset history to the **second** to last commit hash on the Live environment, rather than the first - to avoid resetting dev's history to a build artifact.
+
+</Alert>
+
+1. Identify the **second** most recent commit deployed to Live.
```
dficker, line 201: Same here. Maybe it shouldn't be a numbered list of steps? Maybe just a version of this that is identifying

source/content/guides/git/05-undo-commits.md
```diff
@@ -183,6 +188,30 @@ You can reset your Dev environment history to match the current state of your Li
...
+
+1. Run the command below to overwrite the history on Dev's codebase to reflect Live (replace `<site>` with your site's name):
+
+    ```bash{promptUser: user}
+    git reset --hard `terminus env:code-log <site>.live --format=string | grep -m2 'live' | tail -n 1 | cut -f 4`
```
dficker, line 206: Nice! Yes, this command works as I would expect, it gets the second newest commit, and resets the local. So that's good!

dficker (commented):
Whoa, I didn't even know that that Terminus Examples page existed.

But the page I was looking at looks to make sense, though I do think that the 1, 2 numbered steps are more confusing and should be revised, or maybe we should put that in another PR/issue?

dficker (approved):
Looks good to me! Thanks!

PR #9398, opened 2025-01-28
SITE-3453 Update 2025-01-16-php-53-55-eol.md

scottbuscemi:
## Summary

**[PHP 5.3/5.5 EOL release note](https://docs.pantheon.io/release-notes/2025/01/php-53-55-eol)** - Note that PHP 5.6 can be tested now

### DIFF
--- a/source/releasenotes/2025-01-16-php-53-55-eol.md
+++ b/source/releasenotes/2025-01-16-php-53-55-eol.md
@@ -3,12 +3,13 @@ title: "PHP 5.3 and 5.5 no longer available starting February 25, 2025"
 published_date: "2025-01-16"
 categories: [infrastructure, security, action-required]
 ---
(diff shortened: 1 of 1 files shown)


source/releasenotes/2025-01-16-php-53-55-eol.md
```diff
@@ -11,6 +11,6 @@ We will soon provide long term support (LTS) security coverage for the remaining
```
rachelwhitton, line 11: I reviewed this one with the team during docs coworking and we'd like to add an editorial note describing how and why the entry was modified after initial publishing date. @scottbuscemi what do you think? 

```suggestion
_Editorial note: As of January 29, 2025, this release note has been edited to amend the action required. Where previously we required upgrading 5.3 and 5.5 to 7.2 we are now allowing an upgrade to 5.6 as well. This can be helpful if you choose to upgrade PHP versions incrementally rather than making the jump to 7.x_
```

I couldn't add the suggestion where I wanted so that's why it's randomly on line 11 - once copy is finalized it'll be relocated to the top of the entry in italics

PR #9405, opened 2025-01-30
Fix code block on Object Cache Pro page

jazzsequence:
I noticed that there was an issue with the code block inside the Note about Lando configuration with OCP. This PR fixes that.

## Summary

**[Enable Object Cache Pro for WordPress](https://docs.pantheon.io/object-cache/wordpress#local-configuration-with-lando)** - Local configuration with Lando section, specifically

[Preview](https://pr-9405-documentation.appa.pantheon.site/object-cache/wordpress#local-configuration-with-lando)


jazzsequence:
My suspicion is that the issue has to do with the nested Alert inside the numbered list. I've removed the indent so the code block is interpreted (more) normally and hoping that does the trick.

ccharlton (commented):
Wouldn't the conditional snippet be more flexible if it was checking simply for 'Not Pantheon Environments' versus sniffing for Lando itself?

PR #9418, opened 2025-02-05
Restructure SSH keys documentation

### LINKED ISSUE #9138: Generate and Add SSH Keys Doc Update
Re: [Generate and Add SSH Keys](https://docs.pantheon.iossh-keys)

Priority: High

## Suggested Resolution

**This is for Windows users:**

Recommend installing Git Bash into their Local Windows Device: https://git-scm.com/download/win
Once done, follow the commands in Mac/Linux: https://docs.pantheon.io/ssh-keys#macoslinux

Install and run git bash and follow:

MacOS/Linux
Open your terminal and enter the following command to generate a key:

`ssh-keygen -t rsa`
Please do not edit the default location of ~/.ssh/id_rsa unless you have a reason to change it. If the command says the key already exists, you can overwrite it, or continue to the next step with your existing key.

Set a passphrase for better security.

We recommend using a passphrase, but it can conflict with some tools.

Copy the contents of `~/.ssh/id_rsa.pub` to your clipboard after the files are created.

MacOS users can cat the file to the terminal and copy the output:

`cat ~/.ssh/id_rsa.pub`
Run eval to start the SSH agent. The Agent pid output confirms that the agent started:

Example: 

```
eval `ssh-agent`
Agent pid 86810
```

Run `ssh-keygen -lf ~/.ssh/id_rsa.pub -E md5` to check if the hash you have in your Pantheon User Dashboard's SSH Keys matches.

You are now set.

rachelwhitton:
Fixes #9138 

## Summary

**Generate and Add SSH Keys**:  
[Preview Link](https://pr-9418-documentation.appa.pantheon.site/ssh-keys) | [Production](https://docs.pantheon.io/ssh-keys) 

* Simplify steps to create key pairs locally 
  * Relocate non-WSL Windows steps to an accordion 
* Add subsection for managing multiple key pairs 
* Update instances where we talking about SSH keys being better than passwords, since SSH key is the only supported auth method 
* Use tabs for new and legacy dashboard steps (for both adding keys and revoking keys) 
* Rewrite subsection for testing keys

### DIFF
--- a/source/content/agency-tips.md
+++ b/source/content/agency-tips.md
@@ -77,7 +77,7 @@ Drupal now includes Composer usage within Drupal Core itself. For details, see [
 ## Advanced Tools
 
 ### Terminus
-[Terminus](/terminus), the Pantheon command-line interface, has a near 1:1 parity with the Pantheon Dashboard. After installing this tool locally, login using a [machine token](/machine-tokens). This allows for passwordless authentication, which is often necessary for bots, scripts and continuous integration.
+[Terminus](/terminus), the Pantheon command-line interface, has a near 1:1 parity with the Pantheon Dashboard. After installing this tool locally, login using a [machine token](/machine-tokens).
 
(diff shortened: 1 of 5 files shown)


jms-pantheon (approved):

jazzsequence (approved):
reorg looks _much_ better

Issue #9420, opened 2025-02-06
New page about Drupal CMS and known issues


stevector:
I'm seeing a need to explain the concept of Drupal CMS, how we expect it to be used, and known issues.

stevector:
Known issues include at least

* timeouts when selecting multiple recipes (even just more than one?) during installation.
* Project browser timeouts
* Project browser requires the file system to be writable (therefor only works in dev/mutlidev environments in SFTP mode)
* Often needing to wipe/start over

alex-moreno:
a good addition would be the AI timeouts. Most complex queries (ie create a content type) take longer than the standard timeout that we have by default in php, so you get a lot of errors that confuse people. "Is it that my api is not working?"

stevector:
> a good addition would be the AI timeouts. Most complex queries (ie create a content type) take longer than the standard timeout that we have by default in php, so you get a lot of errors that confuse people. "Is it that my api is not working?"



Let's get https://github.com/pantheon-systems/documentation/pull/9448 reviewed/merged and then handle AI timeouts in a follow up PR.

PR #9428, opened 2025-02-10
PR-9420 Added new DrupalCMS guide on using Project Browser.

alex-moreno:
## Summary

As per issue https://github.com/pantheon-systems/documentation/issues/9420


alex-moreno:
Feels that this probably should live inside a specific troubleshooting folder, maybe? Thoughts @stevector ?

stevector:
@alex-moreno I think we should close this PR in favor of what is now in #9448. Though I see at least one detail in this PR that I forgot to bring over to #9448. I'll do that now.

PR #9497, opened 2025-03-20
fix: round totalColumns up to ensure all categories are displayed

gfbollingerHakuna:
## Summary

Fix: Ensure All Categories Are Displayed by Rounding totalColumns Up
This PR ensures that totalColumns is always an integer by rounding it up with Math.ceil(). It also fixes the `Popover` container width issue that was causing columns to overflow on smaller resolutions.  

**Changes:**
- Applied Math.ceil() to totalColumns to avoid decimal values.
- Updated `itemsPerColumn` to 7 to avoid a single element in the last column (this change can be reversed if needed).
- Updated `pds-toolit-react` to version `1.0.0-dev.236` to have access to the `classNameContainer` prop in the ´Popover´ component.  
- Adjusted `Popover` styles and classes to fix the container width issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,12 +7,11 @@
     "": {
       "name": "pantheon-documentation",
       "version": "0.1.0",
-      "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
         "@mdx-js/mdx": "^1.6.22",
         "@mdx-js/react": "^1.6.22",
-        "@pantheon-systems/pds-toolkit-react": "^1.0.0-dev.228",
+        "@pantheon-systems/pds-toolkit-react": "^1.0.0-dev.236",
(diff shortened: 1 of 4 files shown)


mel-miller (commented):
Thanks @gfbollingerHakuna. This looks good to me. I'll leave it to @rachelwhitton for the final review though.

rachelwhitton (approved):
Ran local VRT with passing results, also manually tested across different templates. There are style changes here that make the dropdown much easier to read - labels that previously spanned multiple lines are now on a single line. Thank you!! +1  

I could stand for "Categories" heading in the dropdown to be a bit bigger but I don't feel too strongly there and defer to ya'll 

Screenshot 2025-03-28 at 3 03 15 PM


screencapture-localhost-9323-2025-03-28-15_02_19

PR #9494, opened 2025-03-19
Debug broken anchor

rachelwhitton:
Fixes #9490 
Replaces #9502 

Debug broken anchors across traffic intro page, for example: 

https://docs.pantheon.io/guides/account-mgmt/traffic#what-about-denial-of-service-and-other-attacks 

---

### Bug
The presence of a `Popover` component caused other elements to render incorrectly, wrapping them inside a div with pds-popover classes.

### Fix
Ensured Popover only renders on the client side using `useState` and `useEffect`, preventing issues during server-side rendering (SSR).

PR #9502, opened 2025-03-29
Fix Components rendering issue in MDX files

gfbollingerHakuna:
## Summary

The presence of a `Popover` component caused other elements to render incorrectly, wrapping them inside a div with pds-popover classes.

###Fix
Ensured Popover only renders on the client side using `useState` and `useEffect`, preventing issues during server-side rendering (SSR).

This avoids conflicts where MDX processing caused unintended DOM structure changes.


rachelwhitton:
Closed in favor of #9494

PR #9514, opened 2025-04-14
Solution added for Media bulk upload module issue on Pantheon.

sandipsmarty:
## Summary

**[Drupal Modules with Known Issues](https://docs.pantheon.io/modules-known-issues)** - Solution for Drupal Media Bulk Upload issue on Pantheon.

### DIFF
--- a/source/content/modules-known-issues.md
+++ b/source/content/modules-known-issues.md
@@ -268,9 +268,9 @@ ___
 ## [Media Bulk Upload](https://www.drupal.org/project/media_bulk_upload)
 <ReviewDate date="2024-06-24" />
 
-**Issue**: The default configuration for this module has been known to cause issues related to the "Upload location" setting, which out of the box uses the `/tmp` directory.
+**Issue**: The Media Bulk Upload module uses dropzoneJS which doesn't interact well with multi-container environments like Pantheon, where each web server uses its own local temp directory. The upload handler should not use local server directories. Instead, rely on the shared file system.
 
-**Solution**: Configure the "Upload location" to `public://sites/default/files/private` You must include the `public://` protocol for this solution to work.
+**Solution**: This issue can fixed by changing the `tmp_upload_scheme` configuration from `temporary` to `public` in the `dropzonejs.settings.yml` file.
 
 ___
 


rachelwhitton:
@sandipsmarty can you confirm the path set by the override has been created on the test and live environments? After [the patch](https://www.drupal.org/project/dropzonejs/issues/2916330#comment-13893203) is applied and the following config override added to settings.php, the folder (e.g., `/files/private`)  must be created manually via SFTP if it does not already exist on the given environment: 

```
// Ensure that DropZoneJS works in the live environment
// (which has multiple web heads).
$config['dropzonejs.settings']['tmp_upload_scheme'] = 'private';
```

sandipsmarty:
@rachelwhitton I am not sure about this patch, haven't tried it, also in my `settings.php` the above config which you mentioned doesn't exists. The solution about media bulk upload mentioned in the pantheon public facing doc doesn't work on Pantheon Test/Live envs. I faced this issue for EFL customer, recently migrated on Pantheon from Acquia. I did some research and found the alternate solution I mentioned in this PR which resolved the customer issue.

lcatlett:
That proposed patch only works for subset of use cases. The comments in the varios d.o issues about this are missing that the primary reason why dropzone has these challenges - it is because it doesnt chunk uploads, so what I recommend is to use this patch to fix the underlying issue https://www.drupal.org/project/dropzonejs/issues/3125682 and then a callout that if errors persist that the path should be updated to the non-temporary options on Pantheon that @sandipsmarty validated on current customer engagements

sandipsmarty:
I verified, this patch https://www.drupal.org/project/dropzonejs/issues/3125682 doesn't resolve the bulk media upload issue on Pantheon Test/Live environments. But changing the dropzonejs settings `['tmp_upload_scheme']` from `temporary` to `public` or `private` actually fixes the issue.

sandipsmarty:
@lcatlett @rachelwhitton 
Alternate solution of Media bulk upload would be https://www.drupal.org/project/media_library_bulk_upload/ along with a patch https://www.drupal.org/project/media_library_bulk_upload/issues/3461352. This module doesn't depend on  dropzoneJS, but it's developed based on Drupal's file upload mechanism. I have tested it out and it works fine, I had to disable dropzoneJS module.

PR #9515, opened 2025-04-14
Update  Tika versions

AronNovak:
## Summary

The current wording suggests that 3 versions are available. Only 1.21 and 1.18 are there.


AronNovak:
For the record, using `php:eval "system('cd /tmp && curl https://dlcdn.apache.org/tika/2.9.3/tika-app-2.9.3.jar -o /tmp/tika.jar');"`, version 2.9.3 already works very well on Pantheon, it would be the time to perhaps integrating it.

stevector:
@scottbuscemi, can you review?

scottbuscemi:
Thanks @AronNovak - we are working on updating our supported tika versions with the new PHP runtime rolling out. This would be a good update to the verbiage in the meantime. @stevector

stevector (approved):

PR #9516, opened 2025-04-14
Update external-libraries.md with tesseract

AronNovak:
## Summary

Add Tesseract as an available tool. It is already inside the containers.


stevector:
@scottbuscemi can you review?

scottbuscemi:
Hi @AronNovak - We are in the process of rolling out an updated PHP Runtime which will not include Tesseract so now wouldn't be a good time to document this package. I wouldn't suggest building on top of it as a result. Could you share more info about your use case for Tesseract?

rachelwhitton:
Closing based on above discussion, additional conversations/discussions much welcomed

PR #9525, opened 2025-04-16
d7 lts tag1 data retention notice

scottbuscemi:
## Summary

** Update the D7 LTS page to note how often a site needs to connect

### DIFF
--- a/source/content/supported-drupal.md
+++ b/source/content/supported-drupal.md
@@ -111,6 +111,16 @@ When updates are available, Drupal's Update Status module will notify you of the
 
 ![Tag1 D7ES available updates](../images/supported-drupal-d7es-update-status-security-alert.png)
 
+<Alert title="Data retention policy"  type="info" >
+
+##### Requirement for maintaining contrib module LTS coverage
+
+Tag1 only keeps 90 days of reported data from Pantheon-hosted Drupal 7 sites. The `tag1_d7es` module phones home to Tag1 when Drupal cron is executed. Under normal circumstances, it will call home every 24 hours and transmit a list of installed modules. If Drupal cron is not regularly executed, after 90 days all data for the site will be pruned. If the site was depending on security notification emails, these will no longer be emitted. 
+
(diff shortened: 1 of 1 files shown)


source/content/supported-drupal.md
```diff
@@ -111,6 +111,16 @@ When updates are available, Drupal's Update Status module will notify you of the
 
 ![Tag1 D7ES available updates](../images/supported-drupal-d7es-update-status-security-alert.png)
 
+<Alert title="Data retention policy"  type="info" >
+
+##### Requirement for maintaining contrib module LTS coverage
```
stevector, line 116:
```suggestion
##### Functioning cron is necessary for maintaining contrib module LTS coverage
```

We should maybe rewrite the following paragraph to lead with the action necessary instead of leading with the explanation and build to to the "so what"

source/content/supported-drupal.md
```diff
@@ -111,6 +111,16 @@ When updates are available, Drupal's Update Status module will notify you of the
...
+
+Tag1 only keeps 90 days of reported data from Pantheon-hosted Drupal 7 sites. The `tag1_d7es` module phones home to Tag1 when Drupal cron is executed. Under normal circumstances, it will call home every 24 hours and transmit a list of installed modules. If Drupal cron is not regularly executed, after 90 days all data for the site will be pruned. If the site was depending on security notification emails, these will no longer be emitted.
```
stevector, line 118:
```suggestion
LTS support depends on cron to communicate between the Drupal 7 site and Tag1's infrastructure. If cron is not executed regularly, the site may cease to receive updates.

Tag1 only keeps 90 days of reported data from Pantheon-hosted Drupal 7 sites. The `tag1_d7es` module phones home to Tag1 when Drupal cron is executed. Under normal circumstances, it will call home every 24 hours and transmit a list of installed modules. If Drupal cron is not regularly executed, after 90 days all data for the site will be pruned. If the site was depending on security notification emails, these will no longer be emitted. 
```

stevector (changes requested):
I made some suggested changes to make the language more direct and immediate that you need to run cron.

rachelwhitton (approved):

PR #9533, opened 2025-04-25
DS-1245 Refactor TOC component and update styles to make it sticky

### LINKED ISSUE #9480: Improve ux for WP known issues page
Re: [WordPress Known Issues](https://docs.pantheon.io/wordpress-known-issues) Due to the amount of entries in this doc it's hard to use. The right hand TOC experience cuts off unexpectedly so it cannot be used reliably, if we proceed with the following we should remove the righthand TOC from this page at the same time ([remove `showtoc: true` frontmatter in the markdown file](https://github.com/pantheon-systems/documentation/blob/main/source/content/wordpress-known-issues.md?plain=1#L6)) 

Priority: Low 

**If this is likely to be impacted by switching frameworks from Gatsby to NextJS - we'd rather hold work here until the framework switch is achieved.**

--- 

Evaluate potential to extend or re-use the json search used by the Terminus Manual's ["Command Directory"](https://docs.pantheon.io/terminus/commands#command-structure) for the [WordPress Known Issues](https://docs.pantheon.io/wordpress-known-issues) doc page 

Relevant files from the existing implementation: 
* [source/data/commands.json](https://github.com/pantheon-systems/documentation/blob/main/source/data/commands.json)
  * Terminus implementation data file `commands.json` is automatically maintained - by comparison, WP Known Issues usage would instead be generated from a manually updated source. We'd like to avoid overhead of maintaining two sources in diff formats (markdown/json) - ideally we would continue to manage the content from the markdown file source manually and the json source be converted/generated on build based on the contents of that markdown source. But converting this content to be maintained in json format in service of the better user experience would be acceptable. 
* [src/templates/terminusCommand.js](https://github.com/pantheon-systems/documentation/blob/main/src/templates/terminusCommand.js)
  * Terminus implementation generates a URL for each individual command using this template file. I'm not sure we need/want the WP Known Issues to follow that same behavior. Guidance welcomed  
* [src/components/commands/index.js](https://github.com/pantheon-systems/documentation/blob/main/src/components/commands/index.js)
  * Terminus implementation uses this component for the object search 

Image

gfbollingerHakuna:
Fixes #9480 
## Summary

- Refactored the TOC component, to accommodate long pages.
- Updated the CSS styles to make the TOC sticky, improving user navigation.
- Converted this component from a jQuery-based implementation to a React-based one.

### DIFF
--- a/src/components/toc/index.js
+++ b/src/components/toc/index.js
@@ -1,43 +1,49 @@
-import React, { useState, useEffect } from "react"
-import "./style.css"
+import React, { useState, useEffect } from 'react';
+import './style.css';
 
 const TOC = ({ title }) => {
-  const [initialized, setInitialized] = useState(false)
+  const [initialized, setInitialized] = useState(false);
 
   useEffect(() => {
     if (!initialized) {
-      const windowGlobal = typeof window !== "undefined" && window
+      const windowGlobal = typeof window !== 'undefined' && window;
 
       const settings = {
-        tocSelector: ".toc-placeholder",
-        contentSelector: ".doc, .terminus",
+        tocSelector: '.toc-placeholder',
+        contentSelector: '.doc, .terminus',
         orderedList: false,
-        headingSelector: "h2:not(.toc-ignore), h3:not(.toc-ignore)",
-        ignoreSelector: ".panel-title",
+        headingSelector: 'h2:not(.toc-ignore), h3:not(.toc-ignore)',
+        ignoreSelector: '.panel-title',
         hasInnerContainers: true,
-        extraListClasses: "nav nav-list",
-        listItemClass: "tocify-item",
-      }
+        extraListClasses: 'nav nav-list',
+        listItemClass: 'tocify-item',
+      };
 
       if (windowGlobal && windowGlobal.tocbot) {
-        jQuery(".tab-pane h2, .tab-pane h3").addClass("toc-ignore")
-        windowGlobal.tocbot.init(settings)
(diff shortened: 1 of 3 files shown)


rachelwhitton:
Overall feels like an improvement but still cuts off at "Force Login" on this page: 

https://pr-9533-documentation.appa.pantheon.site/wordpress-known-issues#force-login

gfbollingerHakuna:
@rachelwhitton I just saw it and I did a little change. Let me know what you think. Thanks!

rachelwhitton:
Just reviewed with the team and we all agree this is a huge improvement, thank you! @gfbollingerHakuna One request, can you hide the TOC for mobile visitors? Currently it just loads at the bottom of the page and we'd prefer it to be hidden for non-desktop users

gfbollingerHakuna:
@rachelwhitton I've hidden the TOC column for resolutions under 900px since it's the value used in this variable too: `MOBILE_MENU_BREAKPOINT`. Let me know your thoughts. Thanks!

rachelwhitton:
This is great! Thank you! I have just one final request, can you make the active state a bit more distinct from the rest? Currently there's not enough changing in the active state to make a correlation between where you are on the page in relation to the TOC (there's a small margin difference now which draws the eye but it's over so quick you can't really tell which item is active). I think it would be helpful if the background color and/or the font color changes for active state TOC items, not exactly this but similar: 

https://github.com/user-attachments/assets/a95b9118-5646-4811-aec6-6f393ad4e537

PR #9552, opened 2025-05-09
Update 07-site-plans.md, small change for interstitial header value

conorbauer23:
Updating the header value to 1 instead of true to match what GCDNs VCL is looking for

## Summary

**[Site Hosting Plans and Addons](https://docs.pantheon.io/guides/account-mgmt/plans/site-plans#interstitial-warning-pages)** - Updating the header value to 1 instead of true to match what GCDNs VCL is looking for


stevector (approved):
thanks @conorbauer23 !

PR #9560, opened 2025-05-21
php runtime gen2: clamav known issue, only for non live

scottbuscemi:
## Summary

** update PHP Runtime Generation 2 beta doc with a new known issue, and mark it as recommended only for non-live environments

### DIFF
--- a/source/content/php-runtime-generation-2.md
+++ b/source/content/php-runtime-generation-2.md
@@ -33,6 +33,8 @@ During the beta phase, you can opt in to test your sites on the new PHP runtime.
 - Identify any potential issues early
 - Provide valuable feedback to our team
 
+We currently recommend only using the PHP Runtime Generation 2 beta for testing with non-Live environments.
+
 ### How to Opt In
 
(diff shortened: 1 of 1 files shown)


source/content/php-runtime-generation-2.md
```diff
@@ -33,6 +33,8 @@ During the beta phase, you can opt in to test your sites on the new PHP runtime.
 - Identify any potential issues early
 - Provide valuable feedback to our team
 
+We currently recommend only using the PHP Runtime Generation 2 beta for testing with non-Live environments.
```
stevector, line 36: This should probably be a big bold warning at the top of the page, yes?

stevector (changes requested):

source/content/php-runtime-generation-2.md
rachelwhitton: I agree, done 👍

rachelwhitton (approved):

scottbuscemi:
gtg

PR #9607, opened 2025-07-11
php runtime gen2 - extension version updates

scottbuscemi:
## Summary

* Update extension versions for PHP Runtime Gen 2\

https://docs.pantheon.io/php-runtime-generation-2

### DIFF
--- a/source/content/php-runtime-generation-2.md
+++ b/source/content/php-runtime-generation-2.md
@@ -12,7 +12,7 @@ integration: [--]
 ---
 
 <Alert title="Warning" type="danger">
-We currently recommend only using the PHP Runtime Generation 2 beta for testing with non-Live environments.
+If your site uses Apache Tika, we currently recommend only using the PHP Runtime Generation 2 beta for testing with non-Live environments.
 </Alert>
 
 ## Overview
--- a/source/content/php-runtime-generation-2.md
+++ b/source/content/php-runtime-generation-2.md
@@ -25,7 +25,9 @@ Depending on your website's features, this change may have major effects on the 
 
 - Long Term Support (LTS) coverage for EOL PHP versions
 - Better compatibility with updated PHP extensions
-- PHP 8.4 availability
+- PHP 8.4 availability*
+
+ * For CMS compatibility details, see [Supported PHP Versions](/guides/php#supported-php-versions).
 
 ## Opt-In Beta Phase (April - August 2025)
 
--- a/source/content/php-runtime-generation-2.md
+++ b/source/content/php-runtime-generation-2.md
@@ -106,7 +106,8 @@ Does your application require an OS package or PHP extension that is no longer a
 ## Known Issues
 
 - ClamAV is currently unavailable.
-- Object Cache Pro installation via Terminus is currently unavailable.
+- [Object Cache Pro installation via Terminus for standard WordPress configurations](/object-cache/wordpress#installation-and-configuration) is currently unavailable. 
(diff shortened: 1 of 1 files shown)


source/content/php-runtime-generation-2.md
```diff
@@ -106,6 +106,7 @@ Does your application require an OS package or PHP extension that is no longer a
 ## Known Issues
 
 - ClamAV is currently unavailable.
+- Object Cache Pro installation via Terminus is currently unavailable.
```
rachelwhitton: @scottbuscemi is this referring to the [standard WP procedure](https://docs.pantheon.io/object-cache/wordpress#installation-and-configuration)? `terminus install:run <site>.<environment> ocp` 

Reviewing install docs it looks like standard WP procedure is the only one using Terminus to install OCP, no other frameworks are impacted by this known issue - is that accurate? 

Is there any workaround for this known issue?

rachelwhitton: ~Also wondering if this known issue should be discussed in the [release note](https://github.com/pantheon-systems/documentation/pull/9609/files)? Where we talk about CMS support?~

jazzsequence: > is this referring to the [standard WP procedure](https://docs.pantheon.io/object-cache/wordpress#installation-and-configuration)? terminus install:run <site>.<environment> ocp

Yes

> it looks like standard WP procedure is the only one using Terminus to install OCP, no other frameworks are impacted by this known issue - is that accurate?

yes. the issue, if I understand it correctly, has to do with the cloud run environments that are spun up to run the installation workflow. it's an open issue that SiteX is working on.

> Is there any workaround for this known issue?

I mean, composer-based installs would be unaffected (but no one is going to use Composer for just one plugin). Also, if you got the zip file for OCP some other way and committed it manually, it would work. But because it's a premium plugin, we don't just make that available (hence the Terminus command and automated workflow).

rachelwhitton: i guess one workaround could be to install OCP before upgrading to php runtime gen2?

rachelwhitton:
```suggestion
- [Object Cache Pro installation via Terminus for standard WordPress configurations](/object-cache/wordpress#installation-and-configuration) is currently unavailable. 
  - Workaround: [Downgrade to PHP Runtime Generation 1](/php-runtime-generation-2#q-can-i-switch-back-to-the-previous-php-runtime-if-i-encounter-issues), follow the installation procedure, then upgrade back to PHP Runtime Generation 2.
```

source/content/php-runtime-generation-2.md
```diff
@@ -25,9 +25,9 @@ Depending on your website's features, this change may have major effects on the
 
 - Long Term Support (LTS) coverage for EOL PHP versions
 - Better compatibility with updated PHP extensions
-- PHP 8.4 availability (coming soon)
+- PHP 8.4 availability
```
rachelwhitton: I think this page needs to crosslink to the updated table here: https://pr-9608-documentation.appa.pantheon.site/guides/php#supported-php-versions 
```suggestion
- PHP 8.4 availability*

 \* For CMS compatibility details, see [Supported PHP Versions](/guides/php#supported-php-versions).
```

source/content/php-runtime-generation-2.md
jazzsequence: Yes that would work. Probably also downgrading to gen 1 for the install would work, too.

pwtyler: could also downgrade in just a CDE and then merge back.

PR #9634, opened 2025-08-01
[SITE-4854] Adds release note for wkhtmltopdf to dompdf

AnaisPantheor:
Release not for the switch from wkhtmltopdf to dompdf with the introduction of PHP runtime generation


scottbuscemi:
I goofed up the git here. Creating a new PR.

PR #9635, opened 2025-08-01
update drupal redis patch links

### LINKED ISSUE #10077: Redis Module Patch Name Incorrect
Re: [Object Cache FAQs](https://docs.pantheon.io//object-cache/faq)

Priority: Low/Medium/High (choose one, remove the other options)

## Issue Description:

In the section about Patches to Drupal's Redis module, there's two issues that have the same name. The first one is correct and links to an issue called "TTL handling broken, always permanent". But the second one is a different issue.

## Suggested Resolution

The second link's text should be "Allow failing over to database or other backends if Redis is down".

scottbuscemi:
## Summary

https://www.drupal.org/project/redis/releases/8.x-1.10

The latest version of the Redis module fixes an issue that we were previously suggesting a patch for.

fixes #10077

### DIFF
--- a/source/content/addons/object-cache/reference/faq.md
+++ b/source/content/addons/object-cache/reference/faq.md
@@ -74,9 +74,8 @@ Our customers have encountered edge cases that may affect Redis performance unde
 
 | Issue Description | Patch Link |
 | -------- | ------- |
-| [TTL handling broken, always permanent](https://www.drupal.org/project/redis/issues/3179757) | [Patch](https://www.drupal.org/files/issues/2020-10-30/3179757-4.patch) |
 | [Always return array to mget to avoid breaking transactions](https://www.drupal.org/project/redis/issues/3216874) | [Patch](https://www.drupal.org/files/issues/2021-11-16/3216874-2.patch) |
-| [TTL handling broken, always permanent](https://www.drupal.org/project/redis/issues/3102739) | [Patch](https://www.drupal.org/files/issues/2023-07-11/3102739-28.patch) |
+| [Allow failing over to database or other backends if Redis is down](https://www.drupal.org/project/redis/issues/3102739) | [Patch](https://www.drupal.org/files/issues/2023-07-11/3102739-28.patch) |
 
 ### If Redis Hits the Upper Limit of Memory Usage, Is This Logged on Pantheon?
 


namespacebrian (approved):

lcatlett:
A few things folks should be aware of:

- This shouldn't be merged until the settings snippets are updated to account for new settings like https://git.drupalcode.org/project/redis/-/commit/ff3890a4da59101d0d59507979a22a017e5a7d66#e81e5be2706877eff882fb5ee8272aa85817cdac_41_42.

- This release is not just the patch that has now been tested extensively on Pantheon. There are changes to recommended defaults such as eviction policy that have never been tested with any customer set up. A significant example is `volatile-lfu` is now the recommended default eviction policy, so that's going to either need platform changes, or additional work in the form of a custom module to enforce.

- Eviction policy is one example but there are multiple critical changes that need to be tested both on setting up redis for the first time on a site, and upgrading existing sites to use these new capabilities / recs 

- the above needs to be coordinated with the d11 roadmap for cache enhancements. Settings like `$settings['cache_preload_tags']` are not compatible with the Drupal redis module currently https://www.drupal.org/node/3505248

stevector (changes requested):
Looks like we should address @lcatlett comments before merging https://github.com/pantheon-systems/documentation/pull/9635#issuecomment-3148612551

stevector:
@scottbuscemi, can you respond?

scottbuscemi:
@lcatlett Does anything you brought up affect this PR? Are you advocating for customers to not use the latest `redis` module release on Pantheon? I'm not understanding.

lcatlett:
> @lcatlett Does anything you brought up affect this PR? Are you advocating for customers to not use the latest `redis` module release on Pantheon? I'm not understanding.

@scottbuscemi To clarify: I'm not saying customers shouldn't use the latest Redis module release. I'm saying this PR, as written, could inadvertently guide customers into configurations that haven't been validated on Pantheon yet.

The issue is that the 8.x-1.10 release bundles more than just the patch fix. It also changes recommended defaults and introduces new settings. If we update the docs to point to this release without also updating the settings snippets, customers following our documentation could end up with:

- Eviction policy mismatches (`volatile-lfu` vs what the platform expects)
- Missing or incorrect settings that the new module version now requires
- Conflicts with upcoming D11 cache enhancements

When we publish configuration in our docs, customers reasonably assume it has been validated on Pantheon. That's the implicit contract of official documentation. So if we're going to point to this new release, the settings snippets also need to be updated to reflect tested, Pantheon-validated configurations for it.

So yes, this affects the PR directly. The doc changes need to be paired with updated settings snippets that account for the new module's requirements. Otherwise we're telling customers "use this new version" without giving them the correct configuration to go with it.

lcatlett:
An update: The D11 issues are more severe than I realized and redis cannot be used **at all** as a backend for D11 until critical regressions are fixed in the redis module AND the default settings.php config in Pantheon docs are updated. This impacts any Drupal 11 site attempting to use redis, not just sites on Pantheon. Sites that update to D11 without fixes will immediately WSOD and there is no workaround that can be deployed via settings.php to resolve this.

I filed an issue on behalf of Pantheon: https://www.drupal.org/project/redis/issues/3564915

lcatlett:
@scottbuscemi - Update on the upstream Redis module issue:

Since this is a blocking issue that could result in downtime for any pantheon customer using redis and D11, I've submitted new merge requests to the Redis module that properly fix the D11 container compilation WSOD issue with tests:

- **2.x branch:** [!80](https://git.drupalcode.org/project/redis/-/merge_requests/80)
- **8.x-1.x branch:** [!81](https://git.drupalcode.org/project/redis/-/merge_requests/81)

Full technical discussion and explanation posted here:
https://www.drupal.org/project/redis/issues/3564915#comment-16420043

The fix wraps the `\Drupal::database()` call in a try-catch for `ContainerNotInitializedException`, which handles the case where the container exists but services aren't fully initialized during D11 container compilation.

Once these MRs are merged upstream, we can update this documentation PR accordingly, as it will change the patch recommendations completely for all versions of redis. 

Note that this does not address any of the other documentation updates I listed in https://github.com/pantheon-systems/documentation/pull/9635#issuecomment-3676582476

PR #9638, opened 2025-08-05
php 8.4 auto upgrade to runtime gen2

scottbuscemi:
## Summary

* Update PHP 8.4 release note to state it automatically upgrades you to runtime gen2
* Add to PHP Runtime Gen 2 FAQ about downgrading

### DIFF
--- a/source/content/php-runtime-generation-2.md
+++ b/source/content/php-runtime-generation-2.md
@@ -149,7 +149,7 @@ php_runtime_generation: 1
 
 ### Q: After upgrading to PHP Runtime Generation 2, I removed php_runtime_generation from my pantheon.yml but the environment did not go back to Generation 1. How do I downgrade?
 
-If your environment is using PHP Runtime Generation 2, removing the `php_runtime_generation` parameter from your `pantheon.yml` does not automatically downgrade you back to `1`. To revert back to the first generation PHP runtime, set the following in your `pantheon.yml`:
+To revert back to the first generation PHP runtime, set the following in your `pantheon.yml`:
 
 ```yaml:title=pantheon.yml
 php_runtime_generation: 1
--- a/source/content/php-runtime-generation-2.md
+++ b/source/content/php-runtime-generation-2.md
@@ -17,7 +17,7 @@ If your site uses Apache Tika, we currently recommend only using the PHP Runtime
 
 ## Overview
 
-A new generation of our Serverless PHP Runtime is available in beta. This upgrade represents our commitment to providing a modern, secure, and efficient PHP runtime for your websites.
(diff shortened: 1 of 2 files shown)


source/content/php-runtime-generation-2.md
```diff
@@ -128,6 +128,17 @@ If you encounter any issues while testing your site on the new PHP runtime gener
...
+
+### Q: How do I opt out of the upcoming platform rollout?
+
+You may opt-out of the upcoming platform rollout by setting the following in your `pantheon.yml`:
+
+```yaml:title=pantheon.yml
+php_runtime_generation: 1
+```
+
+Note: All sites will be auto-upgraded in Q1 2026 if they haven't already, including sites that specify the opt-out above.
```
jazzsequence, line 140: is this too opaque for people who don't speak in quarters? Also, is it too ambiguous (financial quarters -- which is what's generally referred to by Q1/2, etc -- might be different for different organizations/parts of the world)?

source/content/php-runtime-generation-2.md
```diff
@@ -136,12 +147,10 @@ Yes, you may revert back to the first generation PHP runtime by setting the foll
 php_runtime_generation: 1
 ```
 
-### Q: How do I opt out of the upcoming platform rollout?
+### Q: After upgrading to PHP Runtime Generation 2, I removed php_runtime_generation from my pantheon.yml but the environment did not go back to Generation 1. How do I downgrade?
 
-You may opt-out of the upcoming platform rollout by setting the following in your `pantheon.yml`:
+If your environment is using PHP Runtime Generation 2, removing the `php_runtime_generation` parameter from your `pantheon.yml` does not automatically downgrade you back to `1`. To revert back to the first generation PHP runtime, set the following in your `pantheon.yml`:
```
pwtyler: I think this behavior has changed as of today. Need to confirm before we publish.

pwtyler: Ok, confirmed. If a site is set to gen 2, and then removed from pantheon.yml, it will change back to _the site environment's default value_, that is, today it will fall back to gen 1.

In the future, if that underlying value changes (i.e. we rolled out an upgrade for a given site environment to gen 2), it will use the new value, so a site could remove `php_runtime_generation: 2` from pantheon.yml and still be on gen 2, or a site could remove `php_runtime_generation: 1` from pantheon.yml and find themselves on gen 2.

rachelwhitton:
```suggestion
To revert back to the first generation PHP runtime, set the following in your `pantheon.yml`:
```

rachelwhitton (approved):

PR #9643, opened 2025-08-06
Update Tin Canny Reporting issue

jazzsequence:
Update Tin Canny known issue to report a specific version

## Summary

**[WordPress Known Issues](https://docs.pantheon.io/wordpress-known-issues)** - Uncanny Owl has reported that the fix should be in versions above 5.1.0.3. This PR updates the note to say that a fix does not exist _yet_ and gives an (expected) version number for support. We can drop this or update again when the updated version is released to just say "update to the latest version".

### DIFF
--- a/source/content/wordpress-known-issues.md
+++ b/source/content/wordpress-known-issues.md
@@ -998,11 +998,11 @@ ___
 
 ### Tin Canny Reporting
 
-<ReviewDate date="2025-07-21" />
+<ReviewDate date="2025-08-06" />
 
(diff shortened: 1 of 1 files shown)


rachelwhitton:
@jazzsequence to confirm issue is resolved in newest release and if confirmed remove issue/solution entirely

jazzsequence:
Just reviewed the email thread with Uncanny Owl. The patch did _not_ make it into 5.1.0.3, so the text as it is written is still accurate. Going to merge this and we can update when the fix has been released.

PR #9671, opened 2025-09-09
New sites use PHP Runtime Generation 2 [Sept 10]

scottbuscemi:
## Summary

* Release Note announcing new sites use PHP Runtime Generation 2

Hold for:

* [ ] @scottbuscemi go-ahead
* [ ] @pwtyler go-ahead


rachelwhitton (approved):
Noticed that the URL was competing with the Sept 5th entry and updated the filename so we have 2 distinct slugs. Otherwise looks good! thank you!

PR #9685, opened 2025-09-16
BUGS-10162 Document the OCP recommended config constant

scottbuscemi:
## Summary

* Object Cache Pro - document the recommended config constant

### DIFF
--- a/source/content/addons/object-cache/howto/wordpress.md
+++ b/source/content/addons/object-cache/howto/wordpress.md
@@ -238,16 +238,17 @@ Refer to the [official Object Cache Pro documentation](https://objectcache.pro/d
 		'port' => getenv('CACHE_PORT') ?: 6379,
 		'database' => getenv('CACHE_DB') ?: 0,
 		'password' => getenv('CACHE_PASSWORD') ?: null,
-		'maxttl' => 86400 * 7,
-		'timeout' => 0.5,
-		'read_timeout' => 0.5,
+		'maxttl' => 86400,
+		'timeout' => 2.0,
+		'read_timeout' => 2.0,
+		'retry_interval' => 100,
 		'split_alloptions' => true,
 		'prefetch' => true,
 		'debug' => false,
 		'save_commands' => false,
 		'analytics' => [
 			'enabled' => true,
-			'persist' => true,
+			'persist' => false,
 			'retention' => 3600, // 1 hour
 			'footnote' => true,
 		],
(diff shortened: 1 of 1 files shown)


jazzsequence (approved):

source/content/addons/object-cache/howto/wordpress.md
```diff
@@ -371,6 +372,45 @@ Lando's [Pantheon recipe](https://docs.lando.dev/plugins/pantheon/) includes Red
...
+	'prefetch' => true,
+	'debug' => false,
+	'save_commands' => false,
+	'analytics' => [
+		'enabled' => true,
+		'persist' => false,
+		'retention' => 3600, // 1 hour
+		'footnote' => true,
+	],
+	'prefix' => $ocp_redis_prefix,
```
rachelwhitton, line 406: @scottbuscemi there are a few differences in this config and our recommended config for Composer managed sites, specifically `token` and `prefix` 

```
	'token' => $token,
```

```
	'prefix' => "ocppantheon", // This prefix can be changed. Setting a prefix helps avoid conflict when switching from other plugins like wp-redis.
```

I understand why token is different, given the install steps in the Composer tab. But does it matter that prefix is different? Should we indicate somewhere (in either the intro paragraph or an alert) that the config given is for non-composer managed sites specifically? 

I defer to @scottbuscemi and @jazzsequence , if ya'll think this isn't an issue please feel free to dismiss

jazzsequence: @rachelwhitton Functionally it's not different. We don't commit it to an upstream, but the normal workflow (using the terminus command on non-composer sites) creates a `wp-config-ocp.php` file that defines `$ocp_redis_prefix = "ocpantheon"`. The update here is just making clear what the settings are. It _might_ be beneficial for full disclosure to show that in the snippet though, e.g. what's actually in the file:

```php
<?php

if ( ! defined( 'WP_REDIS_CONFIG' )) {
	/*
	* https://objectcache.pro/docs/configuration-options/#prefix
	* This prefix can be changed. Setting a prefix helps avoid conflict when
	* switching from other plugins like wp-redis. The prefix should also be
	* changed when switching between serialization or compression types to
	* prevent fatal errors.
	*/
	$ocp_redis_prefix = "ocppantheon";

	define( 'WP_REDIS_CONFIG', [
		'token' => getenv('OCP_LICENSE') ?: null,
		'host' => getenv('CACHE_HOST') ?: '127.0.0.1',
		'port' => getenv('CACHE_PORT') ?: 6379,
		'database' => getenv('CACHE_DB') ?: 0,
		'password' => getenv('CACHE_PASSWORD') ?: null,
		'maxttl' => 86400 * 7,
		'timeout' => 0.5,
		'read_timeout' => 0.5,
		'split_alloptions' => true,
		'prefetch' => true,
		'debug' => false,
		'save_commands' => false,
		'analytics' => [
			'enabled' => true,
			'persist' => true,
			'retention' => 3600, // 1 hour
			'footnote' => true,
		],
		'prefix' => $ocp_redis_prefix,
		'serializer' => 'igbinary',
		'compression' => 'zstd',
		'async_flush' => true,
		'strict' => true,
	] );
}

```

rachelwhitton (approved):

PR #9742, opened 2025-10-28
Update wordpress-known-issues.md

dficker:
Customer suggested these fixes via ZD ticket. I haven't tested the code, but it seems like it makes sense to me that it needs to be tweaked.

## Summary

**[WordPress Known Issues](https://docs.pantheon.io/wordpress-known-issues)** - Fix a few syntax errors in sample code.


rachelwhitton:
Reviewed with @jazzsequence on zoom, merging now. Thanks @dficker!

PR #9821, opened 2025-12-30
Removing duplicative templates

stevector:
In this pull request I am cleaning up duplicative templates and React components. It should be safe to merge if their are no visual regressions.


stevector:
@rachelwhitton I'm adding you as assignee for PRs that are ready to merge. The CI failure here is 1 false positive on visual regression.

rachelwhitton (approved):

PR #9833, opened 2026-01-16
Fix formatting in 05-custom-domains.md

incredimike:
Correct formatting issues in the custom domains guide.

https://docs.pantheon.io/guides/domains/custom-domains

The current formatting looks like this:
screenshot_451


rachelwhitton:
Thank you!!

PR #9847, opened 2026-01-29
Add MariaDB version and update link to introduction

T-Fletcher:
Added information about default MariaDB versions for new environments and upgrade requirements.

### DIFF
--- a/src/source/content/guides/mariadb-mysql/01-introduction.md
+++ b/src/source/content/guides/mariadb-mysql/01-introduction.md
@@ -35,13 +35,15 @@ MariaDB provides a full database solution, including:
 
 The default database version for new sites is MariaDB `10.4`.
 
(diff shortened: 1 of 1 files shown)


jazzsequence:
Hey @T-Fletcher thanks for the PR. The note that Pantheon doesn't support MariaDB 10.5+ is inaccurate, though. Pantheon supports 10,3, 10.4 or 10.6 https://docs.pantheon.io/pantheon-yml#specify-a-version-of-mariadb

The version of MariaDB that is shipped by default is dependent upon what's in `pantheon.upstream.yml` -- it's not a given that it's 10.4 by default. See [this example in WordPress (Composer Managed)](https://github.com/pantheon-systems/wordpress-composer-managed/blob/default/pantheon.upstream.yml) in which MariaDB defaults to 10.6.

scottbuscemi:
@rachelwhitton Why did we merge this? 

>Pantheon currently does not support MariaDB `10.5` or later. This may change in the future.

This is not accurate

PR #9849, opened 2026-01-30
[Release note] Content Publisher CLI Rename and other updates

rachelwhitton:


rolandbenedetti73:
Minor comments
First:
> The App Router starter kit no longer uses Static Site Generation (SSG) by default

SSG was not used on all pages I think,  would it be better phrased

> The App Router starter kit default installation no longer uses Static Site Generation (SSG) 
the nuance might be small but the first somehow implies everything is SSG. Hoping this makes sense...

Second:
> We've renamed the Content Publisher CLI from pcc to cpub to reflect our product branding

would it be more accurate:

> ... reflect our product naming

PR #9859, opened 2026-02-05
A tags in sidebar

stevector:
The live site is currently struggling to load pages. Pages loaded via RSC/link tags seem particularly problematic. This could be a helpful band-aid.


rachelwhitton:
@stevector is this okay to merge? 

reviewed in todays docs coworking and seems to work the same as current prod

stevector:
No, I'll close. this might have been a mitigation for unreliability we had two weeks ago.

PR #9872, opened 2026-02-11
Add Claude local settings to gitignore

scottbuscemi:
## Summary
- Adds `.claude/settings.local.json` to gitignore to prevent Claude Code local settings from being committed

## Test plan
- Verify `.claude/settings.local.json` is properly ignored by git
- Confirm no other files are affected by this change

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,7 @@ out
 /public/current-sha.txt
 
 pnpm-lock.yaml
+
+# Claude
+
+.claude/settings.local.json


stevector:
Would we advise our customers to put in the `.gitignore` of each of their repos are of their global settings? I asked Claude and got this response:

```
This should happen at the global level of your machine.   
                                                            
  Why global:                                               
  - .claude/settings.local.json is a user/machine-specific  
  file for the Claude Code tool, not a project artifact     
  - It will appear in every repo where you use Claude Code  
  - Similar to .DS_Store, .idea/, or .vscode/settings.json —
   these are conventionally globally ignored rather than    
  added to every project's .gitignore                       
                                                            
  How to set it up:                                         
                                                            
  1. Create or edit your global gitignore file:             
  git config --global core.excludesfile ~/.gitignore_global 
  2. Add the entry to ~/.gitignore_global:                  
  .claude/settings.local.json                               
                                                            
  This keeps your project .gitignore files focused on       
  project-specific ignores and means you only configure it  
  once for all repos. 
```

jazzsequence:
I agree. This should be in a global `.gitignore` or excluded from the local repository in `.git/info/exclude`

PR #9885, opened 2026-02-14
Update wordpress-known-issues.md

danreb:
Added a new known issue for the Yoast SEO 25.3 release, which is incompatible with Pantheon

### DIFF
--- a/src/source/content/wordpress-known-issues.md
+++ b/src/source/content/wordpress-known-issues.md
@@ -1787,6 +1787,15 @@ ___
 
 ___
 
+### Yoast LLMS feature
+
+<ReviewDate date="2026-02-14" />
+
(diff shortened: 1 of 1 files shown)


rachelwhitton:
@jazzsequence can you take a look at this one?

jazzsequence:
The issue is Pantheon's read-only filesystem. Yoast writes `llms.txt` to the root directory and keeps it up-to-date as content changes. Without write access in production, the cron jobs would fail even if the `llms.txt` was created in the dev environment initially, meaning the content of the file would fall out of sync.

The solution is to symlink `/llms.txt` to a writeable path (e.g. `/wp-content/uploads/`). Yoast provides a filter to change where the path for `llms.txt` is saved for this purpose. 

This is not a hard conflict. A workaround exists. Updated the PR to include the workaround.

jazzsequence (approved):

PR #9890, opened 2026-02-19
SITE-5518 update tika 3 documentation

tundeafolabi:
- Document the Tika 3.x OCR behavior change (AUTO mode) and its impact on PDF processing times
  - Add path to the bundled tika-config.xml that disables OCR and usage with the --config flag

### DIFF
--- a/src/source/content/external-libraries.md
+++ b/src/source/content/external-libraries.md
@@ -63,7 +63,7 @@ Valid values are `3` or `none` (to disable Tika).
 
 #### OCR in Tika 3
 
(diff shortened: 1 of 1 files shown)


src/source/content/external-libraries.md
```diff
@@ -61,6 +61,14 @@ tika_version: 3
 
 Valid values are `3` or `none` (to disable Tika).
 
+#### OCR in Tika 3
+
+Tika 3.x defaults to `AUTO` OCR mode, which can significantly increase PDF processing times when Tesseract is available. A `tika-config.xml` that disables OCR is available at `/opt/pantheon/tika/tika-config.xml` and can be passed to Tika using the `--config` flag:
```
scottbuscemi: "which can significantly increase PDF processing times when Tesseract is available"

Tesseract is always available... Should we say "which can significantly increase PDF processing times when Tesseract is used for OCR" or something along those lines?

PR #10016, opened 2026-04-13
Terminus 4 Linux Install Goes To Old Version

dficker:
Update it to not use Terminus 4.1.1  and instead use the latest version of Terminus.


src/source/content/terminus/02-install.md
```diff
@@ -111,7 +111,7 @@ The commands below will:
 
 ```bash{promptUser: user}
 mkdir -p ~/terminus && cd ~/terminus
-curl -L https://github.com/pantheon-systems/terminus/releases/download/4.1.1/terminus.phar --output terminus
+curl -L https://github.com/pantheon-systems/terminus/releases/latest/download/terminus.phar --output terminus
 chmod +x terminus
 ./terminus self:update
```
greg-1-anderson: We can remove the `self:update` call here, since we're downloading the latest version now.

PR #10108, opened 2026-06-05
cookies: warn against high-cardinality STYXKEY values

### LINKED ISSUE #10107: cookies.md: STYXKEY guidance omits value-cardinality warning and dual-purpose behavior
The [Cookies and Caching](https://docs.pantheon.io/cookies) page documents the `STYXKEY` *mechanism* (the `STYXKEY[a-zA-Z0-9_-]+` naming rule and cache-varying behavior) but does not state two operational guardrails that readers currently only learn through support. A reader following the doc literally has no signal before degrading their own cache.

## Gap 1: No value-cardinality warning

A `STYXKEY` cookie value becomes part of the Edge cache key, so **each distinct value produces a separate cached object**. The page never says this. A reader who places a high-cardinality or user-unique value (a user ID, a session token, a JWT) in a `STYXKEY` cookie fragments the cache to one entry per value, which effectively disables caching for those responses and consumes cache capacity, the opposite of the intended outcome.

The page has no language about cardinality, cache fragmentation, or hit ratio.

## Gap 2: Dual-purpose behavior is not called out

`STYXKEY` does two things at once:

1. It varies the Edge cache key by the cookie value.
2. It allows the cookie to survive through to the application (cookies that match neither a cache-busting pattern nor `STYXKEY` are normalized away for caching purposes).

Because of (2), `STYXKEY` is sometimes reached for purely as a way to forward a cookie to the CMS/application. That is a legitimate use, but it silently changes cache behavior via (1). If the forwarded value is user-unique, the cache is fragmented per user without the author realizing it. The two uses share one mechanism and pull in opposite directions, and the doc does not warn about this.

## Suggested fix

Add a short note/warning to the **Cache-Varying Cookies** section, for example:

> Use `STYXKEY` cookies only for low-cardinality values (for example role, locale, plan, or A/B test bucket). Each distinct cookie value creates a separate cached object at the Edge, so a high-cardinality or user-unique value effectively disables caching for those responses. `STYXKEY` both varies the cache key and forwards the cookie to your application, so do not use it as a generic cookie passthrough for per-user values.

I will open a follow-up PR with this change.

marfillaster:
Closes #10107

## Problem

The [Cookies and Caching](https://docs.pantheon.io/cookies) page documents the `STYXKEY` mechanism but omits two operational guardrails that readers currently only learn through support:

1. **No value-cardinality warning.** Each `STYXKEY` cookie value becomes part of the Edge cache key, so each distinct value produces a separate cached object. A user-unique value (user ID, session token, JWT) fragments the cache to one entry per value, effectively disabling caching for those responses.
2. **Dual-purpose behavior is not called out.** `STYXKEY` both varies the cache key and forwards the cookie to the application. It is sometimes reached for purely to forward a cookie to the CMS, which silently degrades cache behavior.

## Change

Adds a single `Warning` callout at the end of the **Cache-Varying Cookies** section covering both points. Content-only; no structural changes.

Screenshot 2026-06-05 at 11 53 54 PM

### DIFF
--- a/src/source/content/cookies.md
+++ b/src/source/content/cookies.md
@@ -124,7 +124,7 @@ If you prefer a regular expression, the name must match the following format: `S
 
 Use `STYXKEY` cookies only for low-cardinality values (for example, role, locale, plan, or A/B test bucket). Each distinct cookie value becomes part of the Edge cache key and produces a separate cached object, so a high-cardinality or user-unique value (such as a user ID, session identifier, or JWT) fragments the cache to one entry per value. This effectively disables caching for those responses and consumes cache capacity, the opposite of the intended outcome.
 
-The `STYXKEY` prefix serves two purposes at once: it varies the Edge cache key by the cookie value, and it allows the cookie to pass through to your application. Do not reach for `STYXKEY` purely as a way to forward a per-user cookie to the CMS, as doing so will fragment your cache.
+`STYXKEY` cookies are not stripped at the Edge on cache misses. The cookie value is forwarded to your application to generate the cached variant.  Using `STYXKEY` cookies to forward per-user cookies to the CMS requires a unique value per user, which fragments your cache as described above. To forward a cookie to your application on every request regardless of cache state, use a [cache-busting cookie](/cookies#cache-busting-cookies) instead.
 
 </Alert>
 
--- a/src/source/content/cookies.md
+++ b/src/source/content/cookies.md
@@ -120,6 +120,14 @@ If you prefer a regular expression, the name must match the following format: `S
 
 1. Use the stored value to generate varied content as appropriate for your use case and implementation.
 
+<Alert title="Warning" type="danger">
+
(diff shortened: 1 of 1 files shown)


src/source/content/cookies.md
```diff
@@ -120,6 +120,14 @@ If you prefer a regular expression, the name must match the following format: `S
 
 1. Use the stored value to generate varied content as appropriate for your use case and implementation.
 
+<Alert title="Warning" type="danger">
+
+Use `STYXKEY` cookies only for low-cardinality values (for example, role, locale, plan, or A/B test bucket). Each distinct cookie value becomes part of the Edge cache key and produces a separate cached object, so a high-cardinality or user-unique value (such as a user ID, session identifier, or JWT) fragments the cache to one entry per value. This effectively disables caching for those responses and consumes cache capacity, the opposite of the intended outcome.
+
+The `STYXKEY` prefix serves two purposes at once: it varies the Edge cache key by the cookie value, and it allows the cookie to pass through to your application. Do not reach for `STYXKEY` purely as a way to forward a per-user cookie to the CMS, as doing so will fragment your cache.
```
jazzsequence:
```suggestion
`STYXKEY` cookies are not stripped at the Edge on cache misses. The cookie value is forwarded to your application to generate the cached variant.  Using `STYXKEY` cookies to forward per-user cookies to the CMS requires a unique value per user, which fragments your cache as described above. To forward a cookie to your application on every request regardless of cache state, use a [cache-busting cookie](/cookies#cache-busting-cookies) instead.
```

jazzsequence (approved):

PR #10122, opened 2026-06-18
Clarify billing policy for annual to monthly switches

carly-armstrong:
Updated the billing policy for annual plans switching to monthly billing to clarify immediate effect and prorated credits.


carly-armstrong:
Added Change to logic about billing terms

rachelwhitton (approved):

PR #10165, opened 2026-07-10
docs(EDRT-9210): platform vanity domains supported on GCDN Beta

john-lauro:
## What

Updates the GCDN Beta guide to reflect that **platform vanity domains are now supported** on the new GCDN.

- Remove the `Platform Vanity Domains | Not yet supported` row from the Eligibility table.
- Flip the FAQ *"I use a platform vanity domain. Can I migrate?"* from "Not yet" to **Yes** — supported via a CSE-assisted migration (not the self-serve dashboard flow).

## Why

Vanity-domain support on the new GCDN is now live and validated in production (epic EDRT-8780). The page previously stated the opposite in two places, which is now stale.

Jira: EDRT-9210 (epic EDRT-8780)


djschaap (approved):
🚀

PR #10191, opened 2026-08-03
set html lang for a11y

rachelwhitton:
~* Sets min-height and min-width for pds buttons to 44px~ 
* Sets html language to english


rachelwhitton:
@mel-miller FYI we're doing this based on a SiteImprove a11y review. Let me know if you'd like more info on those reports

jazzsequence (approved):

PR #10195, opened 2026-08-05
Add content encoding

rachelwhitton:


jazzsequence (approved):
seems legit. make sure the multidev doesn't break and check the headers on that to validate

rachelwhitton:
We want to first upgrade docs.pantheon.io to nextgen gcdn, adding blocked label until that can be scheduled 


Currently, no response header is being returned despite this config working locally

PR #10209, opened 2026-08-10
Bring Terminus docs data automation into this repo

pwtyler:
## Why

After [updatinator#56](https://github.com/pantheon-systems/updatinator/pull/56) moves the WordPress jobs out, the only job left in that repo is the nightly `terminus-update-docs` cron — and it has been broken for months.

`update-tool terminus:update-docs` writes to **hardcoded** `source/data/...` paths. The Next.js migration (#9816, 2025-12-16) moved everything under `src/`, so those paths no longer exist. Consequences:

- Last successful bot PR was #9754 (Terminus 4.1.1, 2025-11-06).
- The gap went unnoticed until it was filled by hand — e.g. "Manually update commands.json for Terminus 4.2.0" (#10015), and most recently #10233/#10231 catching both files up to 4.3.3.
- Without this automation, every Terminus release means another manual regeneration, and it's easy for that step to get missed.

Terminus, update-tool, and documentation are all public, and the job's only privileged action is opening a PR **against this repo** — which the built-in `GITHUB_TOKEN` can do. Today it instead mints a Vault-backed GitHub App token over OIDC, using an App that per updatinator's README isn't even installed on this repo. Moving it here drops that dependency and retires the last reason updatinator exists.

## What

- `scripts/terminus-docs.ts` — regenerates `src/source/data/commands.json` (from `terminus.phar list --format=json`) and `src/source/data/terminusReleases.json` (from the Releases API).
- `.github/workflows/terminus-docs-update.yml` — nightly (`0 0 * * *`, the slot updatinator used) plus `workflow_dispatch` with an optional pinned tag. Opens/updates a PR on branch `terminus-docs/<version>`.

Release notes and the version/EOL table in `10-supported-terminus.md` still need judgment and stay hand-written, so this doesn't conflict with step 5 of the terminus-release runbook.

## Details worth reviewing

- **Byte-exact formatting.** Reproduces PHP's `json_encode` output (escaped slashes; 2-space for commands, 4-space for releases, no trailing newline on releases), so diffs show only real content changes.
- **Determinism.** Symfony interpolates the phar's absolute path into the `help`/`list`/`completion` help text. That differed per machine and would churn on every run, so it's normalized to `terminus`.
- **Release cap of 40.** Paginating fully returns 118+ releases back to 2015 and inflates the file well past what the changelog component ever renders (last year only). 40 keeps well over a year of history with room for the current release cadence. The PHP tool kept 28 by accident — one unpaginated call.
- **`02-install.md` is deliberately untouched.** The old tool regex-bumped a pinned download URL there. The Terminus 4 tab now uses a floating `releases/latest/download/` URL, and the only pinned URL left is the **legacy Terminus 3** one, which must stay `3.6.2`. Carrying that logic forward could only corrupt it.
- Actions are SHA-pinned.

## Testing

This was exercised locally to create #10233.

## Follow-ups (separate PRs)

1. Remove `UpdateTerminusDocsCommands.php` from update-tool once this has produced a good PR.
2. Remove the `terminus-update-docs` job from updatinator, then archive the repo and retire the `ci-updatinator` Vault role.
3. Note in the terminus-release runbook that these two files are now automated.


rachelwhitton (approved):

pwtyler:
Addressed CodeQL alert #124 (`js/http-to-file-access`) by removing the inline `curl`-based phar download from `scripts/terminus-docs.ts` and installing Terminus via [`pantheon-systems/terminus-github-actions`](https://github.com/pantheon-systems/terminus-github-actions) instead.

- The workflow now resolves the target release once (`gh release view` or the pinned `workflow_dispatch` input) and passes that same tag to both the `terminus-github-actions` setup step and the script, so they can't diverge.
- `scripts/terminus-docs.ts` now invokes `terminus` directly from PATH and asserts the installed version matches `--terminus-release` before writing any files, rather than downloading/exec'ing a phar itself.
- This removes the flagged download-and-write pattern from this repo's scanned code entirely rather than dismissing the alert, and drops the `tmpdir`/`chmodSync` phar-management logic from the script.

Verified locally with `--dry-run` against Terminus 4.3.2 on PATH; `actionlint` and `tsc`/`prettier` pass on the changed files.

PR #10223, opened 2026-08-18
Fix link

rachelwhitton:
Since the target anchor is within a tab list, we can't use it. 

Linking to nearest header, aligning with supporting copy


rachelwhitton:
@jazzsequence can you review and merge?

stevector (approved):

PR #10240, opened 2026-08-28
Document SNI/Host match requirement for third-party CDNs in front of next-gen GCDN

conorbauer23:
## What

Adds documentation to the Next Generation Global CDN guide covering the requirements for placing a third-party CDN or reverse proxy in front of the next-generation GCDN:

- New section **Using a Third-Party CDN in Front of Pantheon** explaining that the fronting service must send a TLS SNI value matching the HTTP Host header, why mismatched requests are rejected with a 403 at the edge (domain fronting protection), and how to configure origin address, Host header, and SNI correctly. Points existing Cloudflare zone users to the O2O section.
- New **Known Limitations** entry documenting that Azure Front Door is not compatible with this architecture, because AFD always sets outbound SNI to the configured origin hostname with no independent override.

## Why

Support has hit this twice recently. Most recent case: a customer fronting a custom domain with Azure Front Door received 403s on every request and correctly isolated the SNI/Host mismatch as the cause (Slack: https://pantheon.slack.com/archives/C0A0Z8Y71NX/p1787683776779489). We had nothing citable in the docs stating the SNI/Host match requirement or the AFD limitation.

### DIFF
--- a/src/source/content/guides/global-cdn/06-next-gen-global-cdn.md
+++ b/src/source/content/guides/global-cdn/06-next-gen-global-cdn.md
@@ -6,7 +6,7 @@ description: Pantheon's next-generation GCDN introduces built-in bot protection.
 tags: [cache, cdn, security]
 contributors: [conorbauer, jazzsequence]
 showtoc: true
-reviewed: "2026-07-13"
+reviewed: "2026-09-02"
 permalink: docs/guides/global-cdn/next-gen-global-cdn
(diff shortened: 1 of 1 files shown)


rachelwhitton (approved):

rachelwhitton:
@jazzsequence this looks good to me, can you review and merge when convenient?

jazzsequence (approved):
lgtm :shipit:

PR #10243, opened 2026-08-28
Fix Terminus command directory spacing so names group with their own description

### LINKED ISSUE #10241: Terminus commands page: spacing makes command names group with the wrong description
## Summary

The styling of the Terminus commands table at https://docs.pantheon.io/terminus/commands regressed recently. The vertical spacing between each command name and its description is now larger than the spacing between a description and the *next* command name, so each command name visually groups with the preceding command's description rather than its own.

## Reported feedback

> The style of this docs page changed recently and in my opinion the table is much much much harder to parse visually than it was before. The spacing between the commands and their description makes it difficult to see which command goes with which description because the command name is actually closer to the prior command's description than its own description. Example: https://docs.pantheon.io/terminus/commands

## Impact

The commands list is the primary reference page for Terminus. In its current state a reader scanning the list can easily attribute a description to the wrong command, which is worse than a purely cosmetic regression — it can lead to running the wrong command.

## Likely origin

Recent Terminus docs data automation work is the most probable trigger. Not yet root-caused; noting for whoever picks this up:

- #10209 — Bring Terminus docs data automation into this repo (merged 2026-08-10)
- #10233 — Regenerate `commands.json` and `terminusReleases.json` for 4.3.3 (merged 2026-08-20). Catches `commands.json` up from 4.2.0 after the previous nightly automation had been broken for months, so this is the first large change to that data file in a long while.

`src/source/data/commands.json` feeds `src/templates/terminus.tsx` / `src/templates/terminus-command.tsx`. The regression may be in the regenerated data shape, in the rendering of it, or in unrelated CSS that landed in the same window — that determination is part of the fix work, not this report.

## Expected

Each command name and its description read as a single visual unit: intra-row spacing tighter than inter-row spacing, so the grouping is unambiguous at a glance.

pwtyler:
Fixes #10241

The Terminus command directory rendered each command name closer to the *previous* command's description than to its own, making it easy to attribute a description to the wrong command.

https://pr-10243-pandocs.pantheonsite.io/terminus/commands 

Screenshot 2026-09-02 at 1 08 32 PM

## Cause

Command name and description share one cell; the usage string sits in the adjacent cell:

```
<td> <Link .commandName>{name}</Link> <small>{description}</small> </td>
<td> <small>{usage[0]}</small> </td>
```

Two things combined:

- Cells inherit `vertical-align: middle` from the design system (`pds-core.css` sets `td,th{padding:.8rem .64rem}` and no alignment, so cells default to `middle`).
- The name/description pair was spaced by a fixed `margin-block-end: var(--pds-spacing-s)` on `.commandName`.

Row height is set by whichever cell is taller. When a usage string wraps to more lines than the left cell needs, the surplus height is distributed *around* the vertically centered name/description block. The gap between rows therefore grows with the usage column while the gap inside the pair stays pinned at `s`, inverting the visual grouping.

This is pre-existing fragility rather than something introduced by a specific change: `src/components/common/commands/` has no commits since the Next.js migration. The 4.3.3 data regeneration (#10233) exposed it by lengthening usage strings (max 287 → 351 chars; `site:create` +64, `vcs:connection:add` +50 from the new `--vcs-host` / `--vcs-token` flags). A long enough flag list would have triggered it regardless.

## Fix

Top-align the cells and group the name and description in a flex container, so their spacing comes from a `gap` that neighboring row height cannot exceed. The pairing now holds however tall the usage column makes a row, and the table is meaningfully shorter as a bonus.

Uses `--pds-spacing-2xs` (0.512rem) inside the pair against 0.8rem cell padding, so intra-pair spacing stays well under the ~1.6rem inter-row gap.

## Verification

`next build` compiles successfully. The build then fails on a pre-existing, unrelated type error in `src/components/footer/index.tsx:63` that reproduces on a clean `main` checkout with these changes stashed — not introduced here.

Rendered before/after against the real `commands.json` to confirm: in the current layout `art`, `completion`, and `help` each read as belonging to the preceding description; with the fix each name binds to its own.

## Related

#10242 opened a throwaway revert of #10233 to produce a preview build of the pre-regression layout for comparison. That PR should be closed, not merged — it would republish stale command data and does not address this cause.


rachelwhitton (approved):

rachelwhitton:
@mel-miller can you review when convenient?

mel-miller (approved):

PR #10250, opened 2026-09-03
docs(integrated-composer): [SITE-6176] add Composer plugins through a Custom Upstream

mehta-asim:
## Summary

- New section in the Integrated Composer Custom Upstream guide covering how to add a Composer plugin, such as `mglaman/composer-drupal-lenient`, through a Custom Upstream.

## Context

[SITE-6176](https://getpantheon.atlassian.net/browse/SITE-6176)

Composer loads plugins once at startup from the vendor directory already on disk, so a plugin that arrives in the same upstream update that first needs it never runs. Customers hit this trying to relax a contrib module's `drupal/core` constraint for a Drupal 11 upgrade. The guide covered upstream dependencies but not plugins.

The section documents the two-commit ordering, the root-only `config` and `repositories` properties, and the silent overwrite of site-local root `composer.json` edits.

## Do not merge yet

The platform-side fix is site-job-scripts#173, which installs the locked dependencies before the upstream merge. Until that ships, the update path has no vendor directory at all and this workflow still fails.

## Test plan

- [ ] Preview renders both Alerts and the JSON blocks
- [ ] Links resolve: `/core-updates`, `/guides/sftp`, `/guides/integrated-composer/ic-troubleshooting#changes-lost-during-upstream-updates`
- [ ] Review of the customer-facing wording

[SITE-6176]: https://getpantheon.atlassian.net/browse/SITE-6176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/source/content/guides/integrated-composer/06-ic-upstreams.md
+++ b/src/source/content/guides/integrated-composer/06-ic-upstreams.md
@@ -33,6 +33,55 @@ Follow the steps to [Create a Custom Upstream](/guides/custom-upstream/create-cu
 
 <Partial file="upstream-management-dependencies.md" />
 
+## Add a Composer Plugin to Your Upstream
+
+Composer loads plugin code once, when Composer starts, from the `vendor` directory that is already on disk. A plugin that arrives in the same update that first needs it is not on disk yet, so it does not run.
+
+Add the plugin to your upstream and deploy it to every site first, then make the change that depends on it. This example uses [`mglaman/composer-drupal-lenient`](https://github.com/mglaman/composer-drupal-lenient), which relaxes a contributed module's published `drupal/core` constraint.
+
+1. In the **root** `composer.json` of your Custom Upstream, require the plugin and allow it to run:
+
+    ```json:title=composer.json
+    "require": {
+        "mglaman/composer-drupal-lenient": "^1.0"
+    },
+    "config": {
+        "allow-plugins": {
+            "mglaman/composer-drupal-lenient": true
+        }
+    }
+    ```
+
+1. Commit and push the upstream, then apply the update to your sites and let each one build.
+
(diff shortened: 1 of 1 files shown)


jazzsequence (approved):
this lgtm, you fixed the thing I was going to comment about

src/source/content/guides/integrated-composer/06-ic-upstreams.md
```diff
@@ -33,6 +33,65 @@ Follow the steps to [Create a Custom Upstream](/guides/custom-upstream/create-cu
 
 <Partial file="upstream-management-dependencies.md" />
 
+## Add a Composer Plugin to Your Upstream
+
+Composer loads plugin code once, when Composer starts, from the `vendor` directory that is on disk at that moment. A plugin that arrives in the same upstream update that first needs it is not installed yet, so it does not run.
+
+Deploy the plugin to your sites in one upstream commit, then push the change that depends on it in a second commit. The example below uses [`mglaman/composer-drupal-lenient`](https://github.com/mglaman/composer-drupal-lenient), which relaxes a contributed module's published `drupal/core` constraint so that the module can install on a newer version of Drupal core.
```
g1a-pantheon: I don't think that you should say "commit" here; it sounds almost like you could just make two commits in the upstream and then run the upstream update command, which conflicts with what is below, which could cause confusion. I would recommend including standard Pantheon terms *upstream update*.

e.g.

> Add the plugin to your upstream in one release, and then run an upstream update on all sites created from that upstream. Once the plugin has been deployed everywhere, add the change that uses it in a later release.

PR #10256, opened 2026-09-09
EDRT-9481: Bot bypass tokens for Next Generation GCDN — docs and release note

conorbauer23:
## What

Replaces the "Bot Exclusions" section of the [Next Generation GCDN guide](https://docs.pantheon.io/guides/global-cdn/next-gen-global-cdn) with setup instructions for the new self-service bot bypass token, updates the matching FAQ entry, and adds the companion release note announcing the feature.

Bot bypass tokens let customers exempt their own trusted automation (uptime monitors, CI/CD, feed importers, API clients) from bot protection via a Terminus command, without contacting support.

## Product decision reflected here

Bot bypass tokens are the primary, self-service path. Customers who still need a manual exception (can't use the token, unusual constraint) are pointed at Pantheon support in both the guide and the FAQ.

## Follow-ups (not in this PR)

- Release note's `published_date`/`published_at` use today's date — update before merge if the actual announcement timing differs.

Refs EDRT-9481.

### DIFF
--- a/src/source/releasenotes/2026-09-09-gcdn-bot-bypass-token.md
+++ b/src/source/releasenotes/2026-09-09-gcdn-bot-bypass-token.md
@@ -1,0 +1,21 @@
+---
+title: "Self-service bot bypass tokens for Next Generation GCDN"
+published_date: "2026-09-09"
+published_at: "2026-09-09T20:33:13Z"
+categories: [new-feature, tools-apis]
+description: "Generate a per-site token with Terminus to exempt your own trusted automation from Next Generation GCDN bot protection, without contacting support."
+---
+
+Sites on the [Next Generation GCDN](/guides/global-cdn/next-gen-global-cdn) can now exempt their own trusted automation from bot protection, without contacting support.
+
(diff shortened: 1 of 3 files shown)


jazzsequence (approved):
This is fine for now. 
However, this page is getting longer and increasingly complex. We might want to break this into multiple docs and add a separate next-gen cdn section for them to live in rather than cramming it all into a single page.

This page also is the _most trafficked page_ on the docs site after the home page. So, a _lot_ of eyes are on it, which is another reason we could/should improve the information architecture. 

Not blocking to do that now, just flagging for our future selves cc @rachelwhitton

john-lauro (approved):

src/source/releasenotes/2026-09-09-gcdn-bot-bypass-token.md
```diff
@@ -0,0 +1,21 @@
+---
+title: "Self-service bot bypass tokens for Next Generation GCDN"
+published_date: "2026-09-09"
```
jazzsequence:
```suggestion
published_date: "2026-09-10"
```

src/source/releasenotes/2026-09-09-gcdn-bot-bypass-token.md
```diff
@@ -0,0 +1,21 @@
...
+published_at: "2026-09-09T20:33:13Z"
```
jazzsequence:
```suggestion
published_at: "2026-09-10T15:36:09Z"
```

